PR #11, opened 2017-10-11
add start script

g-sam:
Something like this? As discussed in https://github.com/elifesciences/elife-xpub-formula/pull/14


giorgiosironi:
Where does `npx` come from? `npm` is provided by the apt package `nodejs` from `http://deb.nodesource.com/node_7.x`, but we aren't installing `npx` (yet).
Also, this repository is deployed in `/srv/elife-xpub` and `xpub` is in `/srv/xpub` so we may need to pass in a folder as a parameter?
(e.g.
```start.sh /srv/xpub```
```
cd $1;
# now run the command to start xpub
```

tamlyn:
`npx` comes with `npm` 5.2+ so probably not present as you're using node 7.x. We've updated our target node version to 8.x. But I think it's very likely that `npx` starts a child process in the same way as `npm`. 

In my experience `pm2` is the most common way to run node apps in production. It comes with monitoring and clustering and other useful things. But it's difficult to apply in this case due to the `pubsweet` cli hiding the entry point from us. 🤔

giorgiosironi:
Ok, so as-is this start script wouldn't work as we would have to change node versions (will do that separately) but without gaining much in stability :thinking:

tamlyn:
I've opened a ticket to talk about this https://gitlab.coko.foundation/pubsweet/pubsweet-cli/issues/89

g-sam:
This was addressed

PR #118, opened 2018-05-03
Allow the configuration of: 'pubsweet-server.baseUrl'

diversemix:


.env
```diff
@@ -2,3 +2,4 @@ COMPOSE_PROJECT_NAME=xpub
 XPUB_VERSION=5c6caa8fa1059391be5d8fdbdc57af34e5fc0098
 ORCID_CLIENT_ID=fake_client_id
 ORCID_CLIENT_SECRET=fake_client_secret
+PUBSWEET_BASEURL=http://localhost:3000
```
giorgiosironi: note: this value will be used in local checkouts of this project, but it should be overwritten by environment variables in the formula https://github.com/elifesciences/elife-xpub-formula/pull/21/files#diff-5dcf01c0620b18629b36a7139dab2b0aR28 as we expect

Issue #152, opened 2018-05-26
Manuscript Workflow


diversemix:
@JGilbert-eLife and I will be discussing this on 29th ... in a few spare moments today I have come up with https://github.com/diversemix/manuscript-workflow 

This will need reviewing with James and there is probably an Architectural discussion with @tamlyn at some point too.

diversemix:
Yesterday, after doing an hour or so's research, I can confirm that Tam is correct. That is to say that there seems to be no "out-of-the-box" way to incorporate user interactions (or any API/programmatically) to interact with the workflow.

Today, I am going to invest an hour or so to research https://github.com/paed01/bpmn-engine

diversemix:
This has been updated after the conversation / meeting

Issue #161, opened 2018-06-02
Add ADR for why we are using PubSweet


diversemix:
It is still not entirely apparent what PubSweet are bringing to the table. Gather this information and add to an ADR.

jsms90:
@diversemix Is your question in relation to any particular components/packages, or rather why we are using anything from pubsweet at all?

diversemix:
I want to gather the information on the technical need for pubsweet ... then we can address the other question ;)

diversemix:
Blocked on @BlueReZZ  sending some information.

hdrury1:
Do we have an ADR for this?

diversemix:
Ah moving this to In Progress again ... got some information from Paul - this has been delayed because of the other stuff going on. Also like to run it past Paul again - we need careful wording :)

diversemix:
Also as a part of this I will be re-indexing which affect the workflow ADR from #152

hdrury1:
PR: https://github.com/elifesciences/elife-xpub/pull/174

PR #162, opened 2018-06-04
Add an PR for Workflow elifesciences/elife-xpub#152

diversemix:
Created to allow discussion.

### DIFF
--- a/docs/adr/001-Workflow.md
+++ b/docs/adr/001-Workflow.md
@@ -1,6 +1,6 @@
 # 1. Workflow Solution for xPub
 
-Date: 2018-06-01
+Date: 2018-06-04
 
 ## Status
 


g-sam:
The library looks good. I was away when this was talked about so I'm not sure exactly what the ADR is proposing in terms of how the library will be used, but I get the picture. Seems like a good idea, as long as we don't end up with a bunch of conditional logic inside our transition functions.

diversemix:
The conversation was mostly around whether we should use a Workflow service. I'm still in favour of this in the long term.

diversemix:
I'm going to close this pull request as its quite old (now with conflicts). 

I will merge adding the workflow ADR (as accepted) as part of #161 -  if agreed @g-sam  @hdrury1  ?

PR #168, opened 2018-06-04
feat(submission): create graphql mutation for form submission on final page

dryhten:
* like this *

### DIFF
--- a/app/components/submission/SubmissionPage.js
+++ b/app/components/submission/SubmissionPage.js
@@ -13,11 +13,11 @@ import { schema as authorDetailsSchema } from './AuthorDetails/AuthorDetailsSche
 import { schema as manuscriptMetadataSchema } from './ManuscriptMetadata/ManuscriptMetadataSchema'
 import { schema as reviewerSuggestionsSchema } from './ReviewerSuggestions/ReviewerSuggestionsSchema'
 
-const FormStep = ({ history, nextUrl, updateSubmission, ...props }) => (
+const FormStep = ({ history, nextUrl, submitMutation, ...props }) => (
   <Formik
     onSubmit={(values, { setSubmitting, setErrors }) => {
       const data = omitDeep(values, ['__typename', 'files'])
-      return updateSubmission({ variables: { data } })
+      return submitMutation({ variables: { data } })
         .then(() => setSubmitting(false))
         .then(() => history.push(nextUrl))
         .catch(errors => {
@@ -30,7 +30,7 @@ const FormStep = ({ history, nextUrl, updateSubmission, ...props }) => (
 
 const SubmissionPage = ({ match, history }) => (
   <WithCurrentSubmission>
-    {(updateSubmission, initialValues) => (
+    {(initialValues, { updateSubmission, finishSubmission }) => (
       <Switch>
         <Route
           path={`${match.path}/upload`}
@@ -47,7 +47,7 @@ const SubmissionPage = ({ match, history }) => (
                   <FileUploadsPage {...props} />
                 </ComponentWithSaving>
               )}
-              updateSubmission={updateSubmission}
+              submitMutation={updateSubmission}
               validationSchema={fileUploadsSchema}
             />
           )}
@@ -67,7 +67,7 @@ const SubmissionPage = ({ match, history }) => (
                   <ManuscriptMetadata {...props} />
                 </ComponentWithSaving>
               )}
-              updateSubmission={updateSubmission}
+              submitMutation={updateSubmission}
               validationSchema={manuscriptMetadataSchema}
             />
           )}
@@ -87,7 +87,7 @@ const SubmissionPage = ({ match, history }) => (
                   <ReviewerSuggestions {...props} />
                 </ComponentWithSaving>
               )}
-              updateSubmission={updateSubmission}
+              submitMutation={finishSubmission}
               validationSchema={reviewerSuggestionsSchema}
             />
           )}
@@ -106,7 +106,7 @@ const SubmissionPage = ({ match, history }) => (
                   <AuthorDetailsPage {...props} />
                 </ComponentWithSaving>
               )}
-              updateSubmission={updateSubmission}
+              submitMutation={updateSubmission}
               validationSchema={authorDetailsSchema}
             />
           )}
--- a/app/components/submission/WithCurrentSubmission.js
+++ b/app/components/submission/WithCurrentSubmission.js
@@ -1,5 +1,5 @@
 import gql from 'graphql-tag'
-import { withApollo, Mutation } from 'react-apollo'
+import { withApollo } from 'react-apollo'
 import React from 'react'
 import ErrorPage from '../global/ErrorPage'
 
@@ -7,7 +7,6 @@ const manuscriptFragment = gql`
   fragment WholeManuscript on Manuscript {
     id
     title
-    source
     manuscriptType
     subjectAreas
     suggestedSeniorEditors
@@ -88,6 +87,15 @@ const UPDATE_SUBMISSION = gql`
   ${manuscriptFragment}
 `
 
+const FINISH_SUBMISSION = gql`
+  mutation FinishSubmission($data: ManuscriptInput!) {
+    finishSubmission(data: $data) {
+      ...WholeManuscript
+    }
+  }
+  ${manuscriptFragment}
+`
+
 class WithCurrentSubmission extends React.Component {
   constructor() {
     super()
@@ -133,13 +141,24 @@ class WithCurrentSubmission extends React.Component {
       return <ErrorPage error={error} />
     }
 
-    return (
-      <Mutation mutation={UPDATE_SUBMISSION}>
-        {updateSubmission =>
-          this.props.children(updateSubmission, initialValues)
-        }
-      </Mutation>
-    )
+    function updateSubmission({ variables }) {
+      return this.props.client.mutate({
+        mutation: UPDATE_SUBMISSION,
+        variables,
+      })
+    }
+
+    function finishSubmission({ variables }) {
+      return this.props.client.mutate({
+        mutation: FINISH_SUBMISSION,
+        variables,
+      })
+    }
+
+    return this.props.children(initialValues, {
+      updateSubmission: updateSubmission.bind(this),
+      finishSubmission: finishSubmission.bind(this),
+    })
   }
 }
 
--- a/app/components/submission/WithCurrentSubmission.js
+++ b/app/components/submission/WithCurrentSubmission.js
@@ -141,23 +141,17 @@ class WithCurrentSubmission extends React.Component {
       return <ErrorPage error={error} />
     }
 
-    function updateSubmission({ variables }) {
-      return this.props.client.mutate({
-        mutation: UPDATE_SUBMISSION,
-        variables,
-      })
-    }
-
-    function finishSubmission({ variables }) {
-      return this.props.client.mutate({
-        mutation: FINISH_SUBMISSION,
-        variables,
-      })
-    }
-
     return this.props.children(initialValues, {
-      updateSubmission: updateSubmission.bind(this),
-      finishSubmission: finishSubmission.bind(this),
+      updateSubmission: ({ variables }) =>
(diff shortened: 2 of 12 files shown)


dryhten:
Moved from gitlab: https://gitlab.coko.foundation/xpub/xpub-elife/merge_requests/72

dryhten:
We need some tests for the validations on last page submission of manuscript. These can be done as part of this issue or a new one.

server/submission/resolvers.js
```diff
@@ -68,6 +74,47 @@ const resolvers = {
...
+      await db.update(newManuscriptDb, data.id)
+
+      const mailData = {
+        from: config.get('mailer.from'),
+        to: newManuscript.submissionMeta.author.email,
+        subject: 'Congratulations! You submitted your manuscript!',
+        text: 'Your manuscript has been submitted',
+        html: '<p>Your manuscript has been submitted</p>',
+      }
+      Email.send(mailData)
```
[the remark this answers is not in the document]
dryhten: Right now if something throws in a resolver the error will be sent back to the client by apollo, and the frontend will display the error page with what error was generated. This happens in all resolvers (e.g. even when storing the data to the backend every 5s) so the user can get an error page without doing anything. Should this be treated in a separated issue?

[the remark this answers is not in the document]
dryhten: Also, regarding the retry strategy, that sounds good, but I think it should be done at pubsweet component level.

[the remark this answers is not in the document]
diversemix: hmmm.... that didn't happen for me... I got no error page ... have you tested it?

app/components/submission/SubmissionPage.js
```diff
@@ -30,7 +30,7 @@ const FormStep = ({ history, nextUrl, updateSubmission, ...props }) => (
 
 const SubmissionPage = ({ match, history }) => (
   <WithCurrentSubmission>
-    {(updateSubmission, initialValues) => (
+    {(initialValues, { updateSubmission, finishSubmission }) => (
```
tamlyn: Why not go full [RORO](https://medium.freecodecamp.org/elegant-patterns-in-modern-javascript-roro-be01e7669cbd) on this 😄

app/components/submission/WithCurrentSubmission.js
```diff
@@ -132,13 +140,24 @@ class WithCurrentSubmission extends React.Component {
...
+        variables,
+      })
+    }
+
+    function finishSubmission({ variables }) {
+      return this.props.client.mutate({
+        mutation: FINISH_SUBMISSION,
+        variables,
+      })
+    }
```
tamlyn: If you defined these as arrow functions you wouldn't need to `bind` them.

server/db-helpers/index.js
```diff
@@ -31,6 +31,7 @@ function manuscriptGqlToDb(manuscript, owner) {
 function manuscriptDbToGql(manuscriptDb, id) {
   const manuscript = _.cloneDeep(manuscriptDb)
   manuscript.id = id
+  delete manuscript.type
```
tamlyn: What does this do?

server/submission/definitions.js
```diff
@@ -167,4 +167,121 @@ const emptyManuscript = {
...
+  Joi.object().keys({
+    cosubmission: Joi.boolean().required(),
+    cosubmissionTitle: Joi.string().allow(''),
+    cosubmissionId: Joi.string().when('a', {
+      is: true,
+      then: Joi.required(),
+      otherwise: Joi.string().allow(''),
+    }),
+  }),
+)
```
tamlyn: Surely we should be reusing the yup validation schema instead of duplicating it all in joi?

server/submission/resolvers.js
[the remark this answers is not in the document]
tamlyn: This function is async so needs to be `await`ed.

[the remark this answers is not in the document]
dryhten: You are right, I tested this only with the query. For some reason the server/apollo doesn't log anything if you throw in a query resolver (e.g. `currentSubmission`) but it prints the error if you throw it in a mutation resolver. Also, the error page will only be displayed if the `currentSubmission` query fails but when a mutation fails nothing is displayed (this can be changed).

So do we do this try-catch only for the e-mail sending component in this mutation? There are other operations in different queries/mutations that could throw (e.g. db operations) and nothing is done in that regards.

If yes, what should be the expected behavior? If the e-mail is not sent/failed should we show some error to the user? Do we use the error page for that?

I suggest that for now we just catch errors from operations that can throw and log something on the backend. This should be done as part of this issue for `finishSubmission` mutation and part of another issue for other queries/mutations. Some errors like failing to store data to the database should be shown to the user in a nice format. Also, successfully storing the manuscript to the database at this stage should show success to the user somehow (do we have wireframes for this?). If the e-mail fails I won't show something to the user since this is just a confirmation e-mail that their manuscript has been received, but I would log something on the backend.

Should I work out some retry strategy and do a MR to pubsweet for their component? (this can take a while) or do I hack something in this PR for the moment?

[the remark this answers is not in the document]
dryhten: @tamlyn I have my doubts about that. Like I said in my previous comment, if we don't show the result of the e-mail component to the user not sure if we should wait for it. It's fine to wait for it to be stored to the db (so we can show success/failure), but does it make sense to wait for it and show something like "failure to send your confirmation e-mail"?

server/db-helpers/index.js
dryhten: Converts from DB schema to GraphQL schema :) there's no type in the GraphQL schema. This will help when we'll end up using Manuscript object for the business logic.

app/components/submission/WithCurrentSubmission.js
dryhten: Is that a preferred approach? Why? I was hesitating between the two.

server/submission/definitions.js
dryhten: I assume we still want to use Joi since this is backend stuff? I'll look into how we can easily convert between the two.

tamlyn: Why do you assume that?

app/components/submission/WithCurrentSubmission.js
tamlyn: See the top voted comment on that article for an accurate representation of my opinion on the issue. 

But I wasn't talking about class properties. I just meant to convert the local function statements into local arrow function expressions. Just because it's a bit simpler.

server/submission/resolvers.js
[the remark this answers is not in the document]
tamlyn: Sure, you don't have to await it but you absolutely do have to handle the error some way. The way you have it here will lead to an unhandled promise rejection and this will crash the server in future versions of node.

app/components/submission/SubmissionPage.js
dryhten: No particular reason. My thinking was that the second parameter is the mutations object. This would mean two things: first, the mutations are grouped together and second, you don't have to know all the possible mutations if you wrap your code in `<WithCurrentSubmission>`.

The blog post you linked makes sense, but I would normally have to think more about something before adopting it.

app/components/submission/WithCurrentSubmission.js
dryhten: In this case, since the functions are short (only one instructions) I prefer arrow functions so you have the code in the same place. If the functions were long I would define them somewhere else since this makes the code harder to read (my opinion).

dryhten:
For now I just log an error if the validations or e-mail sending fail. We need to do this for the DB operations as well, but considering this will change (relatively soon) I will leave that part of the code alone. I will open a new issue where we discuss what to do about re-sending e-mails (I assume someone else will open issues with what to display to the user when something on the backend fails throughout submission).

server/submission/definitions.js
dryhten: For a couple of reasons: PubSweet already uses Joi for validation on the backend. Keeping the number of packages we use across the entire project to a minimum sounds like a good thing. Second, coming from PubSweet to elife-xpub could be more confusing to a new developer if we switch from Joi to Yup all of a sudden. Third, they are not entirely identical, notice that this one has stage required, while the frontend one doesn't.

Also, something similar was previously discussed: https://gitlab.coko.foundation/xpub/xpub-elife/merge_requests/11

Your own words here :)
> I think reusing validation schemas between server and client is a non-goal so I'd be happy with using different libs on the server and the client.

But I do think we should store the schema in one place only and maybe slightly change it for frontend/backend stuff. I'm open to discussions here. I would be OK with using yup on both frontend and backend if everyone agrees that the arguments I mentioned above are not strong enough.

diversemix:
I'd do it as a part of this issue. I think the "principle" at play here is the commenting on the issue.

server/submission/definitions.js
tamlyn: Don't quote me, I don't know anything!

I think the difference is that in PubSweet the Joi validation schemas are intended to determine whether the entity as a whole is in a valid state. Here we're only interested in whether the user submitted data is valid which is what the yup schema tells us. Extra fields which are not supplied by the user (like stage) don't need to be validated. 

However the current schema doesn't account for all possibilities. For instance it doesn't validate that `submissionMeta.author` is required because in the context of the form that's not an error that can occur. I think it should be possible to augment the schema to validate these cases but it's worth checking to make sure.

server/submission/resolvers.js
```diff
@@ -69,6 +75,55 @@ const resolvers = {
...
+        throw new Error(errorCosubmission)
+      }
+
+      newManuscript.submissionMeta.stage = 'QA'
+      const newManuscriptDb = db.manuscriptGqlToDb(newManuscript, ctx.user)
+      await db.update(newManuscriptDb, data.id)
+
+      const mailData = {
+        from: config.get('mailer.from'),
+        to: newManuscript.submissionMeta.author.email,
```
tamlyn: Should this be sent to the author or the submitter?

server/submission/definitions.js
dryhten: As per our discussion, this was moved to https://github.com/elifesciences/elife-xpub/issues/206

server/submission/resolvers.js
dryhten: What's the submitter e-mail? Right now `submissionMeta.author.email` is the e-mail from first page.

tamlyn: Sorry, by submitter email I meant the email of the user submitting the submission (!). No matter though since the email is no longer required.

dryhten: So the ORCID user submitting the manuscript? Why is it no longer required?

dryhten:
Realized this was merged: https://github.com/elifesciences/elife-xpub/pull/202

This forces me to do some changes. I will rebase and push those changes tomorrow.

server/submission/resolvers.test.js
```diff
@@ -152,6 +150,113 @@ describe('Submission', () => {
...
+        },
+        hasCorrespondent: false,
+        discussedPreviously: false,
+        consideredPreviously: false,
+        cosubmission: false,
+      },
+    }
+    it('stores data to the backend with a new stage of QA', async () => {
+      const initialManuscript = getInitialManuscript()
+      const id = await save(manuscriptGqlToDb(initialManuscript, userId))
```
tamlyn: Would it make sense to put this stuff in a `beforeEach` just for this `describe`?

dryhten: Yep

dryhten: Damn, now I need 2 reviews again :)

server/db-helpers/index.js
jsms90: Why do we have 'manuscript type' in the GraphQL schema but not the DB schema? Does 'type' refer to research article/feature article/etc?

dryhten: Just saw this, it's the other way around. This function converts from a DB object to a GraphQL object. All DB objects have a type field since the only table is `entities` and you need a way to know the entity type (e.g. User, Manuscript etc). So no, it doesn't refer to research/feature article.

Issue #151, opened 2018-05-25
Facility to save form data


hdrury1:
- [ ] Form data entered should be saved when the user clicks away from the current field. 
- [ ] The cover letter should also be saved periodically during input, in a Google Docs-type manner.

tamlyn:
The form library we are using `Formik` only supports saving on submit but I suggest we use https://github.com/ramitos/formik-observer to circumvent that limitation. It's not widely used yet but it was created by another team at YLD so it should be solid.

dryhten:
I guess this is blocked by https://gitlab.coko.foundation/xpub/xpub-elife/merge_requests/64 right now, I'll look over it.

dryhten:
@hdrury1 another question from my discussion with @diversemix earlier: if someone starts a form submission on one computer, stops during the submission process then switches to a different computer, the form submission should appear as in progress in his dashboard, right? This would mean that we have to store everything in a database and not on a local storage on the browser. I would assume this is the case, otherwise the app behavior would be inconsistent (e.g. stopping during the submission process would show the submission as in progress on your dashboard but switching computers will make it disappear).

diversemix:
I did have a quick chat with @NickDuf ... we agreed that we perhaps need to make sure that whatever behaviour results in the implementation is "by design". I think the current design does make sense, however what are the data requirements (if any) for storing a user's data on the server without them actively submitting it?

tamlyn:
Does this functionality also apply to QC? i.e. if a staff member changes a field, should the changes be saved on blur? And if so, should they be saved even if the field is currently invalid?

hdrury1:
If I understand the query correctly, yes fields should be saved no matter who is changing them (i.e. staff or the submitter).

I personally think we should save invalid fields (@NickDuf?) but another thought did occur to me- staff should be able to override an invalid field in any instance.

tamlyn:
How important is it to save the form fields other than cover letter on blur only rather than saving all fields periodically à la Google Docs? I ask because it turns out latter is much easier to implement. The former will require wrapping each form field in a new component as I can't see a way to handle blur events at the form level.

hdrury1:
I don't see there being an issue from a UI perspective provided the interval is short enough.

diversemix:
Tam and I talked about this IRL ... as I remembered talking to @NickDuf about this. We agreed that we leave the UI indicator for now - waiting for a design from Nick and implement the - google-esk save after a few seconds of no user interaction.

[...]

PR #175, opened 2018-06-06
feat(submission-frontend): save form data to backend every 5s

dryhten:
re #151

### DIFF
--- a/app/components/submission/FileUploads/FileUploadsPage.js
+++ b/app/components/submission/FileUploads/FileUploadsPage.js
@@ -1,7 +1,6 @@
 import React from 'react'
 import gql from 'graphql-tag'
 import { Mutation } from 'react-apollo'
-import _ from 'lodash'
 import FileUploads from './FileUploads'
 
 const UPLOAD_MUTATION = gql`
--- a/app/components/submission/FileUploads/FileUploadsPage.js
+++ b/app/components/submission/FileUploads/FileUploadsPage.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import gql from 'graphql-tag'
 import { Mutation } from 'react-apollo'
+import _ from 'lodash'
 import FileUploads from './FileUploads'
 
 const UPLOAD_MUTATION = gql`
--- a/app/components/submission/SubmissionPage.js
+++ b/app/components/submission/SubmissionPage.js
@@ -2,8 +2,8 @@ import React from 'react'
 import { Route, Switch } from 'react-router-dom'
 import { Formik } from 'formik'
 import omitDeep from 'omit-deep-lodash'
-import _ from 'lodash'
 import WithCurrentSubmission from './WithCurrentSubmission'
+import ComponentWithSaving from './ComponentWithSaving'
 import AuthorDetailsPage from './AuthorDetails/AuthorDetailsPage'
 import FileUploadsPage from './FileUploads/FileUploadsPage'
 import ManuscriptMetadata from './ManuscriptMetadata/ManuscriptMetadata'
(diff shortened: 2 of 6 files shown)


dryhten:
There is some duplicated code (when we set the timer for each page) because I couldn't find any way to get form values from outside Formik (`props.values`). Feel free to comment here if you know how I could do this.

app/components/submission/FileUploads/FileUploadsPage.js
```diff
@@ -16,42 +17,58 @@ const UPLOAD_MUTATION = gql`
...
-            })
-          }
-          setFieldValue={setFieldValue}
-          {...props}
-        />
-      )
-    }}
-  </Mutation>
-)
+class FileUploadsPage extends React.Component {
```
diversemix: Could you not create a common base class called "ComponentWithSaving" and then the FileUploadsPage and AuthorDetails page extend this base?

app/components/submission/SubmissionPage.js
```diff
@@ -27,64 +38,100 @@ const FormStep = ({ history, updateSubmission, nextUrl, ...props }) => (
...
-            />
-          )}
-        />
-      </Switch>
-    )}
-  </WithCurrentSubmission>
-)
+const SubmissionPage = ({ match, history }) => {
+  // TODO update interval in ms - define this somewhere else
+  const updateInterval = 5000
```
dryhten: Open to suggestions for this one.

app/components/submission/SubmissionPage.js
```diff
@@ -12,9 +13,19 @@ import { schema as authorDetailsSchema } from './AuthorDetails/AuthorDetailsSche
...
+  }
+  const data = omitDeep(values, ['__typename', 'files'])
+  updateSubmission({ variables: { data } })
+  return true
+}
+
+const FormStep = ({ history, nextUrl, updateSubmission, ...props }) => (
   <Formik
     onSubmit={(values, { setSubmitting, setErrors }) => {
+      history.push(nextUrl)
```
dryhten: Ops, I'll remove this.

app/components/submission/FileUploads/FileUploadsPage.js
dryhten: I haven't seen much inheritance in Javascript before but I can't see why this would be bad in this case. This would solve my comment before https://github.com/elifesciences/elife-xpub/pull/175#issuecomment-395062980

g-sam: https://reactjs.org/docs/composition-vs-inheritance.html

dryhten: I was just thinking about composition :)

dryhten: I think I'm just going to go ahead and wrap all the pages in another component which should have access to `props.values` then, and do this bit of logic there. If anybody objects say it now :)

app/components/submission/SubmissionPage.js
g-sam: it could be a prop of the wrapper you are making?

dryhten: That's exactly what I did, I moved everything related to this data saving on timer to one place only.

app/components/submission/FileUploads/FileUploadsPage.js
diversemix: Ah HTML inheritance :D ... inheritance vs composition is very religious there are pluses and minuses on both sides. The article that @g-sam referenced is unfortunately more religious than helpful with its un-qualified assertions ...  This article is much better IMHO https://www.thoughtworks.com/insights/blog/composition-vs-inheritance-how-choose

g-sam:
@dryhten if you rebase on develop ci should pass

Issue #42, opened 2018-03-27
Create subject area control molecule (wireframe 006)


g-sam:
Latest wireframes:
![image](https://user-images.githubusercontent.com/15656538/40132256-3be78b34-5934-11e8-8738-73c791371746.png)

[requirements here](https://github.com/elifesciences/elife-xpub/issues/42#issuecomment-387726579)

---

Old version:
![multiple-choice-menu](https://user-images.githubusercontent.com/15656538/39183884-6a3f910c-47b9-11e8-8da0-0bf274b61399.png)

### Tasks
+ [ ] Build UI molecule
+ [ ] Request list of subject areas from eLife API on page load

tamlyn:
Does it make sense to park this until the mega-people-picker component is better specified? We can use the existing [Tags](https://ui-staging-zpqsip.gateway.ps.semioticsquares.com/#!/Tags) atom in the interim.

NickDuffield:
I think it might. The subject picker will differ from the people picker. I would avoid using the tags for now though as I am unsure if this is the correct component for this use.

g-sam:
This is now confirmed as a dropdown with checkboxes

jsms90:
Design TBC - see https://github.com/elifesciences/elife-xpub/issues/45#issuecomment-383931608

NickDuf:
**Example A:**  Default text area input
**Example B:** On focus scrollable full subject area list is displayed. Display 6 items before scroll. List closes on item selection or blur. List should be keyboard navigable using return for selection.
**Example C:** List is refined by entering letters. First letter of subject is used for matching and is further refined by following letters.
**Example D:** Single keyword input can be removed by clicking cross icon.
**Example E:** Two inputs displays validation message and disables the menu until an item if removed.

**Reference:** http://textextjs.com/manual/examples/autocomplete-with-tags.html
Some issues with the above jQuery example. Selecting items from the list using touch doesn't work properly. Scrolling can be problematic.

[subject-keyword-picker.zip](https://github.com/elifesciences/elife-xpub/files/1987828/subject-keyword-picker.zip)

jsms90:
Pre-existing components:
- https://github.com/bharani91/react-select-popover
- https://github.com/instructure-react/react-tokeninput
- https://github.com/reactjs/react-autocomplete
- https://github.com/i-like-robots/react-tags

jsms90:
| Features | React tags | React select |
| --- | --- | --- |
[...]

PR #182, opened 2018-06-07
WIP create subject area dropdown & add to manuscript metadata page

jsms90:
#42 

WIP while I write tests. Should have the necessary functionality for a review from product/design :+1:

image

dropdown

image

image


g-sam:
@jsms90 pls rebase on master for ci to work

jsms90:
@g-sam you mean on develop?

jsms90:
@g-sam Checks have passed but getting `503 Service Temporarily Unavailable` at http://review-42-subject-glqd9a.gateway.xpub.semioticsquares.com/ :slightly_frowning_face:

hdrury1:
Can now be used in context of whole form here: https://xpub-elife-demo.gateway.xpub.semioticsquares.com/login

@jsms90 How much extra work at this stage would it be to add bold/underline to typed letters within the filtered list? If a lot, I will open a separate ticket to come back to later.

jsms90:
@hdrury1 Now sorting out the bug where the values that a user selects aren't saved when they navigate away from the page & back again. Please feel free to click around and try and break things :wink:

app/components/submission/ManuscriptMetadata/ManuscriptMetadata.js
```diff
@@ -48,6 +54,16 @@ const ManuscriptMetadata = ({ handleSubmit, values, setFieldValue }) => (
       ]}
     />
 
+    <ValidatedField
+      component={SubjectAreaDropdown}
+      label="Subject areas"
+      name="subjectAreas"
+      onBlur={e => setFieldTouched('subjectAreas', true)}
```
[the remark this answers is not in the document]
jsms90: You mean inside `onChange` as well?

Nah, when a field's value is changed (i.e. using `setFieldValue` in the onChange handler), formik changes `touched` to `true` under the hood :+1: 

But on blur, we need to "set the state of touched imperatively", [as it says in formik's docs](https://github.com/jaredpalmer/formik#setfieldtouched-field-string-istouched-boolean-shouldvalidate-boolean--void), so that it will get picked up by the validation on submit.

Tbh I don't feel like this is paticularly clear from the documentation & kinda requires you to understand the internals of formik. I was thinking of raising an issue on it, on their repo :thinking: , but I'm unsure how they should provide more clarity.

Fortunately, [this example with `react-select](https://codesandbox.io/s/jRzE53pqR) is linked to in Formik's documentation, which was super helpful :grin: It essentially showed me what I needed to write, but I didn't really understand _why_ until I watched [this youtube video](https://www.youtube.com/watch?v=-tDy7ds0dag&t=957s) (where the creator of the repo talks things through), or whether it was definitely the thing I ought to be doing.

Hopefully those links will help you if you're ever grappling with formik yourself. It was one of many things that stumped me for a while, but results in only a short line of code to show for it. Such is the way sometimes :laughing:

Hope that helps?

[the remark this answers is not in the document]
diversemix: Ooo thanks very much - super helpful!

diversemix:
@jsms90 Is there any reason why  ManuscriptMetadata in `submission` and not in the `ui` folder?

jsms90:
Guys, this PR was still a WIP, while we get tests in. Normally, I wouldn't create a PR before the tests were written and the component was ready for review, but I completely understand your concern with how long it has taken to get this component built & I also knew that there were still some product/design questions to be answered. So I thought that it would be valuable for you to be able to see it, even while it's in progress.

@g-sam and I are pairing on the tests at the moment. I could create a new issue for writing tests if you want? But it might be a good idea for us to start ensuring that there are tests against components _before_ we merge? Otherwise we are just closing issues on features that aren't truly 'done' & creating technical debt for ourselves.

Also, I seriously don't mean any offence, but I imagine that, for now, it might be good to have some eyes across this from someone who is familiar with React before we merge? Only because I know React and the setup of our codebase is still quite new & I wouldn't want to be introducing code that doesn't hold up.

> Is there any reason why ManuscriptMetadata in submission and not in the ui folder?

This is a good question & if it were in the wrong place, it may be something good to catch before merging. Essentially, the `ui` folder is for atoms & molecules that are created via the principles of [atom design](http://atomicdesign.bradfrost.com/). It would make sense for a generic multi-select dropdown menu to sit inside this folder, ready to be re-used within other areas of the application. But the SubjectAreaDropdown is not reusable. It is an implementation which is context-specific & therefore belongs in the folder the holds that context (i.e. the metadata page).

If that's still not quite making sense, I would definitely suggest having a read through Brad Frost's website on atom design (linked above).

Let me know what you want to do re: opening a test-specific issue?

hdrury1:
My fault entirely. I got trigger-happy when I saw all the green ticks. WIP PRs shouldn't be merged (and won't be in future).

Best way forward now is probably to creat a new PR for the testing... and please keep regularly updating us via comments.

jsms90:
Ah ok! Noooo problem!

Sorry if that sounded at all harsh :disappointed_relieved: Once we get our flow down, I reckon you having the final sign off and merge would be perfect, @hdrury1 :+1: 
We could add you as a reviewer once there has already been a review from the devs. At that point it just needs sign off in terms of meeting product requirements? :grin:  What do you reckon?

As for testing, 2 days & 3 brains = finally being able to simulate the selection of an option from the dropdown list!!! :tada: :sweat_smile:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -0,0 +1,165 @@
...
+
+const findOptionByValue = (value, options) =>
+  options.find(option => option.value === value)
+
+export class ThemelessSubjectAreaDropdown extends React.Component {
+  constructor(props) {
+    super(props)
+
+    const savedOptions = this.props.savedValues.map(value =>
+      findOptionByValue(value, eLifeOptions),
```
g-sam: i would say this abstraction isn't helpful

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+    this.state = {
+      selectedOptions: savedOptions,
+      hasReachedMultiselectLimit: false,
+    }
+
+    this.handleChange = this.handleChange.bind(this)
```
g-sam: i think these are already bound since you're using what they call property initializers (arrow funcs as class methods)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -0,0 +1,165 @@
...
+      }),
+    }
+  }
+
+  // https://github.com/JedWatson/react-select/blob/v2/src/Select.js
+  // react-select exposes onChange: (ValueType, ActionMeta) => void
+  handleChange = selectedOptions => {
+    if (selectedOptions.length <= 2) {
+      this.setState({ selectedOptions, hasReachedMultiselectLimit: false })
+      if (this.props.onChange) this.props.onChange(selectedOptions)
```
g-sam: `if` unnecessary

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+    if (selectedOptions.length >= 2) {
+      this.setState({ hasReachedMultiselectLimit: true })
+    }
+  }
+
+  handleBlur = e => {
+    if (this.props.onBlur) this.props.onBlur(e)
```
g-sam: same: required anyway

g-sam: but maybe it shouldn't be required?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -0,0 +1,165 @@
...
+      name,
+      onBlur: this.handleBlur,
+      onChange: this.handleChange,
+      options: eLifeOptions,
+      styles: this.customReactSelectStyles,
+      value: selectedOptions,
+      filterOption: (option, rawInput) => {
+        const input = rawInput.replace(/^\s+|\s+$/g, '').toLowerCase()
+        const candidate = `${option.label} ${option.value}`
+          .replace(/^\s+|\s+$/g, '')
```
g-sam: regexp could maybe use a comment

g-sam: "strip whitespace from beginning and end of string"

g-sam: aka "trim"!

g-sam: https://www.w3schools.com/jsref/jsref_trim_string.asp n'est ce pas?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -0,0 +1,165 @@
...
+          .toLowerCase()
+
+        return candidate.substr(0, input.length) === input
```
g-sam: https://www.w3schools.com/jsref/jsref_startswith.asp

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.md
```diff
@@ -0,0 +1,9 @@
+```js
+<ThemelessSubjectAreaDropdown
```
g-sam: I'd add a theme prop or risk bugs!

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Do you mean it's not helpful to have it saved in a separate variable called 'savedOptions'? So you would have it directly inside `this.state`?
```
    this.state = {
      selectedOptions: this.props.savedValues.map(value =>
          findOptionByValue(value, eLifeOptions),
      ),
      hasReachedMultiselectLimit: false,
    }
```

Or do you mean that this mapping isn't necessary at all?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Ah yeah. Didn't clock that we had `transform-class-properties` in our babel setup :+1: 

Most of our components are stateless, so this doesn't really come up. But we do have a few stateful ones. One of these [does contain this binding pattern](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/submission/ComponentWithSaving.js#L15), which is partly why I went with it - for consistency.

Also, as far as I understand, `transform-class-properties` means that you don't need to use a constructor - see https://www.fullstackreact.com/articles/use-property-initializers-for-cleaner-react-components/ but we have this in most of our stateful components.

Is it perhaps worth deciding on a common convention that our team is going to use for these things?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Ah yeah, this was when I was imagining creating a MultiSelectDropdown component that was more generic, rather than SubjectAreaDropdown, which is very specific to this context. So my thinking was that `if (this.props.onChange)` would be good for the same reasons that you and I put it into pubsweet's Menu component - https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/201/diffs#83853fc5eb0bcac69c38969965b30fa045468482_165_163

But you're right, since SubjectAreaDropdown is totally context specific & `onChange` is required, this is meaningless :+1:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Removing `if (this.props.onBlur)` makes sense to me. But why would I not make it required?

When a user leaves the `react-select` component, this will cause a blur event. If my code just says
```
handleBlur = e => this.props.onBlur(e)
```
but onBlur is not a required prop, then won't this cause an error?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Good point! I copied this directly from [the default filter function used within `react-select`](https://github.com/JedWatson/react-select/blob/v2/src/filters.js#L13). I assume he's written it out explicitly with a regex, rather than using the `.trim` method, because of this line in the w3schools article that you linked:
> For browsers that do not support the trim() method, you can remove whitespaces from both sides of a string with a regular expression

But in his implementation, he does name this regex function - 'trimString', which is definitely more readable :+1:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Seems like Microsoft Edge doesn't support this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Interestingly, MDN seems to suggest that it is supported across all browsers
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Here's the reasoning behind adding this to the code at all:

[This is the commit](https://github.com/elifesciences/elife-xpub/pull/182/commits/d3ac5815d07a08aeaa1b52ef1692f97c1f316178) where I added this code in. Beforehand, when a user selected some options, clicked 'Next', then clicked 'Back', their selected options did not appear.

On clicking 'Next', the values were always being saved into the Formik form. So at first I thought I just needed to add a new prop (`savedValues={values.subjectAreas}`) to the ValidatedField component that surrounds the SubjectAreaDropdown - see [this line in that commit](https://github.com/elifesciences/elife-xpub/pull/182/commits/d3ac5815d07a08aeaa1b52ef1692f97c1f316178#diff-bec0d7799dc3aae97e34200382de8504R66)

But this didn't quite work because graphql returns [Object Object] & the user just saw 2 empty blue boxes with no words inside.

That's because I was saving the whole object:
```
{
    value: 'structural-biology-molecular-biophysics',
    label: 'Structural Biology and Molecular Biophysics',
}
```
rather than the value alone.

If we save the value alone in the database, then once we get the values back (e.g. `['structural-biology', 'cancer-biology']`, we need to convert this array back into the object format, containing value & label.

Does any of that make sense?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Ok, realised I can actually configure the existing createFilter function & don't need to define my own. Makes far more sense, since I was copying all this code from there anyway :sweat_smile:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
g-sam: yes if you make it not required you would need to keep the if!

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
g-sam: i was referring only to the `findOptionsByValue` func, which isn't reused

jsms90: Ohhhhh! Gotcha!

PR #183, opened 2018-06-07
feat(elife-theme): use latest version of elife-theme

dryhten:


yarn.lock
```diff
@@ -4198,6 +4198,14 @@ fs-extra@^5.0.0:
     jsonfile "^4.0.0"
     universalify "^0.1.0"
 
+fs-extra@^6.0.1:
```
dryhten: Why did this update fs-extra and request? I only updated `elife-theme` version in `package.json` and ran `yarn`. `elife-theme` only depends on `styled-components`.

g-sam: I think running yarn will do minor updates if the lockfile needs updating

diversemix: This answer it: 

```
$ git blame server/submission/package.json
7dfff4d1 (Mihail Dunaev 2018-06-01 14:40:34 +0100 14)     "fs-extra": "^6.0.1",
```

Remember the root package.json allows > 4.0.2

PR #189, opened 2018-06-08
fix(dropzone-molecule): fix react-dropzone warning about accepted files

dryhten:
#### Background

Second page of submission was giving a warning like this

```
Warning: Failed prop type: Invalid prop `accept` of type `array` supplied to `Dropzone`, expected `string`.
```

The `accept` prop from react-dropzone doesn't accept an array of strings, but apparently it works with a string of comma separated values: https://github.com/react-dropzone/react-dropzone/pull/142

Reloaded page 2 of submission, no more warnings. Also, I uploaded a docx, pdf and png files and it accepted the first 2 but not the last one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "react": "^16.2.0",
     "react-apollo": "^2.1.2",
     "react-dom": "^16.2.0",
-    "react-dropzone": "^4.2.9",
+    "react-dropzone": "^4.2.11",
     "react-hot-loader": "^4.1.2",
     "react-router-dom": "^4.2.2",
     "recompose": "^0.26.0",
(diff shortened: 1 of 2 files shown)


dryhten:
Right, I was looking more in-depth into how react-dropzone works. For anyone interested, react-dropzone uses `attr-accept` to check if the file is accepted. This package allows for both, a list of strings and a string of comma separated values: https://github.com/okonet/attr-accept/blob/master/src/index.js

However, react-dropzone was defining a string in PropTypes for accept, good job with this @tamlyn https://github.com/react-dropzone/react-dropzone/commit/f174b13f067a03b0c88da42e642eb0037b2ae586

I assume either implementation is fine, but I prefer the list one too.

PR #190, opened 2018-06-10
Enable debug logging for development builds

diversemix:
#### Background

Enables debug logging when running in development. Also removed "app.js" as its not used.

#### Any relevant tickets
None

Yes, locally ran test and built the docker containers

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,7 +1,0 @@
-const logger = require('@pubsweet/logger')
-const startServer = require('pubsweet-server')
-
-startServer().catch(err => {
-  logger.error('FATAL ERROR, SHUTTING DOWN:', err)
-  process.exit(1)
-})
--- a/config/development.js
+++ b/config/development.js
@@ -1,4 +1,7 @@
 const { deferConfig } = require('config/defer')
+const winston = require('winston')
+
+winston.level = 'debug'
 
 module.exports = {
   'pubsweet-server': {
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN yarn install --frozen-lockfile \
     && yarn cache clean \
     && rm -rf /npm-packages-offline-cache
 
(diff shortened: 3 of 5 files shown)


app.js
```diff
@@ -1,7 +0,0 @@
-const logger = require('@pubsweet/logger')
```
g-sam: what's the thinking behind deleting this? we need it to start the app.

diversemix: Do we? .... where is it called from? Everywhere I look the app is started via `pubsweet server`.

g-sam: https://gitlab.coko.foundation/yld/xpub-deployment-config-postgres/blob/master/templates/xpub-elife-deployment.yaml#L75

diversemix: I think this also is another good reason we need an explanation / documentation on how the deployment process works :)

g-sam: `pubsweet server` starts the process with forever, this is intended for production environments where we do not know how the app will be started.

diversemix: ... can you expand on "where we do not know how the app will be started" please

config/development.js
```diff
@@ -1,4 +1,7 @@
 const { deferConfig } = require('config/defer')
+const winston = require('winston')
+
+winston.level = 'debug'
 
 module.exports = {
   'pubsweet-server': {
```
dryhten: Don't you need to add the logger to `pubsweet-server.logger` as per https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/logger#usage

diversemix: this is already done in `default.js`

diversemix:
Ooops ... I was rebuilding the branch (as I'm not removing app.js anymore) - didn't realise it would close this PR ... sorry ... keep an eye out for the new one ;)

diversemix:
Re-opened as PR #191

config/development.js
dryhten: OK, I had my doubts that logger will be the same object, but I just tested this and it seems to be the case.

PR #191, opened 2018-06-11
Enable debug logging for development env

diversemix:
#### Background

Enables debug logging for the development env

Tested locally


tamlyn:
I would also explicitly set the logger to this instance of winston. Mostly because it makes it clearer what is happening but also because the default logger set in @pubsweet/logger could change.

Issue #188, opened 2018-06-08
Fix form validation errors


dryhten:
**Describe the bug**
The validation error messages have some problems:
1. The dropzone molecule error message: it shows `Please upload your Manuscript.` randomly, without clicking next.
2. The suggested reviewers show bad validation messages, e.g. `suggestedReviewers[0].email is a required field`.

**To Reproduce**
1. Go to first page, click `Pre-fill my details using ORCiD`, click next to go to second page, click on cover letter, then click outside.
2. Fill in the form up to the last page (reviewers suggestions). Click on a field then move away from it.

**Expected behavior**
1. Shouldn't happen.
2. A nicer message should be displayed (the one passed in to yup)

**Desktop (please complete the following information):**
 - OS: Debian GNU/Linux 9.3 (stretch), Kernel 4.9.0-4-amd64 x86_64 (64 bit gcc: 6.3.0)
 - Browser: Chromium 
 - Version 66.0.3359.117 (Developer Build) built on Debian 9.4

Issue #148, opened 2018-05-25
Copy&Paste rot


diversemix:
Below is an example that needs fixing, also search for others.

app/components/submission/AuthorDetails/AuthorDetailsPage.js
12-    {(updateSubmission, initialValues) => (
13-      <Formik
14-        component={AuthorDetails}
15-        initialValues={initialValues}
16-        onSubmit={(values, { setSubmitting, setErrors }) => {
17:          const data = omitDeep(values, ['__typename', 'files'])
18-          return updateSubmission({ variables: { data } })
19-            .then(() => setSubmitting(false))
20-            .then(() => history.push('/submit/upload'))
21-            .catch(errors => {
22-              setErrors(errors)

app/components/submission/ManuscriptMetadata/ManuscriptMetadataPage.js
14-        {(updateSubmission, initialValues) => (
15-          <Formik
16-            component={ManuscriptMetadata}
17-            initialValues={initialValues}
18-            onSubmit={(values, { setSubmitting, setErrors }) => {
19:              const data = omitDeep(values, ['__typename', 'files'])
20-              return updateSubmission({ variables: { data } })
21-                .then(() => setSubmitting(false))
22-                .then(() => history.push('/submit/suggestions'))
23-                .catch(errors => {
24-                  setErrors(errors)

app/components/submission/ReviewerSuggestions/ReviewerSuggestionsPage.js
14-        {(updateSubmission, initialValues) => (
15-          <Formik
16-            component={ReviewerSuggestions}
17-            initialValues={initialValues}
18-            onSubmit={(values, { setSubmitting, setErrors }) => {
19:              const data = omitDeep(values, ['__typename', 'files'])
20-              return updateSubmission({ variables: { data } })
21-                .then(() => setSubmitting(false))
22-                .then(() => history.push('/'))
23-                .catch(errors => {
24-                  setErrors(errors)

app/components/submission/FileUploads/FileUploadsPage.js
26-      <Mutation mutation={UPLOAD_MUTATION}>
27-        {(uploadFile, { data: uploadData, loading, error: uploadError }) => (
28-          <Formik
29-            initialValues={initialValues}
30-            onSubmit={(values, { setSubmitting, setErrors }) => {
31:              const data = omitDeep(values, ['__typename', 'files'])
32-              return updateSubmission({ variables: { data } })
33-                .then(() => setSubmitting(false))
34-                .then(() => history.push('/submit/metadata'))
35-                .catch(errors => {
[...]

PR #194, opened 2018-06-11
fix: use key to force remount of formik component on each page

tamlyn:
#### Background

Moving from individual form instances to a single form instance for the whole submission form had the side effect of persisting field "touched" status across steps in the wizard. This manifested itself as a bug which caused validation errors for all fields to be displayed as soon as any field was blurred. 

See #188 

Locally, manually.

### DIFF
--- a/app/components/submission/SubmissionPage.js
+++ b/app/components/submission/SubmissionPage.js
@@ -13,8 +13,18 @@ import { schema as authorDetailsSchema } from './AuthorDetails/AuthorDetailsSche
 import { schema as manuscriptMetadataSchema } from './ManuscriptMetadata/ManuscriptMetadataSchema'
 import { schema as reviewerSuggestionsSchema } from './ReviewerSuggestions/ReviewerSuggestionsSchema'
 
-const FormStep = ({ history, nextUrl, submitMutation, ...props }) => (
+const FormStep = ({
+  component: Component,
+  history,
+  nextUrl,
+  submitMutation,
+  initialValues,
+  validationSchema,
+}) => (
   <Formik
+    initialValues={initialValues}
+    // ensure each page gets a new form instance otherwise all fields are touched
+    key={Component.name}
     onSubmit={(values, { setSubmitting, setErrors }) => {
       const data = omitDeep(values, ['__typename', 'files'])
       return submitMutation({ variables: { data } })
@@ -24,7 +34,12 @@ const FormStep = ({ history, nextUrl, submitMutation, ...props }) => (
           setErrors(errors)
         })
     }}
-    {...props}
+    render={formProps => (
+      <ComponentWithSaving updateSubmission={submitMutation} {...formProps}>
+        <Component {...formProps} />
+      </ComponentWithSaving>
+    )}
+    validationSchema={validationSchema}
   />
 )
 
@@ -36,17 +51,10 @@ const SubmissionPage = ({ match, history }) => (
           path={`${match.path}/upload`}
           render={() => (
             <FormStep
+              component={FileUploadsPage}
               history={history}
               initialValues={initialValues}
               nextUrl={`${match.path}/metadata`}
-              render={props => (
-                <ComponentWithSaving
-                  {...props}
-                  updateSubmission={updateSubmission}
-                >
-                  <FileUploadsPage {...props} />
-                </ComponentWithSaving>
-              )}
               submitMutation={updateSubmission}
               validationSchema={fileUploadsSchema}
             />
@@ -56,17 +64,10 @@ const SubmissionPage = ({ match, history }) => (
           path={`${match.path}/metadata`}
           render={() => (
             <FormStep
+              component={ManuscriptMetadata}
               history={history}
               initialValues={initialValues}
               nextUrl={`${match.path}/suggestions`}
-              render={props => (
-                <ComponentWithSaving
-                  {...props}
-                  updateSubmission={updateSubmission}
-                >
-                  <ManuscriptMetadata {...props} />
-                </ComponentWithSaving>
-              )}
               submitMutation={updateSubmission}
               validationSchema={manuscriptMetadataSchema}
             />
@@ -76,18 +77,11 @@ const SubmissionPage = ({ match, history }) => (
           path={`${match.path}/suggestions`}
           render={() => (
             <FormStep
+              component={ReviewerSuggestions}
               history={history}
               initialValues={initialValues}
               nextUrl="/dashboard"
-              render={props => (
-                <ComponentWithSaving
-                  {...props}
-                  updateSubmission={updateSubmission}
-                >
-                  <ReviewerSuggestions {...props} />
-                </ComponentWithSaving>
-              )}
-              submitMutation={finishSubmission}
+              submitMutation={updateSubmission}
               validationSchema={reviewerSuggestionsSchema}
             />
           )}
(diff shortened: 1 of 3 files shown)


app/components/submission/SubmissionPage.js
```diff
@@ -24,10 +32,38 @@ const FormStep = ({ history, nextUrl, updateSubmission, ...props }) => (
...
+    render={formProps => (
+      <ComponentWithSaving updateSubmission={updateSubmission} {...formProps}>
+        <TouchedStateManager component={component} {...formProps} />
+      </ComponentWithSaving>
+    )}
+    validationSchema={validationSchema}
   />
 )
 
+// form touched state needs to be cleared when moving between steps
```
diversemix: Can we not have another file for this class? Pleeeeease :)

jsms90:
@hdrury1 @NickDuf I remember once hearing this idea floated before - having all 4 sections of the form being shown on the same page, rather than being separate pages where you click 'Next' between each one? But I don't remember the reasoning for sticking with 4 separate pages?

Having everything on one page would seem to make sense, as this is (currently) a desktop application & we have plenty of real estate that is going unused at the bottom of the page. (If I remember correctly, there are no plans to make designs for mobile?)

There are also a few UX benefits associated with enabling users to scroll, rather than click 'Next' 3 times.

Imagine being an author who hasn't got their cover letter ready yet. You would still able to fill in the details on page 3 & 4. Wouldn't it be nice to be able to fill in all of the easy information first & come back something that may take more time, like the cover letter? We don't currently allow them to get to page 3 unless they've already submitted their manuscript and cover letter.

Imagine being an author, who has already filled out the whole form, but maybe you've filled in different sections at different times. Before clicking 'Submit' at the end, wouldn't you want to quickly look over everything that you've input so far, before you make the final click? Right now, they have to press 'Back' a bunch of times in order to do that. If you allow them to see everything at once, we might reduce the number of errors that a user makes & therefore alleviate slight pressure from QC.

If the wizard was all available on one screen, then this implementation problem wouldn't even be relevant.

Apologies if this idea has already been discussed at length before & I'm just not aware. But this implementation issue arises from the fact that we're treating it as a single form in one way, but as 4 separate forms in another way. So I guess I'm just wondering what has motivated the 4 steps in the first place?

dryhten:
The code LGTM, other than the hacky stuff happening in `TouchedStateManager` :)

@jsms90 this issue comes from https://github.com/elifesciences/elife-xpub/issues/148

jsms90:
Yeah thanks @dryhten. This is exactly what I mean. By refactoring the code, it's now being treated as a single form - hence the issue with `touched` being true for each field across all 4 pages. We are loading the information for all 4 pages at the same time from GraphQL.

So internally, we are considering these 4 pages to be a single entity. It's only from a user perspective that we're considering them as 4 separate things. This is leading to problems. @tamlyn's "hacky" workaround seems a good compromise to me. But I was just wondering why we aren't treating it as a single page from the user's perspective too?

hdrury1:
@jsms90 To answer you first question, the 'wizard' approach to the form was decided upon to make it as pain-free as possible for an author to fill out. 70% of submissions are rejected at this point so it's going to be the only instance of xPub a lot of authors experience and we want them to come back and use it again. We present them with one task at a time, pre-filling as many of the fields as we can.

User research suggests that at the point someone sets out to make an initial submission to eLife they have everything ready to go, including their cover letter and manuscript. Again, from research, we anticipate the majority of users will be completing the form in one go rather than leaving it and coming back days later, but of course we need to allow for edge cases by saving as we go along.

Once the form hits QC- and from then on- it is all on one page. This is because of the different way that staff are interacting with it and in this case having to click 'Next' repeatedly to access the next section of information is an unnecessary hardship.

jsms90:
Thanks so much @hdrury1. That makes perfect sense!. I assume that breaking it down into stages is about trying to reduce how overwhelming it might feel to fill out (vs doing it all at once)? Hence making them more likely to "come back and use it again"?

And especially knowing that they have all their info ready beforehand, I guess they wouldn't need to preview it (ie.. see all pages at once) before clicking Submit?

Thanks for the context. Really helps!

--- 

FYI @hdrury1 this isn't going to be merged. @tamlyn's going with [@g-sam's suggestion above](https://github.com/elifesciences/elife-xpub/pull/194#pullrequestreview-128365127)

hdrury1:
@diversemix Can this be closed if we're going with another solution?

diversemix:
@hdrury1 I don't feel qualified to review this ... though @dryhten has given it a lgtm ... perhaps he can approve?

tamlyn:
I got a moment. It works. Thanks @g-sam! 🎉 🎉 🎉

app/components/submission/WithCurrentSubmission.js
```diff
@@ -89,42 +89,47 @@ const UPDATE_SUBMISSION = gql`
...
+              .then(result => this.setData(result.data))
+              .catch(error => this.setError(error))
+          }
+        },
+        error => this.setError(error),
+      )
+  }
+
+  setData(data) {
+    this.setState({ data, error: undefined, loading: false })
```
jsms90: Out of curiosity, why do we need `error: undefined` here, rather than `error: null` like we have in the initial state?

app/components/submission/WithCurrentSubmission.js
```diff
@@ -89,42 +89,47 @@ const UPDATE_SUBMISSION = gql`
...
+              })
+              .then(result => this.setData(result.data))
+              .catch(error => this.setError(error))
+          }
+        },
+        error => this.setError(error),
+      )
+  }
+
+  setData(data) {
```
dryhten: Do we have something to deal with this? According to [transform-class-properties](https://babeljs.io/docs/en/babel-plugin-transform-class-properties/) you should have something like this

```
setData = (data) => {...}
```

Same goes for `setError()`.

app/components/submission/WithCurrentSubmission.js
tamlyn: Good point. I don't think we do.

app/components/submission/WithCurrentSubmission.js
tamlyn: You mean that we should always use class properties instead of class methods? Has that been decided?

dryhten: My first question was how can you use `this` inside that function if it was not bound?

Regarding class properties instead of class methods, what is the outcome of this? https://github.com/elifesciences/elife-xpub/issues/201

tamlyn: You only need to bind `this` if you're passing the function as an argument (or prop). If you're calling it as `this.whatever()` then that implicitly binds `this`. It's weird but that's how JS was designed.

dryhten: What if how we call `setData` changes? Is it safe to leave it like this? I would expect this to lead to bugs.

tamlyn: > I would expect this to lead to bugs.

Yes, `this` is a footgun. Continuing the discussion over at https://github.com/elifesciences/elife-xpub/issues/201#issuecomment-397652378

dryhten: Ok, I guess this can go for now and we continue discussion there. I assume you tested and it works, everything else LGTM.

app/components/submission/WithCurrentSubmission.js
jsms90: Is this being changed before we merge?

jsms90: No?

I'm not saying it should have been. Just unsure why it's like this? :woman_shrugging:

tamlyn: Ah, sorry! Should have been, yeah. Will change when I'm next there.

Issue #195, opened 2018-06-11
Test SubjectAreaDropdown component


jsms90:
Tests for SubjectAreaDropdown #42 (component easily reviewed from PR #182)

Product requirements for component behaviour:
- [x] When a user first visits the page, then no validation message is displayed
- [ ] Given that no input is typed & no option chosen, when the user leaves the field (blur), then validation message is shown, prompting user to choose at least 1 option
- [x] When a user focuses on the field, then a dropdown appears with full list of options
- [x] Given that no input has been typed, when a single letter is typed, then the dropdown list is filtered by first letter
- [x] Given that a single letter has been typed, when the 2nd letter is typed, then the dropdown is filtered further (by first 2 letters)
- [x] When an option is selected from dropdown -> tag appears & dropdown closes
- [x] When 2 options are selected -> displays success message
- [ ] When 2 options are selected -> disables the dropdown
- [ ] Given that 2 options are selected, when 1 option is deleted, then the dropdown is re-enabled

Product requirements outside scope of testing (because this functionality is encapsulated within `react-select` itself):
- List should be keyboard navigable
- Use return for selection
- Tag can be removed by clicking cross icon

jsms90:
Simulating the selection of an option from the dropdown list has been a lot trickier than expected. As I understand it, this is largely the case because of the non-intuitive behaviour of enzyme.

You can see that this is somewhat common problem for people, based on the following issues:
- "[react-select simulate 'change' not working](https://github.com/airbnb/enzyme/issues/400)" (enzyme's repo)
- "[Simulate 'change' doesn't work for select](https://github.com/JedWatson/react-select/issues/1357)" (react-select repo)
- "[Help with using react-select and adding test](https://github.com/JedWatson/react-select/issues/1523)" (react-select repo)
- "[Testing react select component](https://stackoverflow.com/questions/41991077/testing-react-select-component)" (Stack Overflow question)

But despite the titles of these issues, the problem does not seem to originate with `react-select`. Many issues have been raised in enzyme that point to similar problems for people who are not using react-select.

It seems as though the crux of the issue is that the `.simulate` method does not behave quite as you might expect.

One prevalent answer to people's questions across the enzyme repo is the idea that calling `props` is better than using the `.simulate` method:
> [We get a lot of these 3rd party lib questions like this where simulating a prop "event" doesn't work.](https://github.com/airbnb/enzyme/issues/222#issuecomment-256717634)

[...]

PR #198, opened 2018-06-13
make requested changes to subject area dropdown component

jsms90:
#### Background
Response to @g-sam's requested/suggested changes in [old WIP PR](https://github.com/elifesciences/elife-xpub/pull/182/) - relates to issue #42 

### What does this PR do?
- use existing filter function & configure to match from start of string, rather than creating own filter function  
resolves https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194454268 & https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194455388
- add theme prop in styleguidist file  
resolves https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194455641
- remove unnecessary if statements from event handlers  
resolves: https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194452719 & https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194453834
- remove function bindings because of `transform-class-properties` in babel  
https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194452326
- remove findOptionByValue abstraction  
resolves https://github.com/elifesciences/elife-xpub/pull/182#discussion_r194451042

Encapsulated in separate ticket #195

### DIFF
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled, { withTheme } from 'styled-components'
 import { th } from '@pubsweet/ui'
-import Select from 'react-select'
+import Select, { createFilter } from 'react-select'
 
 const Root = styled.div`
   display: flex;
@@ -61,24 +61,18 @@ const eLifeOptions = [
   },
 ]
 
-const findOptionByValue = (value, options) =>
-  options.find(option => option.value === value)
-
-export class ThemelessSubjectAreaDropdown extends React.Component {
+class SubjectAreaDropdown extends React.Component {
   constructor(props) {
     super(props)
 
     const savedOptions = this.props.savedValues.map(value =>
-      findOptionByValue(value, eLifeOptions),
+      eLifeOptions.find(option => option.value === value),
     )
 
     this.state = {
       selectedOptions: savedOptions,
       hasReachedMultiselectLimit: false,
     }
-
-    this.handleChange = this.handleChange.bind(this)
-    this.handleBlur = this.handleBlur.bind(this)
 
     this.customReactSelectStyles = {
       multiValue: (base, state) => ({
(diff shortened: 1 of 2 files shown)


app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -93,16 +90,14 @@ export class ThemelessSubjectAreaDropdown extends React.Component {
...
     }
     if (selectedOptions.length >= 2) {
       this.setState({ hasReachedMultiselectLimit: true })
     }
   }
 
-  handleBlur = e => {
-    if (this.props.onBlur) this.props.onBlur(e)
-  }
+  handleBlur = e => this.props.onBlur(e)
```
dryhten: Is this line needed anymore?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.md
```diff
@@ -1,9 +1,11 @@
 ```js
-<ThemelessSubjectAreaDropdown
+const theme = require('@pubsweet/elife-theme').default
```
dryhten: Do we need default here?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: Yes

Without it, a blur event occurs on the input field - an element encapsulated within the Select component (from the `react-select` library) - but this isn't handled within the SubjectAreaDropdown component & therefore the ValidatedField component doesn't know to do anything on blur

dryhten: What I meant was: you wrote a function that calls another function with the same arguments. A better way to write this (in my opinion) would be:

```
...
onBlur: this.props.onBlur
...
```

dryhten: Now looking better at your code, just destructure it on line 104, along with `name` and other props. Also, sorry for being too vague if that's the case.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.md
jsms90: Yes

Because of `exports default myComponentName`, the theme prop looks like this
image without `.default` on the end

jsms90:
@g-sam Just removed the findOptionsByValue function. Thanks :+1:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.md
dryhten: Wait, now I'm really confused. Why do you use `ThemelessSubjectAreaDropdown` just to pass in the theme component? Can't you use `SubjectAreaDropdown`? It will have the theme passed down from the Wrapper here https://github.com/elifesciences/elife-xpub/blob/90240af14c656f3fe4e29234615eab6bcc7bc7fe/styleguide/components/Wrapper.js

Also, why did you use a different name for the class without theme? And why did you export it? You can't use `ThemelessSubjectAreaDropdown` without theme in props, that would lead to bugs. I can't comment on the original PR but my suggestion here is to do this:

1. rename `ThemelessSubjectAreaDropdown` to `SubjectAreaDropdown` and don't export it, like so
```
class SubjectAreaDropdown extends React.Component {
...
```
2. export SubjectAreaDropdown, but with theme object added to context
```
export default withTheme(SubjectAreaDropdown)
```

Look at this https://github.com/elifesciences/elife-xpub/blob/develop/app/components/ui/atoms/icons/Upload.js

Then the README should be changed to reflect that.

g-sam:
just simplicity -- otherwise you need to wrap it in a themeprovider. I didn't think about the name, but you're right, it's wierd.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.md
jsms90: See the conversation with @g-sam below

g-sam: yeah, mihail is right -- I forgot it was already wrapped

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: I'm not sure that I understand. How would you "destructure" this? Where would I write `onBlur: this.props.onBlur`?

`handleBlur` is a function definition, not a variable. So even if I called the thing on the left & the thing on the right by the same name, I can't do something like:
`{ onBlur } = e => this.props`.

Besides this, there is a good reason for calling this `handleBlur` instead of `onBlur`. handleBlur is a method on SubjectAreaDropdown & onBlur is a prop. They serve different purposes and the naming conventions reflect that.

jsms90: Ah!!! Destructure [on this line](https://github.com/elifesciences/elife-xpub/pull/198/files#diff-1be53589f0d3af8629dbf22cdcb80cc6R99)

Beautiful, thanks @dryhten :+1:

jsms90:
Okie doke!! All requests addressed & rebased into 1 commit. Ready to meeeeerge

diversemix:
@jsms90 I restarted the "test" job in the pipeline for some reason it had reported it was cancelled!

PR #200, opened 2018-06-13
Update issue templates

hdrury1:


jsms90:
@hdrury1 My understanding was that we were using milestones = features & each milestone/feature contains multiple issues/user stories

So I would suggest that perhaps we change the template to have "user story" at the top before "Plain text description"?

Seems like it would be worth leaving the "title" part in? But maybe not "Label", unless you still want epic labels attached?

g-sam:
I requested title be removed as there is already a title.

g-sam:
and not sure this line is needed: https://github.com/elifesciences/elife-xpub/pull/200/files#diff-1c7f865ae6347b0b4a90ead76b15594eL15

jsms90:
Also, I'm not sure what this means, when talking about labels: "one to indicate the nature of the ticket"?

jsms90:
Hmm... since we're now defining our issues via user stories & requirements, do we even need the rest of this template?

> #### Technical description (optional):
>
> Include relevant links/references
>
> #### Possible solutions:
> 1. 
> 2. 
> 3. 

This looks more like implementation details i.e. the kind of thing that someone picking up a ticket will decide, only after they start doing it? Thoughts @g-sam @tamlyn @dryhten ?

hdrury1:
@jsms90 You're right. I envisioned someone picking up a ticket and then filling this bit in.

jsms90:
So if you're raising these tickets, we don't need the implementation bit, right? That'll just come in the comments below, if necessary?

hdrury1:
It's a delete-as-applicable scenario. If you can answer as you raise the ticket, great. If not, delete it. Unless there are any major disagreements I think this can be merged now.

.github/ISSUE_TEMPLATE/custom.md
```diff
@@ -1,28 +1,18 @@
 ---
-name: Feature issue template
-about: A broad template for feature-building issues
+name: Issue Template
+about: Each issue should have at least two labels (one to indicate the corresponding epic and one to indicate the nature of the ticket, e.g. client-side or server-side). GitHub milestones correspond to features of the product and issues will be grouped by the feature they relate to.
```
tamlyn: Does this need updating with your latest thinking on milestones?

jsms90: This is the update, no?

jsms90:
@hdrury1 I suggest that we also have a link within each issue, which takes us directly to the relevant invision designs?

For example, @tamlyn just asked whether we have access to the latest designs for the components on page 4 - i.e. the reviewer suggestions page (which I don't think that we do)? If/when issues are raised for those features, it would be useful to have links to invision in there.

I reckon the things that a dev needs, in order to pick up a ticket are:
1. User story
2. Product requirements (functional and non-functional)
3. Designs

If we want a more detailed breakdown to prompt:

### User story
As a...
I want...
So that...

### Requirements
Functional:
- x
- y
- z

Non functional:
- x
- y
- z

Note: some people find this a useful framework, **given**....**when**....**then**....
e.g. given that I have chosen 2 items from the dropdown, when I click on the dropdown again, then the menu does not appear
How prescriptive you go with the language is up to you. Just thought this might be helpful

#### Designs
Link to invision

.github/ISSUE_TEMPLATE/custom.md
```diff
@@ -1,28 +1,18 @@
...
-
-What are you building? What value is this technical task adding? Include relevant links/references
-
-**Technical description (optional)**
+**Technical description (optional):**
 
 Include relevant links/references
 
-**Possible solutions**
+**Possible solutions:**
```
jsms90: If there are many possible solutions, that would indicate to me that a discussion issue is required & presumably then an ADR. Or, if no discussion is needed, then requiring a dev to outline all the ways they _might_ implement something may not be especially illuminating?

So, if "Possible solutions" is filled out by someone, then the issue probably isn't ready to be picked up yet. With that in mind, is this a useful section to have in our template? Aren't we better off raising a new issue, which outlines the possible choices & prompts that discussion?

Genuine question - it might be that I can't think of a use case, but that there is one? :woman_shrugging:

jsms90:
LGTM. Thanks for your patience @hdrury1 :+1:

PR #202, opened 2018-06-13
test: refactor resolver tests to bypass apollo server

tamlyn:
#### Background

PubSweet server uses require-relative to load server components and this breaks jest mocking. Testing the resolvers directly circumvents the issue.

Benefits 
- jest mocks work as expected
- tests are shorter to write
- tests run a bit faster

Down sides 
- a little less realistic as all middleware is bypassed
- each test has to set up the context (currently just the user ID but may grow)

n/a

### DIFF
--- a/server/submission/resolvers.test.js
+++ b/server/submission/resolvers.test.js
@@ -1,16 +1,10 @@
-const supertest = require('supertest')
 const _ = require('lodash')
 const config = require('config')
 const fs = require('fs-extra')
-const { jsonToGraphQLQuery } = require('json-to-graphql-query')
-const app = require('pubsweet-server/src/').configureApp(require('express')())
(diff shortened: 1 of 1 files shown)


diversemix:
@g-sam You really don't think a comment would not help a developer from making a mistake? - My general principle is the Golden Rule - It feels a little deliberately obtuse to not help someone when you have the chance.

g-sam:
Generally I try as hard possible not have extra words unless they seem absolutely necessary (an endeavour at which I constantly, hopelessly, fail). That's just my opinion but I wont complain if you think it's important here.

diversemix:
@g-sam I'm still trying to get used to no headers in files to explain why they exist and their context - I do miss this aspect of commercial coding.

PR #205, opened 2018-06-14
ADR for Code Structure

diversemix:
#### Background

A design record for the way to structure the server-side code.

### DIFF
--- a/docs/adr/0004-code-structure.md
+++ b/docs/adr/0004-code-structure.md
@@ -20,7 +20,7 @@ We considered the following options:
 
 * Other ORMs : Waterline, sequelize, bookshelf - General purpose ORMs were rejected based on the experience that the time spent going up the learning curve and sifting through the docs when things go wrong is greater then simply rolling your own.
 
-* Query Builders : massive.js, Knex.js -These were not entirely ruled out, and were deemed compatible with a good code structure that would allow the database access to be encapsulated in a way that would allow a query builder to be used where necessary.
+* Query Builders : Knex.js -These were not entirely ruled out, and were deemed compatible with a good code structure that would allow the database access to be encapsulated in a way that would allow a query builder to be used where necessary.
 
 ## Decision
 
@@ -31,7 +31,7 @@ In summary the server-side code will be structured below the `server` folder as 
 * **entities/\<entity\>** - The example below shows `Mansuscript` as an example entity, this folder will contain a common set of files that describe the entity's behaviour. Each of these have a particular purpose explained below:
 
   * index.js - The main business logic, e.g. for `Manuscript` this could contain `getAuthor()`
-  * typedefs.js - Contains the GraphQL types pertinant to this entity.
+  * typedefs.js - Contains the GraphQL types pertinent to this entity.
   * resolvers.js - Contains the GraphQL interface specified in terms of `Query` and `Mutation` 's for this entity. These in general should map onto the exposed functions from index.js. For example, `Manuscript` may contain a `Query` for `allManuscripts` and a `Mutation` for `createManuscript`.
   * data-access.js - Contains the functions for saving to and loading from the database. All references to the database and SQL that is written should remain private to this file.
 
--- a/docs/adr/0004-code-structure.md
+++ b/docs/adr/0004-code-structure.md
@@ -1,0 +1,60 @@
+# 4. Code Structure
+
+Date: 2018-06-14
+
+## Status
+
+Proposed
+
+## Context
+
+As the code rapidly grows we need a common structure and approach that will keep the code tidy and coherent.
+There are various aspects, the initial concern was the data access separation. The issue #140 initially talked about how this could be separated in terms of an ORM (Object-Relational Mapping) or a DAL (Data Access Layer).
+However, this discussion also soon covered the structure of the code in general and how to structure it for ease of development and best practise.
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


g-sam:
Another important consequence is that we are not making any plans to structure things in such a way as to make them easily exportable as PubSweet components.

tamlyn:
I think this would be more valuable if it also documented the options that were considered and rejected. Otherwise it's just a description of the file structure, something which is easier to understand from looking at the actual file structure.

jsms90:
Could we change the title of the PR to something like "ADR on ORM vs DAL" or "ADR on choices of DAL" or whatever was discussed? Having a descriptive title is really useful when you're looking through a list of PRs

Then perhaps a link to the relevant issue within the body of the PR? It would be nice to be able to go directly to the issue where the conversation happened & right now you would have to copy/paste what's in the title and change it to the URL format.

What do you reckon? @hdrury1 we could have a PR template too, if you like?

diversemix:
I tried to add the options we rejected in a condensed way - let me know if this is too summarised :)

dryhten:
What defines an entity? Is it a main type from the data model we agreed on here? https://github.com/elifesciences/elife-xpub/blob/develop/server/schema/schema.graphqls

If so, there are multiple mutations/queries that could deal with several entities at the same time. Where do we define those?

dryhten:
Will each `index.js` from an entity directory have validations (e.g. schema) for the object we use in our business logic on the backend?

diversemix:
@seanwiseman  .... feel free to contribute too.

diversemix:
@dryhten  ... good questions and we did talk about this but I didn't want to include this in the ADR (perhaps I should) ....

The entities are those defined in the `schema.graphqls` file 
Though Mutations may effect / involve many entities there should be one entity that naturally contains the logic for the orchestration. So these definitions belong in the entity folder belonging to the entity that is responsible for performing that mutation.... 

Generally what the ADR lays out a more of a set of principles rather than being prescriptive - we all agreed that this may "flex" as we begin to use it ;)

Maybe you could think of an example?  ... @tamlyn I can't seem to recall the ones we talked about.

g-sam:
Submitting a manuscript may involve saving it (~manuscript entity) and sending an email (~user entity).  Stuff like this is probably usually going to end up in manuscript I guess.

hdrury1:
@jsms90 There already is a PR template :)

tamlyn:
> What defines an entity? 

**I think**: An entity maps to a database table. Note that not all graphql types have their own tables – only types that need to be individually addressable by ID do. For instance reviewer suggestions would be a JSON array on manuscript. In general I would favour storing as much as possible as objects/arrays rather than separate tables.

diversemix:
> In general I would favour storing as much as possible as objects/arrays rather than separate tables.

@tamlyn Is there any reason for this? I would favour the reverse for reasons of being able to index on fields and also to reduce storage size (not that its a great concern at the minute).

hdrury1:
Oops. You may need to change to 0005 now ;)

tamlyn:
Simpler queries and less need for object-relational mapping as the storage format is closer to the object. 

You can index on anything in postgres including arrays, JSON columns, fields within JSON columns and even computed properties. The downsides I see are that there's no built in way to validate the contents of a JSON field, and it's less flexible so if you later want to address some part of a JSON field by ID you will probably need to change the schema to put that data into its own table.

Working with MongoDB I got a taste for storing objects instead of normalising everything into separate tables. But I also felt the pain of having no transactions or joins. I'm hopeful Postgres can give us the best of both.

docs/adr/0004-code-structure.md
```diff
@@ -0,0 +1,70 @@
...
+## Decision
+
+In summary the server-side code will be structured below the `server` folder as follows:
+
+* **entities** - This folder will contain subfolders relating to named entities in the system. See ["Data Model Specification"](https://docs.google.com/document/d/1KU-DLMNhPxjQF2j8HVlJvenvttPLxgtbTHo8Sy_PNRc/).
+
+* **entities/\<entity\>** - The example below shows `Mansuscript` as an example entity, this folder will contain a common set of files that describe the entity's behaviour. Each of these have a particular purpose explained below:
+
+  * index.js - The main business logic, e.g. for `Manuscript` this could contain `getAuthor()`
+  * typedefs.js - Contains the GraphQL types pertinant to this entity.
```
tamlyn: pertinent

docs/adr/0004-code-structure.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Rejected Options
+
+We considered the following options:
+
+* The Pubsweet Way - This was to define a base class "Model" and extend this for each entity. This was rejected as its another self made ORM but and has validations that we don't want.
+
+* Other ORMs : Waterline, sequelize, bookshelf - General purpose ORMs were rejected based on the experience that the time spent going up the learning curve and sifting through the docs when things go wrong is greater then simply rolling your own.
+
+* Query Builders : massive.js, Knex.js -These were not entirely ruled out, and were deemed compatible with a good code structure that would allow the database access to be encapsulated in a way that would allow a query builder to be used where necessary.
```
tamlyn: Knex, yes, but Massive is not a query builder and it would not be easy to use it piecemeal.

docs/adr/0004-code-structure.md
diversemix: Better get the spell check extension for vscode installed!

PR #207, opened 2018-06-15
Create Contributing.md for team principles

hdrury1:
Adding team working principles ADR


docs/adr/0004-principles.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Decision
+
+* We will prioritise reviewing Pull Requests (PRs) over beginning new issues.
+* Where possible, issues will include a user story and product requirements/acceptance criteria as well as links to relevant designs.
+* Where possible, issues will be grouped together according to the product feature they relate to.
+* After two days' effort, an assignee will add a comment to an issue and/or consider raising new issues to clearly indicate progress.
+* When feedback is required, an assignee will open a PR with the 'WIP' prefix. See [this ADR](https://github.com/elifesciences/elife-xpub/blob/develop/docs/adr/0001-pull-request-workflow.md).
+* As we facilitate deployment on a per-branch basis, once a PR is merged it is considered 'Done'.
+* Further end-to-end testing will be raised in a new issue with the 'e2e test' label.
```
tamlyn: If a feature can naturally be tested as part of an existing end-to-end test flow, then I think it should usually be done in the same issue.

docs/adr/0004-principles.md
```diff
@@ -0,0 +1,30 @@
...
+* An issue will never be re-opened.
```
tamlyn: Unless closed in error?

hdrury1:
My reasoning behind capturing these principles in an ADR is that they affect contruction techniques (in the same way we have documented the use of PRs).

hdrury1:
Okay folks. Attempt number 2 deletes the ADR and introduces a Contributing.md to the root.

CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
+# Contributing to xPub
+
+Date Last Updated: 15.06.18
```
tamlyn: I'll let git be the judge of that!

Issue #154, opened 2018-05-29
Send verification email to author


tamlyn:
A verification email should be sent to the corresponding author if they are not the submitter.

According the current designs (see #134) this should happen when the submitter clicks **Next** on the first page of the submission wizard *and* the email in the form is not their own.

tamlyn:
I'm not sure on the best approach for implementing this. Once #151 is implemented, `updateSubmission` will be getting called on every field blur. This means the sever no longer knows when **Next** is being clicked. Some options:

- create a new mutation which gets called only when **Next** is clicked
- add a flag to the `updateSubmission` mutation to signal what kind of update it is (autosave vs click)

hdrury1:
Invision Prototype for this step: https://invis.io/FUJPA5J4VMG#/298836915_default

jsms90:
@hdrury1 This invision prototype seems to have the following flow:
- User clicks 'Pre-fill'
- User clicks next
- Verification email has been sent

But [the doc in our google drive](https://docs.google.com/document/d/1cK158VrGlYRVZfA2vJrMNrUY6QYvtdC-jo0-SUvtBN4/edit) & @tamlyn's comment above both seem to suggest that the ribbon only appears if the user changes the what is in the email field.

Is the ribbon supposed to appear no matter what a user does before clicking the Next button?

jsms90:
Ah sorry. @tamlyn just showed me how to access the comments & the product requirements are in there :+1:

tamlyn:
Thoughts on possible implementation:
- Assume that the submitters email in xPub is always up to date with their email in ORCID (mechanism TBD)
- Add a flag `isAutoSave` to the `updateSubmission` mutation
- Add a `authorVerificationEmails` property to the submission
- When `updateSubmission` is called with `isAutoSave` false, compare the author email to the submitting user's email
- If the emails differ, check if  there is a pending or confirmed verification email for the author email
- If there is, do nothing
- If there isn't 
  - generate a token
  - send a verification email
  - store a pending verification email entry in `authorVerificationEmails`
  - revoke any existing pending (and confirmed?) verification email entries

diversemix:
@hdrury1 I question whether we want to implement the email bridge (last bullet above) as a part of this ticket... seems a biggy of itself.

tamlyn:
Good point.

Actually, I hadn't seen that as being the email bridge. It was just a link that they clicked to accept that they are the corresponding author. Like any email signup verification.

[...]

PR #209, opened 2018-06-15
feat: send verification email to corresponding author

tamlyn:
#### Background

See https://github.com/elifesciences/elife-xpub/issues/154#issuecomment-396306126

Integration/unit tests with Jest. No end-to-end tests.


dryhten:
Rebase?

jsms90:
WIP waiting for @tamlyn's return

tamlyn:
Does it need to be rebased before it can be reviewed?

jsms90:
@tamlyn hahahahaha. Sorry! Just going on what @dryhten said at standup :woman_shrugging: 
I'll un-WIP you :wink:

dryhten:
My comment to rebase was before you did it and you had merge conflicts, I haven't checked this since, I'll do it now.

server/submission/definitions.js
```diff
@@ -31,7 +31,9 @@ const typeDefs = `
       noConflictOfInterest: Boolean
       files: [File]
       submissionMeta: SubmissionMeta
+      manuscriptPersons: [ManuscriptPerson!]!
```
dryhten: This needs to be tested. As far as I know the frontend logic will update the frontend manuscript (the one used in Formik) with the values it gets from the backend and later this object will be sent back to backend containing `manuscriptPersons`. I need to check that this is indeed the case and that we don't get an error for having extra stuff in the input type.

server/submission/resolvers.js
```diff
@@ -65,15 +110,24 @@ const resolvers = {
       manuscript.id = await db.save(manuscriptDb)
       return manuscript
     },
-    async updateSubmission(_, { data }, ctx) {
-      const manuscript = await db.selectId(data.id)
+
+    async updateSubmission(_, { data, isAutoSave }, ctx) {
+      let manuscript = await db.selectId(data.id)
```
dryhten: Do you need `let` here? `manuscript` (the object) won't change with `lodash.merge`, but the value will (e.g. the props on it).

tamlyn: See line 124.

server/submission/definitions.js
tamlyn: This field is not queried for on the front-end therefore it doesn't appear in the form data.

dryhten: Right, so when the `updateSubmission` mutation runs from the frontend, the fields we specify that we want back won't include `manuscriptPersons`. Makes sense.

server/submission/resolvers.js
dryhten: Ok

PR #214, opened 2018-06-15
Fix the cleaning of node_modules to remove the one in server/submission

diversemix:
Fixes cleaning ... tested locally...
```
npm run clean
yarn
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "scripts": {
     "build": "NODE_ENV=production pubsweet build",
-    "clean": "rm -rf node_modules",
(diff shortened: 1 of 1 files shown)


tamlyn:
Actually any of the local packages can potentially contain `node_modules` at the whim of the dependency resolution algorithm. 

Locally I just use `find . -type directory -name node_modules -exec rm -rf \{\} \;`

dryhten:
I agree with Tam, `rm -rf server/submission/node_modules` seems too hardcoded.

diversemix:
Let's practise continuous improvement .... I'm going to merge this (as this fixes something that is currently not working) ... then you can improve it.

PR #231, opened 2018-06-19
195 subject area dropdown tests

### LINKED ISSUE #195: Test SubjectAreaDropdown component

jsms90:
####  What does this PR do?
- simplest possible unit test - checking whether a few elements appear on the page when shallow rendered
- integration tests (SubjectAreaDropdown + component from react-select library)

Fixes #195 

#### Not tested
Validation messages - responsibility of `ValidatedField` component & overall page component

Struggling with `react-select` + enzyme. Cannot simulate click behaviour, which makes it difficult to test the following:
- When 2 options are selected -> disables the dropdown
- Given that 2 options are selected, when 1 option is deleted, then the dropdown is re-enabled

### DIFF
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -66,10 +66,6 @@ describe('SubjectAreaDropdown component', () => {
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
       expect(wrapper.find('MultiValue')).toHaveLength(2)
       expect(wrapper.find('ValueContainer').html()).toContain('Cancer Biology')
-
-      // selectWrapper.props().onInputChange('E')
-      // selectInput.simulate('change')
-      // expect(wrapper.contains('Menu')).toBe(false)
     })
 
     it('selecting 2 options disables further selection', () => {
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -20,11 +20,9 @@ const MyDropdown = (
 describe('SubjectAreaDropdown component', () => {
   it('renders label & select', () => {
     const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
-    expect(wrapper.contains('My label')).toBe(true)
-    expect(wrapper.find('[inputId="subject-area-select"]').exists()).toEqual(
-      true,
-    )
-    expect(wrapper.contains('No more than two subject areas')).toBe(false)
+    expect(wrapper.html()).toContain('My label')
+    expect(wrapper.find('[inputId="subject-area-select"]').exists()).toBe(true)
+    expect(wrapper.text()).not.toContain('No more than two subject areas')
   })
 
   describe('Integration with react-select', () => {
@@ -46,7 +44,7 @@ describe('SubjectAreaDropdown component', () => {
 
     it('focusing causes the dropdown to open & display every option', () => {
       typeSubjectArea('') // typing nothing into the field seems to be the only way to programmatically focus react-select)
-      expect(wrapper.find('Menu').exists()).toEqual(true)
+      expect(wrapper.find('Menu').exists()).toBe(true)
       expect(wrapper.find('Option')).toHaveLength(17)
     })
 
@@ -54,13 +52,11 @@ describe('SubjectAreaDropdown component', () => {
       typeSubjectArea('b')
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
       expect(onChange.mock.calls).toHaveLength(1)
-      expect(wrapper.find('Menu').exists()).toEqual(false)
+      expect(wrapper.find('Menu').exists()).toBe(false)
       expect(wrapper.find('MultiValue')).toHaveLength(1) // MultiValue = a "tag" (i.e. selected option) in react-select
-      expect(
-        wrapper
-          .find('MultiValue')
-          .contains('Biochemistry and Chemical Biology'),
-      ).toBe(true)
+      expect(wrapper.find('ValueContainer').html()).toContain(
+        'Biochemistry and Chemical Biology',
+      )
     })
 
     it('2 options can be selected from the dropdown', () => {
@@ -69,7 +65,7 @@ describe('SubjectAreaDropdown component', () => {
       typeSubjectArea('c')
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
       expect(wrapper.find('MultiValue')).toHaveLength(2)
-      expect(wrapper.find('MultiValue').contains('Cancer Biology')).toBe(true)
+      expect(wrapper.find('ValueContainer').html()).toContain('Cancer Biology')
 
       // selectWrapper.props().onInputChange('E')
       // selectInput.simulate('change')
@@ -81,7 +77,7 @@ describe('SubjectAreaDropdown component', () => {
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
       typeSubjectArea('c')
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
-      expect(wrapper.contains('DropdownIndicator')).toBe(false)
+      expect(wrapper.text()).not.toContain('DropdownIndicator')
     })
 
     it('selection/deselection of 2 items triggers success message to show/hide', () => {
@@ -89,12 +85,12 @@ describe('SubjectAreaDropdown component', () => {
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
       typeSubjectArea('c')
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
-      expect(wrapper.contains('No more than two subject areas')).toBe(true)
+      expect(wrapper.text()).toContain('No more than two subject areas')
 
       const crossOnFirstTag = wrapper.find('MultiValueRemove').at(0)
       crossOnFirstTag.simulate('click')
       expect(wrapper.find('MultiValue')).toHaveLength(1)
-      expect(wrapper.contains('No more than two subject areas')).toBe(false)
+      expect(wrapper.text()).not.toContain('No more than two subject areas')
     })
 
     it('typing filters list of options based on first letter(s)', () => {
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -19,7 +19,8 @@ const MyDropdown = (
 
 describe('SubjectAreaDropdown component', () => {
   it('renders label & select', () => {
-    const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
+    // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
+    const wrapper = shallow(MyDropdown).dive()
     expect(wrapper.html()).toContain('My label')
     expect(wrapper.find('[inputId="subject-area-select"]').exists()).toBe(true)
     expect(wrapper.text()).not.toContain('No more than two subject areas')
@@ -43,7 +44,8 @@ describe('SubjectAreaDropdown component', () => {
     }
 
     it('focusing causes the dropdown to open & display every option', () => {
-      typeSubjectArea('') // typing nothing into the field seems to be the only way to programmatically focus react-select)
+      // typing nothing into the field seems to be the only way to programmatically focus react-select)
+      typeSubjectArea('')
       expect(wrapper.find('Menu').exists()).toBe(true)
       expect(wrapper.find('Option')).toHaveLength(17)
     })
@@ -53,7 +55,8 @@ describe('SubjectAreaDropdown component', () => {
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
       expect(onChange.mock.calls).toHaveLength(1)
       expect(wrapper.find('Menu').exists()).toBe(false)
-      expect(wrapper.find('MultiValue')).toHaveLength(1) // MultiValue = a "tag" (i.e. selected option) in react-select
+      // MultiValue = a "tag" (i.e. selected option) in react-select
+      expect(wrapper.find('MultiValue')).toHaveLength(1)
       expect(wrapper.find('ValueContainer').html()).toContain(
         'Biochemistry and Chemical Biology',
       )
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -4,15 +4,15 @@ import theme from '@elifesciences/elife-theme'
 
 import SubjectAreaDropdown from './SubjectAreaDropdown'
 
-const onChange = jest.fn()
-const onBlur = jest.fn()
+const handleChange = jest.fn()
+const handleBlur = jest.fn()
 
 const MyDropdown = (
   <SubjectAreaDropdown
     label="My label"
     name="My name"
-    onBlur={onBlur}
-    onChange={onChange}
+    onBlur={handleBlur}
+    onChange={handleChange}
     savedValues={[]}
     theme={theme}
   />
@@ -55,7 +55,7 @@ describe('SubjectAreaDropdown component', () => {
     it('an option can be selected from the dropdown', () => {
       typeSubjectArea('b')
       selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
-      expect(onChange.mock.calls).toHaveLength(1)
+      expect(handleChange.mock.calls).toHaveLength(1)
       // MultiValue = a "tag" (i.e. selected option) in react-select
       expect(wrapper.find('MultiValue')).toHaveLength(1)
       expect(wrapper.find('ValueContainer').html()).toContain(
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -5,12 +5,13 @@ import theme from '@elifesciences/elife-theme'
 import SubjectAreaDropdown from './SubjectAreaDropdown'
 
 const onChange = jest.fn()
+const onBlur = jest.fn()
 
 const MyDropdown = (
   <SubjectAreaDropdown
     label="My label"
     name="My name"
-    onBlur={jest.fn()}
+    onBlur={onBlur}
     onChange={onChange}
     savedValues={[]}
     theme={theme}
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import { shallow, mount } from 'enzyme'
-import theme from '@pubsweet/elife-theme'
+import theme from '@elifesciences/elife-theme'
 
 import SubjectAreaDropdown from './SubjectAreaDropdown'
 
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { shallow } from 'enzyme'
+import { shallow, mount } from 'enzyme'
 import theme from '@pubsweet/elife-theme'
 
 import SubjectAreaDropdown from './SubjectAreaDropdown'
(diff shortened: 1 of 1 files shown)


app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+  />
+)
+
+describe('SubjectAreaDropdown component', () => {
+  it('renders label & select', () => {
+    const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
+    expect(wrapper.contains('My label')).toBe(true)
+    expect(wrapper.find('[inputId="subject-area-select"]').exists()).toEqual(
+      true,
+    )
```
tamlyn: Use `toBe` for consistency.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+    onChange={onChange}
+    savedValues={[]}
+    theme={theme}
+  />
+)
+
+describe('SubjectAreaDropdown component', () => {
+  it('renders label & select', () => {
+    const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
+    expect(wrapper.contains('My label')).toBe(true)
```
tamlyn: It's useful to think about the error message that an assertion generates when it fails. For example rewriting this as `expect(wrapper.text()).toContain('My label')` would show you the actual text content on failure which could be helpful.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+      expect(wrapper.contains('No more than two subject areas')).toBe(false)
+    })
+
+    it('typing filters list of options based on first letter(s)', () => {
+      typeSubjectArea('c')
+      const totalStartingWithC = wrapper.find('Option').length
+      expect(totalStartingWithC).toBeLessThan(17)
+      wrapper.find('Option').forEach(option => {
+        const optionText = option.childAt(0).text()
+        const wordStartsWithC = new RegExp(/^[cC]\w+/)
```
tamlyn: You can simplify this to `const wordStartsWithC = /^c/i`

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+    it('focusing causes the dropdown to open & display every option', () => {
+      typeSubjectArea('') // typing nothing into the field seems to be the only way to programmatically focus react-select)
+      expect(wrapper.find('Menu').exists()).toEqual(true)
+      expect(wrapper.find('Option')).toHaveLength(17)
+    })
+
+    it('an option can be selected from the dropdown', () => {
+      typeSubjectArea('b')
+      selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
+      expect(onChange.mock.calls).toHaveLength(1)
```
tamlyn: You should probably clear this mock in a `beforeEach`. Otherwise this test could pass in error if a previous test calls `onChange` and this one doesn't.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+    savedValues={[]}
+    theme={theme}
+  />
+)
+
+describe('SubjectAreaDropdown component', () => {
+  it('renders label & select', () => {
+    const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
+    expect(wrapper.contains('My label')).toBe(true)
+    expect(wrapper.find('[inputId="subject-area-select"]').exists()).toEqual(
```
dryhten: What are you testing here? I would see this as a test to see if the SubjectAreaDropdown has an element inside with id `subject-area-select` (implementation specific). I'm not a testing expert, but I usually test behavioral stuff (e.g. given these props, the component displays this text).

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+    onBlur={jest.fn()}
+    onChange={onChange}
+    savedValues={[]}
+    theme={theme}
+  />
+)
+
+describe('SubjectAreaDropdown component', () => {
+  it('renders label & select', () => {
+    const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
```
dryhten: Long line comments anyone? Not a thing in javascript? :)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+      // 1. Add a letter to the value of the input field (does not cause the dropdown to appear by itself)
+      selectWrapper.props().onInputChange(letters)
+      // 2. Open the dropdown
+      selectInput.simulate('change')
+    }
+
+    it('focusing causes the dropdown to open & display every option', () => {
+      typeSubjectArea('') // typing nothing into the field seems to be the only way to programmatically focus react-select)
+      expect(wrapper.find('Menu').exists()).toEqual(true)
+      expect(wrapper.find('Option')).toHaveLength(17)
```
dryhten: Hm, since you have `eLifeOptions` as a global variable in `SubjectAreaDropdown.js`, maybe you can export it from there and not hardcode 17 here?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+  it('renders label & select', () => {
+    const wrapper = shallow(MyDropdown).dive() // dive is necessary because SubjectAreaDropdown is wrapped by withTheme
+    expect(wrapper.contains('My label')).toBe(true)
+    expect(wrapper.find('[inputId="subject-area-select"]').exists()).toEqual(
+      true,
+    )
+    expect(wrapper.contains('No more than two subject areas')).toBe(false)
+  })
+
+  describe('Integration with react-select', () => {
```
dryhten: Do we care about `react-select` at test level?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+      // We need 2 steps to programmatically simulate typing in a letter
+      // 1. Add a letter to the value of the input field (does not cause the dropdown to appear by itself)
+      selectWrapper.props().onInputChange(letters)
+      // 2. Open the dropdown
+      selectInput.simulate('change')
+    }
+
+    it('focusing causes the dropdown to open & display every option', () => {
+      typeSubjectArea('') // typing nothing into the field seems to be the only way to programmatically focus react-select)
+      expect(wrapper.find('Menu').exists()).toEqual(true)
```
dryhten: `toBe` :)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+      expect(wrapper.find('Option')).toHaveLength(17)
+    })
+
+    it('an option can be selected from the dropdown', () => {
+      typeSubjectArea('b')
+      selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
+      expect(onChange.mock.calls).toHaveLength(1)
+      expect(wrapper.find('Menu').exists()).toEqual(false)
```
dryhten: Same as above

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+    it('2 options can be selected from the dropdown', () => {
+      typeSubjectArea('b')
+      selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
+      typeSubjectArea('c')
+      selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
+      expect(wrapper.find('MultiValue')).toHaveLength(2)
+      expect(wrapper.find('MultiValue').contains('Cancer Biology')).toBe(true)
+
+      // selectWrapper.props().onInputChange('E')
```
dryhten: Commented code?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+      ).toBe(true)
+    })
+
+    it('2 options can be selected from the dropdown', () => {
+      typeSubjectArea('b')
+      selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
+      typeSubjectArea('c')
+      selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
+      expect(wrapper.find('MultiValue')).toHaveLength(2)
+      expect(wrapper.find('MultiValue').contains('Cancer Biology')).toBe(true)
```
dryhten: Hm, this seems a bit random, you only check for `Cancer Biology` and not `Biochemistry and Chemical Biology`. You hate `Biochemistry and Chemical Biology` that much?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+      crossOnFirstTag.simulate('click')
+      expect(wrapper.find('MultiValue')).toHaveLength(1)
+      expect(wrapper.contains('No more than two subject areas')).toBe(false)
+    })
+
+    it('typing filters list of options based on first letter(s)', () => {
+      typeSubjectArea('c')
+      const totalStartingWithC = wrapper.find('Option').length
+      expect(totalStartingWithC).toBeLessThan(17)
+      wrapper.find('Option').forEach(option => {
```
dryhten: Why do you do this twice?
```
const wordsStartingWithC = wrapper.find('Option')
expect(wordsStartingWithC.length).toBeLessThan(17);
wordsStartingWithC.forEach(...)
```
Also, do we really need to assert that the length is less than the total number of options? It's really hardcoded, just checking that all the words in this list start with `c` or that this list is the same as the list of words starting with `c` from `eLifeOptions` (I guess building this second one here would work) seems enough.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+        const optionText = option.childAt(0).text()
+        const wordStartsWithC = new RegExp(/^[cC]\w+/)
+        expect(optionText).toMatch(wordStartsWithC)
+      })
+
+      typeSubjectArea('ce')
+      const totalStartingWithCe = wrapper.find('Option').length
+      expect(totalStartingWithCe).toBeLessThan(totalStartingWithC)
```
dryhten: Again, what if all options starting with `c` will change to option starting with `ce`?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: The whole test is basically 'does this render' correctly? You could use a snapshot test for this. What I've done is say
- is there a label present?
- is there an input field present?

For the second bullet point, just plain ol' `wrapper.find('input')` would be enough. Check if there's an input at all.

But the way `react-select` works under the hood, when you select an option from the dropdown, it creates hidden input fields. So I guess I was just going specific, to see if there was _this specific_ input field present, since I've hard coded `id=subject-area-dropdown` into SubjectAreaSelect

jsms90: ...probably unnecessary though :thinking: We'll never have the hidden inputs present, & the main input field _not_ present....

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: "Not a thing"?

Are you just requesting that I put the comment on the line above? Or make the comment shorter? I'm not sure I understand what changes you're requesting?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: I figured exporting `eLifeOptions` would only be a temporary solution, if we're going to call the eLife API at some point.

But yeah, either way, I probably shouldn't hard code 17 into here.

Do you have a suggestion for how I could test that all of the options are rendered, rather than just a subset (e.g. when it's filtered by a given letter), even when the list is dynamically generated?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: why not?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: I need to call `wrapper.find('Option')` twice, because the number of `Option`s on the page changes after I simulate a new letter being typed. So this duplication is necessary:
```
const totalStartingWithC = wrapper.find('Option').length
// simulate click event
const totalStartingWithCe = wrapper.find('Option').length
```

If I wanted the variable name to stay the same I could have done this:
```
let totalOptions

totalOptions = wrapper.find('Option').length
// simulate click event
totalOptions = wrapper.find('Option').length
```

but I specifically wanted the 2 different lengths to be stored separately, so I could use them both in this assertion:
```
expect(totalStartingWithCe).toBeLessThan(totalStartingWithC)
```

But you're totally right. The `toBeLessThan` test is actually totally redundant, since I'm checking whether they start with the appropriate letters anyway. Thanks :+1: 


As for the rest, the assertions are not identical, so this isn't really duplication:
```
// 1st
expect(totalStartingWithC).toBeLessThan(17)
// 2nd
expect(totalStartingWithCe).toBeLessThan(totalStartingWithC)
```

and the `wrapper.find('Option').forEach(option => .....` is definitely duplication & could have been abstracted into a function. Seemed a bit overkill, but probably not & probably would have been far more readable :+1:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: If I'm understanding what you're asking correctly, the `totalStartingWithCe` variable doesn't update, when the number of Options in the "HTML" changes. So there's no danger of the "change" you're referring to.

It's a bit annoying actually, to repeat yourself. But if you want to find the new number of `Option`s on the page, you have to call `wrapper.find` again.

Does that answer your question? I'm not sure if I understood you.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: Ah thanks! I hadn't clocked the effect that this has on the messages :sweat_smile: 
Makes sense :smile:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
...
+import SubjectAreaDropdown from './SubjectAreaDropdown'
+
+const onChange = jest.fn()
+
+const MyDropdown = (
+  <SubjectAreaDropdown
+    label="My label"
+    name="My name"
+    onBlur={jest.fn()}
+    onChange={onChange}
```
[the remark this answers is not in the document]
jsms90: Oops, didn't mean to leave one inline and one outside. I should change it to:
```
const onChange = jest.fn()
const onBlur = jest.fn()

const MyDropdown = (
  <SubjectAreaDropdown
    label="My label"
    name="My name"
    onBlur={onBlur}
    onChange={onChange}
    savedValues={[]}
    theme={theme}
  />
)
```
Right?

Or are you saying that using `jest.fn` twice wrong? What should I be doing?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: I check whether it contains `Biochemistry...` [when I'm testing whether 'an option can be selected from the dropdown'](https://github.com/elifesciences/elife-xpub/blob/c8bfffb3442edb859a5101bec59824006882d48b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js#L59-L63)

I guess I figured that I only needed to check that the 2nd one had been added, without asserting that `Biology...` was still there (since I assert that it's there just above).

I figured it was safe to assume that selecting a second option wasn't going to cause the first tag to disappear. Do you think that's an unsafe assumption?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react'
+import { shallow, mount } from 'enzyme'
+import theme from '@pubsweet/elife-theme'
```
dryhten: Change to `@elifesciences/elife-theme`

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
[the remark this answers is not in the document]
dryhten: Sorry, my bad, ignore this comment, I didn't realize you're using `onChange` somewhere else.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: Probably a snapshot test would be better? Like I said, I wouldn't test for this. I would just try to test functionality, but this is not a `request changes` comment, just an open discussion, so feel free to keep it like this if you think it makes sense :)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: So in most programming languages there is a rule to limit the number of characters per line to 80. If you want to read more about it
https://en.wikipedia.org/wiki/Characters_per_line
https://softwareengineering.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width

I think eslint/prettier should try to split lines longer than 80 characters, not sure what happened here.

Limiting the line to 80 characters seems like a burden when using HTML for example, which is hard to split on multiple lines. That's why webdev seems more relaxed about this rule. However, personally I prefer lines shorter than 80 characters, it's a pain to watch the line wrap itself on multiple "lines" in my editor.

I would personally move the comment up a line, then if it's still longer than 80 characters I would split it into multiple lines like so
```
/**
* line 1
* line 2
* ...
*/
```

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: Again, implementation detail. I don't care that the SubjectAreaDropdown is using `react-select`, I just care if it works or not.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: Not really, if this is going to change I assume it's fine for now (updating the component usually results in updating the tests as well). I would probably try to query the eLife API here once we do this in the component itself.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: I usually try to write unit tests as independent as possible so I would probably test for the second one too without assuming stuff. Again, open discussion :)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: That's not what I meant, sorry. I meant that you call `wrapper.find()` twice in a row for no reason (it's a function). It strictly refers to lines 102, 103 and 104. Look at what I was suggesting to use: call `wrapper.find()` once and store the result in one variable that you can use for both checks you do for letter `c`.

I was also suggesting that taking the `eLifeOptions` list that you defined in this file, filter it to have only words that start with `c` (writing the code for this) in this test, and then comparing with the output of `SubjectAreaDropdown` might be a good test.

Similar for `ce`.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: No, I meant what if eLife will change the number of options starting with `c` and `ce` in the future so they are equal? Is this a relevant test? I wouldn't think so.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: I see. So you're requesting that I move the comment to the line above? No problem.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: True, it is an implementation detail. But every test that's written here assumes an HTML structure that is specific to `react-select`. When testing the integration between components, I assume that specifying which components you're using is fine?

jsms90: Perhaps this wasn't your point? Are you actually commenting on the choice to write integration tests & not really do unit tests?

I've got to say, the lack of unit tests feels a little wrong to me too. But I asked about this when I first started working on testing for this component. It seemed to be a commonly held opinion that integration tests would add more value than unit tests & that unit tests may not even be worth writing. I'd be happy to raise a discussion issue, for defining a testing strategy for the frontend?

Regardless of whether we choose to add unit tests though, integration tests are very valuable. At the end of the day, you want to be sure that the unit-tested functionality of `react-select` combined with the unit tested functionality of SubjectAreaDropdown, give you the behaviour that you want to see.

Does this answer your point?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: Yeah I totally get where you're coming from. But these aren't really unit tests.

I think I've introduced confusion by separating it into so many little blocks in the first place. I originally had most of this in a single `it` block, but it just felt so weird to be writing one giant test for the whole component. My thinking was that, if this whole test fails, we won't know why. But I guess that comes back to the reason why you write unit tests in the first place.

Thinking about it again, it actually doesn't really add any value to separate out the tests that require this code to be repeated:
```
typeSubjectArea('b')
selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
typeSubjectArea('c')
selectInput.simulate('keyDown', { keyCode: 9, key: 'Tab' })
```

I'm going to suggest that I stick the first 5 tests back into a single `it` block? But it sounds like we really do need a broader discussion on whether to add unit tests (e.g. snapshot testing)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: Yeah that's exactly what I thought too! But it doesn't seem to work like that.

If you save it in a variable, e.g. `const totalStartingWithC = wrapper.find()`, then simulate a change....
`totalStartingWithC` will still be equal to the _old_ value (i.e. 4)

You need to actually called `wrapper.find()` again.

Go figure :woman_shrugging:

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: See comment above
> But you're totally right. The toBeLessThan test is actually totally redundant, since I'm checking whether they start with the appropriate letters anyway. Thanks +1

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: Yeah I agree. Seems like a good case for a snapshot because all I really want to say is "is this on the page?" & maybe "does it have this general form?" i.e. is there a label & an input field.

I can understand the resistance to snapshot testing, so I'm happy not to start including snapshots in our codebase. But I reckon having a test which checks that the component renders at all is a good thing & I reckon the first couple of assertions do that job here.

The 3rd assertion is probably misplaced :thinking: 
I guess there should be a test that sets the state (specifically `hasReachedMultiselectLimit=false`) & then checks whether there's a message? What do you think @dryhten?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
jsms90: I think I'm going to remove this test altogether. It's not really adding anything, that isn't already tested within this test: `an option can be selected from the dropdown`. If `react-select` is working as it should, then adding 1 & adding 2 are the same thing.

jsms90:
Made many of the requested changes from @tamlyn & @dryhten 

Updated the top comment to reflect what hasn't been tested (and why) & the naming of tests, for greater clarity.

Since my changes have collapsed the comments by @dryhten, I'll reiterate that there are outstanding questions about what unit testing should be done here.

Anything I've missed @dryhten?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: No, I was just saying I don't agree with how you named your describe section. I would name it differently.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.test.js
dryhten: I don't see what you don't understand from what I was saying. Is this line simulating a change and I don't know about it?
`expect(totalStartingWithC).toBeLessThan(17)`

Issue #215, opened 2018-06-19
Move elife-theme from pubsweet repo to elife-xpub (this) repo


dryhten:
Currently the elife-theme is located on the pubsweet repo here https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/elife-theme

It should be moved to this repo so we can change it faster. The best location for the theme is probably the client directory.

This issue encapsulates the following tasks

- [x] Do a PR with the elife-theme files to this repo https://github.com/elifesciences/elife-xpub/pull/216
- [x] Do a MR on pubsweet removing the elife-theme files [!278](https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/278)
- [x] Make sure we have the rights to publish new versions of the theme to the npm registry
- [x] PR to change the links with this repo to point to the new npm package location (https://www.npmjs.com/package/@elifesciences/elife-theme) #232 
~~Wait for PRs & MR to get merged~~

jsms90:
I've updated the top comment to reflect the fact that we have a 2nd PR open against this issue. Once this is merged, all the necessary changes will have been made for elife-xpub :+1: 

But this ticket will not be considered 'done' until the MR to pubsweet has been merged - so it's 'blocked' until Coko merges.

hdrury1:
Closing down this issue- work on our end is done.

PR #232, opened 2018-06-19
fix(elife-theme): use elifesciences org instead of pubsweet for theme

dryhten:
re #215

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -5,7 +5,7 @@ import { withClientState } from 'apollo-link-state'
 
 import { configureStore, Root } from 'pubsweet-client'
 
-import theme from '@pubsweet/elife-theme'
+import theme from '@elifesciences/elife-theme'
 
 import Routes from './routes'
 import * as AuthorDetailsSchema from './components/submission/AuthorDetails/AuthorDetailsSchema'
--- a/app/components/ui/molecules/FileUpload.test.js
+++ b/app/components/ui/molecules/FileUpload.test.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { mount } from 'enzyme'
 import { MemoryRouter } from 'react-router-dom'
 import { ThemeProvider } from 'styled-components'
-import theme from '@pubsweet/elife-theme'
+import theme from '@elifesciences/elife-theme'
 
 import FileUpload from './FileUpload'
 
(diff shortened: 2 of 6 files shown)


dryhten:
Note that `yarn.lock` didn't want to get updated and `CHANGELOG.md` for the theme should be updated as well.

jsms90:
Can't merge without a 2nd person :sweat:

dryhten:
@jsms90 hannah merged it by force (I don't know what method she used). I assume I can manually merge this commit too from the terminal, but I don't know if it's that urgent to have it. We should test locally before we publish new versions of the theme package anyway.

hdrury1:
Just wanted to point out that @elifesciences/elife-theme is the eLife 2.0 theme, rather than the eLife xPub theme and the two don't match in terms of variables. Is that an issue?

jsms90:
As discussed at standup, the readme is just out of date on https://www.npmjs.com/package/@elifesciences/elife-theme, whereas the code itself reflects the correct variables and values :+1:

This is really common, whenever a readme documents exactly what code that is written. The code gets updated, but the docs do not. Otherwise, every time you make an update to the code, you need to make an identical update in the documentation, which is just a bit of a waste. In the case of a theme, the code is self-documenting.

So, to avoid this kind of confusion in future, we're probably better off removing the code itself from the readme - see [my review for the other PR](https://github.com/elifesciences/elife-xpub/pull/233#pullrequestreview-130303406) :smile: 

Seems like this is ready to merge @hdrury1? And we can make the necessary updates in the other PR?

dryhten:
@hdrury1 as discussed at the standup, I'm not entirely sure what eLife xPub theme is, I've never seen or used it before, and couldn't find it in any repository. The only theme we ever used it the elife-theme that I pushed to this repository.

The [README.md](https://github.com/elifesciences/elife-xpub/blob/develop/client/elife-theme/README.md) file from the elife-theme seems to copy the SASS variables from here https://github.com/elifesciences/pattern-library/blob/develop/assets/sass/_variables.scss

We are using javascript variables now and not SASS ones. So the readme should be changed or removed as Jen mentioned.

@jsms90 just to make sure everything is clear, that node package you linked to is the same with this directory in this repository https://github.com/elifesciences/elife-xpub/tree/develop/client/elife-theme

PR #241, opened 2018-06-22
Replace vertical grid units with space

### LINKED ISSUE #236: vertical spacing variables in theme are broken
**Describe the bug**
Vertical spacing variables have been added to the theme in PR #233 but the values do not evaluate to `px` values - see [@tamlyn's comment on the PR](https://github.com/elifesciences/elife-xpub/pull/233#discussion_r197084939)

Caused by [this code](https://github.com/elifesciences/elife-xpub/blob/develop/client/elife-theme/src/index.js#L44-L50):
```
verticalXxl: 'gridUnit'*20,
verticalXl: 'gridUnit'*12,
verticalL: 'gridUnit'*8,
verticalM: 'gridUnit'*6,
verticalS: 'gridUnit'*4,
verticalXs: 'gridUnit'*2,
verticalXxs: 'gridUnit',
```

So we can't use these variables at the moment.

It might not be the best idea to expose 2 identical variables `verticalXxs` & `gridUnit` anyway?

#### Relevant tickets
- [design issue in pubsweet](https://gitlab.coko.foundation/pubsweet/design/issues/31)
- [original (closed) issue for updating theme] - #183 

#### Suggested solutions
1. How about using `gridUnit` as a measurement (`const` at the top of the file, but not exposed) & only exposing `vertical...`?  
This means that our variables are diverging from Coko's default theme though, in removing `gridUnit`.
2. Remove `verticalXxs`, `verticalXs`, etc altogether & use `gridUnit * 2` (for example), in the same way that other projects already do

---

Another potential concern is that we're using these "vertical spacing" variables for padding/margin left and right as well. So they're not just "vertical". Perhaps `gridUnit` is a better name?

We could have `gridUnitXxl` ... `gridUnitXxs`?
Note: this still means getting rid of `gridUnit` and replacing it with `gridUnitXxs`

jsms90:
#### What does this PR do?r
- replace them variables `verticalXxs`...`verticalXxl` with `space`
- remove `theme.space` from `app.js`, since it now already exists in the theme

#### Any relevant tickets
Fixes #236

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -10,16 +10,6 @@ import theme from '@elifesciences/elife-theme'
 import Routes from './routes'
 import * as AuthorDetailsSchema from './components/submission/AuthorDetails/AuthorDetailsSchema'
 
-const pixelGridUnit = parseInt(theme.gridUnit, 10)
-theme.space = [
-  0,
-  pixelGridUnit / 4,
-  pixelGridUnit / 2,
-  pixelGridUnit,
-  pixelGridUnit * 2,
-  pixelGridUnit * 4,
-]
-
 const history = createHistory()
 const store = configureStore(history, {})
 
--- a/client/elife-theme/src/index.js
+++ b/client/elife-theme/src/index.js
@@ -1,6 +1,8 @@
 /* eslint-disable import/extensions */
 import './fonts/index.css'
 
+const gridUnit = 6
+
 export default {
   /* Colors */
   colorPrimary: '#0288d1',
@@ -40,7 +42,17 @@ export default {
   fontLineHeight6: '16px',
 
   /* Spacing */
-  gridUnit: '6px',
+  gridUnit: `${gridUnit}px`,
+  space: [
+    0,
+    gridUnit,
+    gridUnit * 2,
+    gridUnit * 4,
+    gridUnit * 6,
+    gridUnit * 8,
+    gridUnit * 12,
+    gridUnit * 20,
+  ],
 
   /* Border */
   borderRadius: '3px',
--- a/client/elife-theme/src/index.js
+++ b/client/elife-theme/src/index.js
@@ -41,13 +41,6 @@ export default {
 
   /* Spacing */
   gridUnit: '6px',
-  verticalXxl: 'gridUnit' * 20,
-  verticalXl: 'gridUnit' * 12,
-  verticalL: 'gridUnit' * 8,
-  verticalM: 'gridUnit' * 6,
-  verticalS: 'gridUnit' * 4,
-  verticalXs: 'gridUnit' * 2,
-  verticalXxs: 'gridUnit',
 
   /* Border */
   borderRadius: '3px',


client/elife-theme/src/index.js
```diff
@@ -40,14 +42,15 @@ export default {
...
-  verticalXs: 'gridUnit'*2,
-  verticalXxs: 'gridUnit',
+  gridUnit: `${gridUnit}px`,
+  space: [
+    0,
+    gridUnit,
+    gridUnit * 2,
+    gridUnit * 4,
+    gridUnit * 12,
+    gridUnit * 16,
```
tamlyn: These multiples are different to those in the design style guide. Is that intentional?

jsms90: No, not at all :flushed:
Thanks :smiley:

app/app.js
```diff
@@ -10,16 +10,6 @@ import theme from '@elifesciences/elife-theme'
 import Routes from './routes'
 import * as AuthorDetailsSchema from './components/submission/AuthorDetails/AuthorDetailsSchema'
 
-const pixelGridUnit = parseInt(theme.gridUnit, 10)
-theme.space = [
```
dryhten: Why was this here in the first place? I can't see it being used anywhere.

client/elife-theme/src/index.js
```diff
@@ -1,6 +1,8 @@
 /* eslint-disable import/extensions */
 import './fonts/index.css'
 
+const gridUnit = 6
```
dryhten: Isn't there a better way to do this? I would assume that when `theme.space` is called somewhere in the code you have access to the entire theme object so you can defer the value of `theme.space` until execution (e.g. with a function), and then get `theme.gridUnit` somehow. I would have to test this, but I think something like ``css`
calc(${props => props.theme.gridUnit} * 2);` `` would work (with `css` from styled-components).

app/app.js
tamlyn: It's used but `grid-styled`.

client/elife-theme/src/index.js
tamlyn: Yes, that calculation works and is the expected way to use gridUnit. The `space` property is required by `grid-styled`.

jsms90: @dryhten @tamlyn So should I be removing `space` from the theme after all? Or are we leaving it in here so we can use `space[2]` in place of `verticalXs` (for example)?

I'm eager to get this merged in so that I can use the appropriate variable names while styling components :pray:

tamlyn: `space` is required in order to set `grid-styled` to use our spacing scale (it defaults to powers of 2 otherwise). Whether we use `space[2]` instead of `calc` doesn't affect this PR.

jsms90: ok :+1:

@dryhten happy to approve & merge?

g-sam: Last time I checked PubSweet's gridUnit was the 'average', not the smallest possible spacing. So if we do this our components wont be compatible.

dryhten: So deferring the execution inside `theme.space` seems like overly complicated at this point. I guess this is the simpler way of doing it.

tamlyn: Yes, but this is the new reality.

jsms90: Have you seen the latest eLife theme/styleguide in invision yet? https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926984

PR #242, opened 2018-06-22
fix(elife-theme): use 'boxShadow' instead of 'dropShadow'

### LINKED ISSUE #237: change theme variable "dropShadow" back to "boxShadow"
`dropShadow` is listed in `VARIABLES.md` & therefore also in eLife's style guide within invision
but `boxShadow` is the name used in the actual code within the themes

See [answer in pubsweet design repo issue](https://gitlab.coko.foundation/pubsweet/design/issues/29) and/or [comment in PR](https://github.com/elifesciences/elife-xpub/issues/217#issuecomment-399173032):
> boxShadow would make more sense as it's tight to the css property name

(`boxShadow` was changed to `dropShadow` in the code in PR #233 but the variable `boxShadow` is still being used within the codebase. I imagine this breaks things right now, but will probably be sorted within each of the styling issues for each component?)

jsms90:
#### What does this PR do?
- use `boxShadow` as theme variable name instead of `dropShadow`

#### Any relevant tickets
Fixes #237
Relates https://gitlab.coko.foundation/pubsweet/design/issues/29

### DIFF
--- a/client/elife-theme/src/index.js
+++ b/client/elife-theme/src/index.js
@@ -41,12 +41,12 @@ export default {
 
   /* Spacing */
   gridUnit: '6px',
-  verticalXxl: 'gridUnit'*20,
-  verticalXl: 'gridUnit'*12,
-  verticalL: 'gridUnit'*8,
-  verticalM: 'gridUnit'*6,
-  verticalS: 'gridUnit'*4,
-  verticalXs: 'gridUnit'*2,
+  verticalXxl: 'gridUnit' * 20,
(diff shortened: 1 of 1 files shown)


dryhten:
Should someone do a MR updating this? https://gitlab.coko.foundation/pubsweet/design/blob/master/VARIABLES.md

tamlyn:
I guess so. But I would vote for removing that document in favour of documenting the variables in the actual theme. That would avoid issues like this in the first place.

PR #243, opened 2018-06-22
style(ui): apply styles to Button component to conform with new styleguide

dryhten:
re #218

Tested locally by looking at the button on the page and comparing with the styleguide (checking sizes for font, paddings etc).


client/elife-theme/src/index.js
```diff
@@ -41,12 +42,12 @@ export default {
 
   /* Spacing */
   gridUnit: '6px',
-  verticalXxl: 'gridUnit'*20,
-  verticalXl: 'gridUnit'*12,
-  verticalL: 'gridUnit'*8,
-  verticalM: 'gridUnit'*6,
-  verticalS: 'gridUnit'*4,
-  verticalXs: 'gridUnit'*2,
+  verticalXxl: 'gridUnit' * 20,
```
dryhten: Funny how commitizen fixed the spacing for something that's wrong in the first place anyway :)

tamlyn: This is `prettier` and it only cares about syntax, not semantics.

client/elife-theme/src/components/Button.js
```diff
@@ -0,0 +1,33 @@
+import { css } from 'styled-components'
+
+const extraSmall = css`
+  font-size: 11px;
+  line-height: calc(${props => props.theme.gridUnit} * 2);
+  padding: ${props => props.theme.gridUnit};
```
tamlyn: Why not use `th`?

client/elife-theme/src/components/Button.js
```diff
@@ -0,0 +1,33 @@
...
+const small = css`
+  padding: ${props => props.theme.gridUnit};
+`
+
+const primary = css`
+  border: 0 none;
+`
+
+const reverse = css`
+  background: #262626;
```
tamlyn: Shouldn't this be `colorText`?

client/elife-theme/src/index.js
dryhten: I meant prettier and I'm aware it only checks syntax, I was just saying it's funny.

client/elife-theme/src/components/Button.js
```diff
@@ -0,0 +1,33 @@
...
+  color: ${props => props.theme.colorTextReverse};
+  border-color: ${props => props.theme.colorTextReverse};
+`
+
+export default css`
+  line-height: calc(${props => props.theme.gridUnit} * 4);
+  min-width: calc(${props => props.theme.gridUnit} * 16);
+  padding: calc(${props => props.theme.gridUnit} * 2);
+  font-size: 14px;
```
tamlyn: Should this be `fontSizeBaseSmall`?

client/elife-theme/src/components/Button.js
```diff
@@ -0,0 +1,33 @@
...
+const reverse = css`
+  background: #262626;
+  color: ${props => props.theme.colorTextReverse};
+  border-color: ${props => props.theme.colorTextReverse};
+`
+
+export default css`
+  line-height: calc(${props => props.theme.gridUnit} * 4);
+  min-width: calc(${props => props.theme.gridUnit} * 16);
+  padding: calc(${props => props.theme.gridUnit} * 2);
```
tamlyn: Add a comment that these three lines should be removed once the PubSweet button is updated to use the new smaller gridUnit.

client/elife-theme/src/components/index.js
```diff
@@ -0,0 +1 @@
+export { default as Button } from './Button'
```
tamlyn: Remind me why you think that index files like this are useful for private components?

client/elife-theme/src/components/Button.js
dryhten: I think I started from other component and completely forgot about `th`.

client/elife-theme/src/components/Button.js
dryhten: No, the reverse colors are not in the theme as far as I know. This is the reverse button in the invision file nick upload, which doesn't have any specification. I guess keeping this here for future use when we know all the details won't hurt.

client/elife-theme/src/components/Button.js
dryhten: No, check my discussion with Nick on invision (click "show resolved"). He mentioned that font sizes on buttons should be hardcoded.

client/elife-theme/src/components/Button.js
dryhten: I don't think so? From my understanding the PubSweet girdUnit will change to 8px, so they will have to be updated, but not remove. I can add the update comment.

client/elife-theme/src/components/index.js
dryhten: The first time I used something like this it seemed more maintainable in the theme `index.js` to import/remove imported components (e.g. instead of having a lot of different imports, you can only do one). `xpub-epmc` is doing the same thing, so it might be nice to have a similar style https://gitlab.coko.foundation/xpub/xpub-epmc/tree/dev/app/epmc-theme/elements

client/elife-theme/src/components/Button.js
tamlyn: Might be worth double checking with Nick as I thought the point of the reverse colours is that the text colour becomes the background and the background becomes the foreground. `colorText` is `#212121` which is indistinguishably close this this.

client/elife-theme/src/components/Button.js
tamlyn: My understanding is that when the PubSweet `Button` is updated, it will have these exact three lines. These will then be redundant as they will be duplicating rather than overriding.

client/elife-theme/src/components/Button.js
dryhten: I do have an unanswered question about this in invision.

client/elife-theme/src/components/Button.js
dryhten: If that's what's going to happen then yes. If they will use those lines then the buttons they have now will show bigger, I was expecting them to want to keep the same sizes with different gridUnit.

dryhten:
I'll just mark this as WIP so it doesn't get merged by accident. I consider this blocked right now until the one with `ui-toolkit` gets merged, so I can use theme helper in here.

client/elife-theme/src/elements/Button.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+`
+
+// TODO this should be updated when pubsweet changes gridUnit to 8px
+export default css`
+  line-height: calc(${th('gridUnit')} * 4);
+  min-width: calc(${th('gridUnit')} * 16);
+  padding: calc(${th('gridUnit')} * 2);
+  font-size: 14px;
```
g-sam: shouldnt this be `fontSizeBaseSmall`?

client/elife-theme/src/elements/Button.js
```diff
@@ -0,0 +1,46 @@
...
+  &[disabled] {
+    &:focus,
+    &:hover,
+    &:active {
+      border-color: ${th('colorPrimary')};
+    }
+  }
+`
+
+// TODO this should be updated when pubsweet changes gridUnit to 8px
```
g-sam: why? to what?

client/elife-theme/src/elements/Button.js
```diff
@@ -0,0 +1,46 @@
...
+export default css`
+  line-height: calc(${th('gridUnit')} * 4);
```
g-sam: shouldnt this be set from a line height variable?

client/elife-theme/src/elements/Button.js
g-sam: these should be accessed from `theme.space`

dryhten: Check the discussion me & tam had that was resolved about font size. Regarding `theme.space`, can you point me to one place where this is specified? There was too much spam about this last week and now I'm confused. My take is that it should be documented in one place (e.g. github issue). Also, I initially wanted to use something to account for those size labels but my understanding was that we don't want it?

client/elife-theme/src/elements/Button.js
dryhten: Again, check the discussion me & tam had.

client/elife-theme/src/elements/Button.js
dryhten: You mean that `theme.space` array? Same comment as above.

client/elife-theme/src/elements/Button.js
tamlyn: For the record I'm still not convinced about the 14px.

diversemix:
@g-sam Is there a list of these questions anywhere?

g-sam:
They are in invision

dryhten:
@diversemix you might have to click "Show resolved" from top right.

PR #244, opened 2018-06-22
chore: update pubsweet UI version and new theme helper location

tamlyn:
#### Background

Upgrade PubSweet packages to their latest versions. This is required in order to use the [PDK](https://gitlab.coko.foundation/pubsweet/pdk). It also brings in a couple of bug fixes like 100% width inputs.

Manually, locally.


dryhten:
I trust you updated in all places

Issue #238, opened 2018-06-22
design system decision: which component(s) should be responsible for spacing?


jsms90:
We currently have a combination of different ways to achieve spacing between components & the responsibility for layout lives at different levels of nesting:
- `margin-bottom` on individual atomic components
- `grid-styled` elements at page level

How about leaving margins off atoms altogether & leaving the responsibility for spacing between atoms to the page-level component? The page-level components are then supposed to achieve layout by using `grid-styled`.

The main benefits of this approach are:
- responsibility at single level of nesting - it's easier to reason about spacing if the responsibility lies in a single place & therefore it's easier to override defaults when necessary
- not combining 2 distinctly different methods - `grid-styled` makes a very opinionated decision to have extra HTML elements (divs) being responsible for layout. Since we're using this library, let's have the Flex & Box elements be the ones who are mainly responsible for space between elements

hdrury1:
@davidcmoulton Do you have any thoughts?

jsms90:
Based on [Julien's comment in pubsweet](https://gitlab.coko.foundation/pubsweet/design/issues/32)
> @jsms90 This is exactly the discussion that we're having with @yannis. Idea is: white space around elements should be set by the parent element, and therefore, margins should be removed from atoms.

and many conversations between YLDers while catching @g-sam & @tamlyn up on this end, we've started implementing the approach of removing `margin-bottom` from atoms

Do let us know if there are any alternative ideas on your end?

g-sam:
There is a question about whether atoms/molecules incorporating margins should also be in pubsweet/ui, or whether these belong in the individual app libraries.

g-sam:
e.g. eLife's H1 has 18px margin-bottom and 18px padding-top. Is that shareable?

jsms90:
What's EMPC? Do you have a link @diversemix?

diversemix:
https://gitlab.coko.foundation/xpub/xpub-epmc/blob/dev/app/components/Dashboard.jsx#L21

diversemix:
[...]

Issue #225, opened 2018-06-19
Update Multiselect Menu atom to conform to the styleguide


dryhten:
The Multiselect Menu molecule should look like the one in the [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985) (Select Menu with Keywords).

![image](https://user-images.githubusercontent.com/15656538/41714701-25450100-7549-11e8-84bb-ff4dd6fd4306.png)

![image](https://user-images.githubusercontent.com/15656538/41664524-d50acad6-749d-11e8-9b05-1fa120101e1e.png)

![image](https://user-images.githubusercontent.com/15656538/41664490-c4d074ae-749d-11e8-944b-f09f99d20ec4.png)

### Tasks
Sub components within react-select:
- [x] 1. Update styling of input field
  - [x] border changes based on validation status
  - [x] font family, font colour & other border variables
  - [x] spacing - padding/margin
- [x] 2. Update styling of menu options
  - [x] font colour on hover
  - [x] background colour on hover
  - [x] spacing
- [x] 3. Update styling of tags (aka keywords)
  - [x] font size
  - [x] font colour
  - [x] overwrite `react-select` hover behaviour
  - [x] spacing

---

Right now we don't have any Multiselect Menu component. We have a [Subject Area Dropdown](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js)  which should probably get more modular and be moved to `molecules`.

jsms90:
@dryhten the Subject Area Dropdown component is a specific version of the generic multiselect dropdown that has already been created in the `react-select` library. I disagree that we should create a generic multiselect menu component in `molecules` as an extra middle step.

We currently only have one implementation of a multiselect menu in our application. Until we have more multiselect dropdowns in our app, I'm not sure that a middle abstraction is useful/necessary.

With that in mind, I'm adding "Tasks" to the top comment, with the view to simply styling the SubjectAreaDropdown component as per invision.

jsms90:
1. First step - make sure I'm using the custom styles object in `react-select` correctly.

The sub-component that needs the custom styles is called `Control`, so I've added `borderColor` like this:
```
this.customReactSelectStyles = {
      multiValue: (base, state) => ({
        ...base,
        backgroundColor: this.props.theme.colorPrimary,
      }),
      control: (base, state) => ({
        ...base,
        borderColor: 'green'
      }),
[...]

PR #246, opened 2018-06-26
style(SubjectAreaDropdown): sub components of react-select to conform to new elife theme

jsms90:
#### Background
Note: to style `react-select`'s sub-components, we need to use a custom styles object - see [the docs](https://deploy-preview-2289--react-select.netlify.com/styles#style-object). This looks similar to styled components, but in fact uses "emotion".

This is the component structure for the multi-select version of `react-select`:
 - First diagram - shows when there is only a placeholder present & the ability to type a letter in the input.
- Second diagram - shows how the structure of the ValueContainer changes after a user selects their option(s)

image

### What does this PR do?
1. Tidy up:
    - fix typo in eLife theme - "col**our**Warning" -> `colorWarning`
    - remove bottom margin from overall component - relates #238 

2. Then update SubjectAreaDropdown component to conform to styleguide - see issue #225:

_Border colour_
- alter the Control component's border colour, based on whether the number of options chosen is valid (validity comes from prop on parent component)

_Spacing_
- combine `padding = gridUnit /2` on the parent + `margin = gridUnit / 2` on the child wherever possible  
Now sub-components can slot in and out & still achieve `gridUnit` spacing everywhere  
e.g.  
Parent: `ValueContainer` - responsible for half the gridUnit spacing, via padding  
Each child: `Input`, `Placeholder` & `MultiValue` - each responsible for the other half of spacing, via margin

- override `react-select` defaults & make as many margins and padding `0px` as possible, to reduce complexity
 
_Rest_
- indicate that a tag is delete-able - pointer on the X icon instead of background colour chaning (override default behaviour of `react-select`)
- remove the space between the Select Control and the Menu (`react-select` default), but leave enough room to see the Select's validation border
- remove `react-select`'s hover behaviour on the Select Control - eLife designs don't contain hover behaviour for Input fields
- update all other variables for consistency with eLife theme: tags have white text colour + blue background on, border styling to match other eLife input fields, round the edges on the bottom of the Menu with eLife's border radius

### How has this been tested?
Locally/manually. Should we be writing actual tests for CSS? (I'm mainly thinking validation behaviour, not every bit of styling)

### Known inconsistencies with design
- X icon inside the tags is smaller than designs. I could replace the default `react-select` SVG with our own one if this is important?
- Menu indicator (down chevron) does not turn upside down when the menu is open, as it is supposed to according to the designs. Again, this is default behaviour from `react-select`. Can be replaced with our own upside-down SVG chevron if we want?
- Select Control & error message overlap. This will be fixed when the `ValidatedField` component is styled. We will need to add space between the component and the validation message in `ValidatedField`, rather than relying on `margin-bottom` existing each of its sub-components.

### Issues still to be dealt with
- How to display focus on input fields in general  
Current designs suggest different greys for focused & unfocused, but ideally the colour contrast would be stronger. By default, browsers have a blue focus ring - I have left this in for now.  
See [comments in invision styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985/comments/99774850)
- How to display focus after red/orange/green validation border is applied  
Current designs don't specify. See my comment in the issue on pubsweet's design repo - https://gitlab.coko.foundation/pubsweet/design/issues/11
- Would be better to refactor, so that this Dropdown component isn't so tightly coupled with `ValidatedField` & the multiselect functionality could be re-used in other places
- Dropdown icon does not match designs

### DIFF
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -131,6 +131,14 @@ class SubjectAreaDropdown extends React.Component {
             : this.props.theme.colorText,
         padding: `${gridUnitValue * 2}px`,
       }),
+      control: (base, { isFocused }) => ({
+        ...base,
+        borderRadius: this.props.theme.borderRadius,
+        borderWidth: this.props.theme.borderWidth,
+        borderStyle: this.props.theme.borderStyle,
+        boxShadow: this.props.theme.boxShadow,
+        backgroundColor: this.props.theme.backgroundColor,
+      }),
     }
   }
 
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -109,6 +109,16 @@ class SubjectAreaDropdown extends React.Component {
         ...base,
         padding: 0,
       }),
+      menu: (base, { placement }) => ({
+        ...base,
+        borderRadius: `0 0 ${this.props.theme.borderRadius} ${
+          this.props.theme.borderRadius
+        }`,
+        borderWidth: this.props.theme.borderWidth,
+        marginBottom: 0,
+        // leave room for the bottom border of the Control component to be visible on validation/focus
+        marginTop: this.props.theme.borderWidth,
+      }),
       option: (base, { isSelected, isFocused }) => ({
         ...base,
         backgroundColor:
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -93,6 +93,16 @@ class SubjectAreaDropdown extends React.Component {
         ...base,
         padding: `${gridUnitValue / 2}px`, // combines with margin on multiValue to achieve gridUnit spacing around tags
       }),
+      placeholder: (base, state) => ({
+        ...base,
+        color: this.props.theme.colorText,
+        margin: `${gridUnitValue / 2}px`,
+        padding: `${gridUnitValue / 2}px`,
+      }),
+      input: (base, state) => ({
+        ...base,
+        padding: `${gridUnitValue / 2}px`,
+      }),
       multiValue: (base, state) => ({
         ...base,
         backgroundColor: this.props.theme.colorPrimary,
@@ -163,6 +173,7 @@ class SubjectAreaDropdown extends React.Component {
         },
         boxShadow: this.props.theme.boxShadow,
         backgroundColor: this.props.theme.backgroundColor,
+        minHeight: this.props.gridUnit,
       }),
     }
   }
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -112,8 +112,8 @@ class SubjectAreaDropdown extends React.Component {
         ...base,
         color: this.props.theme.colorTextReverse,
         fontSize: this.props.theme.fontSizeBase,
-        padding: this.props.theme.gridUnit,
-        paddingLeft: this.props.theme.gridUnit,
+        padding: '8px',
+        paddingLeft: '8px',
       }),
       multiValueRemove: (base, state) => ({
         ...base,
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -105,6 +105,18 @@ class SubjectAreaDropdown extends React.Component {
           color: this.props.theme.colorTextReverse,
         },
       }),
+      option: (base, { isSelected, isFocused }) => ({
+        ...base,
+        backgroundColor:
+          isSelected || isFocused
+            ? this.props.theme.colorPrimary
+            : 'transparent',
+        color:
+          isSelected || isFocused
+            ? this.props.theme.colorTextReverse
+            : this.props.theme.colorText,
+        padding: `${gridUnitValue * 2}px`,
+      }),
     }
   }
 
--- a/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
+++ b/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
@@ -190,7 +190,7 @@ class SubjectAreaDropdown extends React.Component {
           ),
         },
         boxShadow: isFocused
-          ? `0 0 0 2pt ${chooseBorderColorFromProps(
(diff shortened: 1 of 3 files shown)


app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -74,10 +73,106 @@ class SubjectAreaDropdown extends React.Component {
...
         backgroundColor: this.props.theme.colorPrimary,
+        color: this.props.theme.colorTextReverse,
+        margin: `${gridUnitValue / 2}px`, // combines with padding on valueContainer to achieve gridUnit spacing around tags
+      }),
+      multiValueLabel: (base, state) => ({
+        ...base,
+        color: this.props.theme.colorTextReverse,
+        fontSize: this.props.theme.fontSizeBase,
+        padding: this.props.theme.gridUnit,
+        paddingLeft: this.props.theme.gridUnit,
```
tamlyn: Isn't this line redundant given the line above?

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -74,10 +73,106 @@ class SubjectAreaDropdown extends React.Component {
...
+      theme,
+      validationStatus = 'default',
+      isFocused,
+    ) =>
+      ({
+        error: theme.colorError,
+        success: theme.colorSuccess,
+        warning: theme.colorWarning,
+        default: isFocused ? '#2684FF' : theme.colorBorder,
+      }[validationStatus])
```
tamlyn: Something very similar to this is in #248. Worth pulling out into a separate file?

jsms90:
Just made a first attempt at the focus ring. Doesn't look right yet, but nearly there.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
jsms90: I thought it was needed, to override [the original](https://github.com/JedWatson/react-select/blob/v2/src/components/MultiValue.js#L36) because `paddingLeft` is more specific than `padding`. But maybe not?

tamlyn: Oh I see because we're merging JS objects, not actually writing CSS. Carry on.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -74,10 +73,112 @@ class SubjectAreaDropdown extends React.Component {
...
+
+    const gridUnitValue = parseInt(this.props.theme.gridUnit, 10)
```
jsms90: @tamlyn Does this make sense to have? I wasn't sure if I was understanding this correctly, but using `calc` inside `customReactSelectStyles` wasn't working, so all my calculations are done like this below: `${gridUnitValue / 2}px`

jsms90:
I was just going off the styling for the separate "Text Input", "Select menu" & "Keywords" components in [the invision doc for the styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985/).

But just noticed that [the other invision for the wizard]() shows how they combine, so I've adjusted spacing to match that:

Designs:
image

In code:
image

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
```diff
@@ -74,10 +73,112 @@ class SubjectAreaDropdown extends React.Component {
...
+        ),
+        // overrides the fact that react-select changes borderColor on hover
+        '&:hover': {
+          borderColor: chooseBorderColorFromProps(
+            this.props.theme,
+            this.props.validationStatus,
+            isFocused,
+          ),
+        },
+        boxShadow: isFocused
```
jsms90: The logic for this feels really convoluted but wasn't sure how else to achieve it. Suggestions welcome.

(I added this since your review btw @tamlyn)

jsms90: Didn't see that this had been done in other components before. Should I be following [the pattern used on the checkbox](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/Checkbox.js#L74)?

hdrury1:
Aloha @jsms90, here is the cross icon .svg - may as well put it in as part of this PR?

https://drive.google.com/open?id=1y3JIUUL_kbiL_MKXnbt0lxQ8D77LABIO

jsms90:
Sure thang @hdrury1 !

@tamlyn has just showed me how we're adding svg icons to our project. Just gonna put the steps in here for future reference:
1. Paste SVG into LHS of https://svgr.now.sh/
1. Check 'Icon' on top left of the screen
1. Copy the react component that it creates on the RHS
1. Paste into the atoms folder: `atoms/icons/name-of-icon-here.svg`

jsms90:
Just tried plugging in our SVG. Doesn't look any different & if I change the fill colour to be black, it doesn't change in the browser, so I'm sure I'm missing something.

Will have a look again tomorrow at [the docs on how to replace sub components](https://deploy-preview-2289--react-select.netlify.com/components#replacing-components)

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
g-sam: i think you could import `css` from emotion and  do 
```
padding: css`calc(stuff)`
```
if you wanted

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
g-sam: Implementation looks good but we need more input on styling from Nick

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
tamlyn: Fair point. I think a basic version would be a good addition to @pubsweet/ui-toolkit but this is sufficiently different.

app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js
tamlyn: I think this is fine. Integrating with third party components is necessarily a bit messy.

jsms90:
```
Interpolating functions in css calls is deprecated and will be removed in the next major version of Emotion.
If you want to have a css call based on props, create a function that returns a css call like this
let dynamicStyle = (props) => css`color: ${props.color}`
It can be called directly with props or interpolated in a styled call like this
let SomeComponent = styled('div')`${dynamicStyle}`
```

I reckon adding 
```
(props) => css`calc(stuff)`
```
into this already beasty object is overkill

tamlyn:
Test is failing when trying to access `theme.space`.

Issue #220, opened 2018-06-19
Update TextField atom to conform to the styleguide


dryhten:
The Text input should look like the one in the [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985). For Text input we use [TextField](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/TextField.js) from pubsweet. If updating that component to look like the one in the styleguide via theme variables is not possible, then we should create our own TextField atom (e.g. in `app/components/ui/atoms`). We could also do the styling in `ValidatedField`.

PR #248, opened 2018-06-26
style(ui): apply styles to TextField component to conform with new styleguide

dryhten:
re #220

Manually, locally


dryhten:
Currently waiting for this [!294](https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/294) to be fully functional.

client/elife-theme/src/elements/TextField.js
```diff
@@ -0,0 +1,44 @@
...
+    default: theme.colorBorder,
+  }[validationStatus])
+
+/**
+ * TODO
+ *
+ * font-size & line-height might change to fontSizeBase in the future
+ *
+ * for some reason th didn't work for spacing
+ * padding: ${th('space.2') + 'px'};
```
dryhten: Suggestions welcomed

client/elife-theme/src/elements/TextField.js
```diff
@@ -0,0 +1,44 @@
...
+    error: theme.colorError,
+    success: theme.colorSuccess,
+    warning: theme.colorWarning,
+    default: theme.colorBorder,
+  }[validationStatus])
+
+/**
+ * TODO
+ *
+ * font-size & line-height might change to fontSizeBase in the future
```
tamlyn: What is the specific event that this change depends on?

client/elife-theme/src/elements/TextField.js
tamlyn: See #249

dryhten: This still doesn't answer my question. Why is this ```padding: ${props => `${props.theme.space[2]}px`}; ``` working and `padding: ${th('space.2') + 'px'};` is not? My guess is that it's some tiny stupid detail difference that I can't see right now.

client/elife-theme/src/elements/TextField.js
dryhten: Using hardcoded font-size for buttons seems like an odd decision. My assumption is that once we discuss these changes a little more with the design team, we will reach a conclusion to use `fontSizeBase` instead.

client/elife-theme/src/elements/TextField.js
```diff
@@ -0,0 +1,44 @@
...
+    border-color: ${borderColor};
+    color: ${th('colorText')};
+    font-family: ${th('fontInterface')};
+    font-size: 16px;
+    line-height: 24px;
+    padding: ${props => `${props.theme.space[2]}px`};
+    height: ${props => `${props.theme.space[5]}px`};
+  `,
+  Label: css`
+    margin-bottom: ${props => `${props.theme.space[1]}px`};
```
g-sam: these will need updating for Tam's pr

client/elife-theme/src/elements/TextField.js
tamlyn: Because `th('space.2')` returns a function, not a number, so you can't add a string to it.

dryhten: That makes sense. I guess doing something like `props => th('space.2')(props) + 'px'` would work but that defeats the purpose of `th`.

client/elife-theme/src/index.js
```diff
@@ -44,16 +44,7 @@ export default {
...
-    0,
-    gridUnit, // XXS
-    gridUnit * 2, // XS
-    gridUnit * 4, // S
-    gridUnit * 6, // M
-    gridUnit * 8, // L
-    gridUnit * 12, // XL
-    gridUnit * 20, // XXL
-  ],
+  space: [0, 1, 2, 4, 6, 8, 12, 20].map(factor => `${factor * gridUnit}px`),
```
dryhten: Isn't it annoying to have to do the math every time we need to figure out what index to use?

tamlyn: Yes it is. Put the comments back (and maybe add the actual pixel sizes too since XL doesn't mean much either).

diversemix: Maybe a comment like:
```
// space.6 = 72px 
```
... that way searching will also reveal it ;)

jsms90: Seems like it defeats the point of using a variable in the first place, no?

If the values change in the theme, you'd then need to update each comment that exists next to wherevver `space.6` is used :disappointed: 

The only reason we need the `space` array, is for `grid-styled`, right? How about we leave the space array in the theme, but let it be used for that purpose only?

Everywhere else in our code, we calculate explicitly i.e. `gridUnit * 12`. You still don't have anywhere that explicitly says "this is 72px" :woman_shrugging:  But it's easy enough to keep one variable in your head (gridUnit = 6px) & do multiplications like this in your head: 6px x 12.

Whereas it's really hard trying to remember what `space.6` refers to. After all `6` is just the 6th position in a list & that list is generated based on a fairly arbitrary scale.

Whaddaya reckon?

client/elife-theme/src/elements/TextField.js
```diff
@@ -0,0 +1,44 @@
...
+  `,
+  Root: css`
+    margin-bottom: ${props => `${props.theme.space[3]}px`};
```
jsms90: We're removing `margin-bottom` from atomic components, right? See https://gitlab.coko.foundation/pubsweet/design/issues/32 & #238 

So we also need an MR to pubsweet to [remove `margin-bottom` from the default TextField](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/TextField.js#L8)

client/elife-theme/src/elements/TextField.js
jsms90: Are we not using `grid-styled` for layout & therefore wouldn't be using `margin-bottom` but using flexbox for spacing between elements?

Stangely, [`Root` is currently already a flexbox](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/TextField.js#L5-7) but pubsweet is not using grid-styled for this :confused:

client/elife-theme/src/elements/TextField.js
```diff
@@ -0,0 +1,44 @@
+import { css } from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+/**
+ * TODO
+ * this should be removed once it's implemented in pubsweet
+ * right now they don't have colorWarning
```
jsms90: `colorWarning` is back in [their TextField](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/TextField.js#L23) - from [this MR](https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/296/diffs#dbe470512771627c07b2f1fdb1d1ade7d82834d3)

client/elife-theme/src/elements/TextField.js
jsms90: The values for font size and line height are identical to that of `fontSizeBase`, so what's stopping us from using these variables right now?

client/elife-theme/src/elements/TextField.js
hdrury1: Have we changed this to colourWarning (UK spelling) now?

hdrury1: Ah, nevermind. We're using colorWarning (misread @jsms90 edit on the other PR)!

client/elife-theme/src/elements/TextField.js
jsms90: You can see what I mean if you have a look at what I started doing for `TextArea`: https://github.com/elifesciences/elife-xpub/pull/259

client/elife-theme/src/elements/TextField.js
dryhten: Cool, thanks @jsms90. I wouldn't worry too much about this now since we're supposed to merge this one today. That MR you linked to doesn't have everything I need so I need to do another MR to pubsweet myself. Also, I think we will end up renaming that `borderColor` function.

client/elife-theme/src/elements/TextField.js
dryhten: @jsms90 there was a discussion me & nick had (and this was discussed over several issues now). I think it's the same case as it is for button, if you go on invision and click "show resolved" from top right, then go to button font size you should see everything that's been discussed.

client/elife-theme/src/elements/TextField.js
dryhten: Updated to use theme helper. @jsms90 my understanding was that we will do margins at page level in a different issue/PR. I initially added those margins there because it was looking weird without them but now for some reason (one of these PRs) there is a huge margin of 24px after each group of Input + Label.

client/elife-theme/src/elements/TextField.js
jsms90: How do you mean, it doesn't have everything you need? This exact function exists in pubsweet's `TextField`. So can we not remove the comment & the `borderColor` function altogether? :confused:

client/elife-theme/src/elements/TextField.js
jsms90: I don't see anything in the comments that applies to this? It seems like that conversation was very specifically about the button

This is what @NickDuf says:
> The font-size and line height is typical to the button
> @MihailDunaev I think it would make sense for button fonts to have their own sizing.

I know you've had a long back and forth on this though. So perhaps there's other conversation elsewhere in Slack or something?

But it seems like that was motivated by the fact that the combination of font size and line height specified for the button was not covered by the typography combinations that exist in our theme?

Whereas this component specifies exactly the combination that corresponds to `fontSizeBase`/`fontLineHeightBase`. So I must be missing something, but why are we not using these variables?

client/elife-theme/src/elements/TextField.js
dryhten: Oh yeah, this function was included, my bad, thanks.

client/elife-theme/src/elements/TextField.js
dryhten: Yeah but component specification is hardcoded as well, I would assume it's a similar case.

PR #249, opened 2018-06-26
fix: update uses of gridUnit

### LINKED ISSUE #247: Update eLife uses of gridUnit
Changing the gridUnit to 6px broke a lot of the styling in the application. As well as updating PubSweet components to use the new value, we need to update all internal uses of the variable to be multiplied by four.

tamlyn:
#### Background

Fixes #247 

- Use `space` instead of `gridUnit` when possible
- Tweak layout of file upload component to use `space` values
- Add units to `space` values in theme

Locally, visually.


app/components/global/ErrorPage.js
```diff
@@ -19,7 +19,7 @@ const Container = styled.div`
 `
 
 const Img = styled.img`
-  height: calc(${th('gridUnit')} * 4);
+  height: ${th('space.6')};
```
diversemix: This is the only instance where the value has changed ... i.e. old equates to 96px and new to 36px ... is this correct?

dryhten: Yes, old value is `96px` but new value is `72px`

diversemix: Surely 72px would need `space.12` (i.e. 6 * 12 = 72 )

g-sam: should nick be specifying 'x' instead of px?

Issue #229, opened 2018-06-19
Change/create Hide/show Panels


dryhten:
We need a new component for the Hide/show Panel from [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985).

![image](https://user-images.githubusercontent.com/15656538/41656871-54a5b540-7489-11e8-9141-41f3a0b957eb.png)

I would place this one in `atoms`.

jsms90:
If I am interpreting the invision designs correctly, the place in the designs where this component is going to be used on the 3rd page of our wizard? So this component would replace what has currently been implemented?

**Current**
![image](https://user-images.githubusercontent.com/15656538/41656820-2d13be5a-7489-11e8-8688-57a1507e1e26.png)

**New**
![image](https://user-images.githubusercontent.com/15656538/41656871-54a5b540-7489-11e8-9141-41f3a0b957eb.png)

If so, this is currently being achieved by a combination of:
- pubsweet's `Checkbox` atom
- our `TextArea` atom
- both wrapped inside our `CalloutBox` component

See comment 5 on the invision designs, which I have just posted.
> What is a "hide/show panel", without a checkbox? Is this not just a TextArea? In which case, do we need the X icon?

If I am correct in my assumptions of where the hide/show component is being used in our app, then we don't need to create any new components.

But the X icon inside the TextArea/TextField (in the invision designs), is new. Is this to be used in all 3 places within the Manuscript Metadata page? Or only for the TextArea under the first checkbox?

If it's only for the first TextArea, then we simply need to alter the existing TextArea atom, to add this functionality. If product wants to have the X icon functionality in the TextFields as well, then we will need to create a new component that wraps TextField & TextArea. @hdrury1 perhaps we should get on a hangouts call to discuss product requirements for the Manuscript Metadata page?

---

Or have I misunderstood & the hide/show component is something which is to be used elsewhere?

jsms90:
Based on @NickDuf's comment in invision, it seems that this refers to a component/components on page 4. Waiting on access to the latest invision designs for page 4.

jsms90:
So, the designs above actually represent 2 distinctly different components:
### 1. 1st hide/show panel
![image](https://user-images.githubusercontent.com/15656538/41721383-9e251060-755d-11e8-8e29-8dfc0edb04b0.png)

This empty box with an X icon is actually part of a more complex component, represented on the new page 4 designs:

[...]

PR #257, opened 2018-06-27
feat: style Headings and CalloutBox

### LINKED ISSUE #226: Update Heading atom to conform to the styleguide
The Heading atom should look like the one in the [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985).

Right now we use [Heading](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/Heading.js) from pubsweet. If updating that component to look like the one in the styleguide via theme variables is not possible, then we should create our own Button atom (e.g. in `app/components/ui/atoms`).

g-sam:
**Includes pubsweet ui version bump**

Closes #226 and #229 

Apologies for the untidy merge commit situation. I did something strange and couldn't escape. (Well I found some 5-step instructions on how to do it...)

### DIFF
--- a/app/components/dashboard/Dashboard.js
+++ b/app/components/dashboard/Dashboard.js
@@ -1,15 +1,16 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
+import { FormH2, FormH3 } from '../ui/atoms/FormHeadings'
 import ButtonLink from '../ui/atoms/ButtonLink'
 
 const Dashboard = ({ manuscripts }) => (
   <React.Fragment>
-    <h1>Dashboard Dummy Page</h1>
+    <FormH2>Dashboard Dummy Page</FormH2>
     <ButtonLink data-test-id="submit" primary to="/submit">
       Submit a manuscript
     </ButtonLink>
 
-    <h3>All manuscripts</h3>
+    <FormH3>All manuscripts</FormH3>
     <ul>
       {manuscripts.map(manuscript => (
         <li key={manuscript.id}>
--- a/app/components/submission/ReviewerSuggestions/FormSections.js
+++ b/app/components/submission/ReviewerSuggestions/FormSections.js
@@ -5,6 +5,14 @@ import { Checkbox } from '@pubsweet/ui'
 import ValidatedField from '../../ui/atoms/ValidatedField'
 import CalloutBox from '../../ui/atoms/CalloutBox'
 import Textarea from '../../ui/atoms/Textarea'
+import { FormH3 } from '../../ui/atoms/FormHeadings'
+
+const createCloseHandler = (index, values, fieldName, setFieldValue) => () =>
+  setFieldValue(
+    fieldName,
+    values[fieldName].filter((_, i) => i !== index),
+    false,
+  )
 
 export const SuggestedSeniorEditorRow = ({ rowIndex }) => (
   <Flex>
@@ -23,8 +31,16 @@ export const SuggestedSeniorEditorRow = ({ rowIndex }) => (
   </Flex>
 )
 
-export const ExcludedSeniorEditor = ({ index }) => (
-  <CalloutBox>
+export const ExcludedSeniorEditor = ({ index, values, setFieldValue }) => (
+  <CalloutBox
+    onClose={createCloseHandler(
+      index,
+      values,
+      'opposedSeniorEditors',
+      setFieldValue,
+    )}
+  >
+    <FormH3>Exclude a Senior Editor</FormH3>
     <ValidatedField
       label="Excluded senior editor"
       name={`opposedSeniorEditors.${index}.name`}
@@ -54,8 +70,16 @@ export const SuggestedReviewingEditorRow = ({ rowIndex }) => (
   </Flex>
 )
 
-export const ExcludedReviewingEditor = ({ index }) => (
-  <CalloutBox>
+export const ExcludedReviewingEditor = ({ index, values, setFieldValue }) => (
+  <CalloutBox
+    onClose={createCloseHandler(
+      index,
+      values,
+      'opposedReviewingEditors',
+      setFieldValue,
+    )}
+  >
+    <FormH3>Exclude a Reviewing Editor</FormH3>
     <ValidatedField
       label="Excluded reviewing editor"
       name={`opposedReviewingEditors.${index}.name`}
@@ -63,7 +87,7 @@ export const ExcludedReviewingEditor = ({ index }) => (
     <ValidatedField
       component={Textarea}
       label="Reason for exclusion"
-      name={`opposedSeniorEditors.${index}.reason`}
+      name={`opposedReviewingEditors.${index}.reason`}
     />
   </CalloutBox>
 )
@@ -85,8 +109,17 @@ export const SuggestedReviewer = ({ index }) => (
     </Box>
   </Flex>
 )
-export const ExcludedReviewer = ({ index }) => (
-  <CalloutBox>
+
+export const ExcludedReviewer = ({ index, values, setFieldValue }) => (
+  <CalloutBox
+    onClose={createCloseHandler(
+      index,
+      values,
+      'opposedReviewers',
+      setFieldValue,
+    )}
+  >
+    <FormH3>Exclude a Reviewer</FormH3>
     <Flex>
       <Box width={1 / 2}>
         <ValidatedField
--- a/app/components/ui/atoms/CalloutBox.js
+++ b/app/components/ui/atoms/CalloutBox.js
@@ -2,16 +2,36 @@ import React from 'react'
 import styled from 'styled-components'
 import { Box } from 'grid-styled'
 import { th } from '@pubsweet/ui-toolkit'
+import Icon from './Icon'
 
-const CalloutBox = styled(({ enabled, ...rest }) => <Box {...rest} />).attrs({
-  mx: -3,
-  px: 3,
-  mb: 3,
+const Frame = styled(props => <Box {...props} />).attrs({
+  mb: 2,
 })`
   border: ${th('borderWidth')} ${th('borderStyle')} ${th('borderColor')};
   border-radius: ${th('borderRadius')};
-  border-color: ${({ enabled = true }) =>
-    enabled ? th('colorBorder') : 'transparent'};
+  border-color: ${th('colorBorder')};
+  position: relative;
 `
 
+const CloseButton = styled(Icon)`
+  position: absolute;
+  top: 0px;
+  right: 0px;
+  padding: ${th('space.2')};
+  svg {
+    stroke: ${th('colorFurniture')};
+    fill: ${th('colorFurniture')};
+  }
+`
+
+const CalloutBox = ({ onClose, children, ...props }) => (
+  <Frame {...props}>
+    {onClose && (
+      <CloseButton onClick={onClose} size={5}>
+        Cross
+      </CloseButton>
+    )}
+    <Box p={3}>{children}</Box>
+  </Frame>
+)
 export default CalloutBox
--- a/app/components/ui/atoms/FormHeadings.js
+++ b/app/components/ui/atoms/FormHeadings.js
@@ -1,0 +1,10 @@
+import { H2, H3 } from '@pubsweet/ui'
+
+export const FormH2 = H2.extend`
+  margin-bottom: 36px;
+`
+
+export const FormH3 = H3.extend`
+  padding-top: 0px;
+  margin-bottom: 24px;
+`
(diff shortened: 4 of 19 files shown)


app/components/ui/atoms/Icon.js
```diff
@@ -7,11 +7,11 @@ import * as icons from './icons'
 
 const Container = styled.span`
   display: inline-flex;
-  padding: calc(${th('gridUnit')} / 2);
+  padding: ${th('space.2')}px;
   svg {
     stroke: ${props => props.color || props.theme.colorText};
-    width: calc(${props => props.size} * ${th('gridUnit')});
-    height: calc(${props => props.size} * ${th('gridUnit')});
+    width: ${props => props.theme.space[props.size]}px;
```
g-sam: oops, needs updating

client/elife-theme/src/elements/Button.js
```diff
@@ -0,0 +1,46 @@
+import { css } from 'styled-components'
```
g-sam: this shouldnt be here!

app/components/ui/atoms/CalloutBox.js
```diff
@@ -2,16 +2,36 @@ import React from 'react'
...
+`
+
+const CalloutBox = ({ closable, onClose, children, ...props }) => (
+  <Frame {...props}>
+    {closable && (
+      <CloseButton onClick={onClose} size={3}>
+        Cross
+      </CloseButton>
+    )}
+    <Box p={3}>{children}</Box>
```
jsms90: This perhaps deserves a wider conversation than this PR, but what do you guys think about using render props instead of higher order components?

One reason for this is readability. When an HOC is used (e.g. [in `FormSections.js`](https://github.com/elifesciences/elife-xpub/pull/257/files#diff-c54657ef098d458843cc24ce86e579f8R116)), it doesn't mean much to the person reading the file, unless they go and look into how the `CalloutBox` itself actually works.

app/components/ui/atoms/CalloutBox.js
```diff
@@ -2,16 +2,36 @@ import React from 'react'
...
+  top: 0px;
+  right: 0px;
+  padding: ${th('space.2')};
+  svg {
+    stroke: ${th('colorFurniture')};
+    fill: ${th('colorFurniture')};
+  }
+`
+
+const CalloutBox = ({ closable, onClose, children, ...props }) => (
```
tamlyn: Would it be better to have `onClose` implicitly make the box closable? Or 
 are there situations where you could use one prop and not the other?

app/components/ui/atoms/FormHeadings.js
```diff
@@ -0,0 +1,19 @@
...
+    H3: css`
+      /* this reset from the theme doesn't work? works applied here */
+      margin-top: 0px;
+      margin-bottom: 24px;
+    `,
+  },
+}
+
+export const FormH2 = props => <H2 theme={formSpacing} {...props} />
+export const FormH3 = props => <H3 theme={formSpacing} {...props} />
```
tamlyn: I would expect styles like this to be applied with styled components `extend` or `styled()` rather than a local css override.

app/components/ui/atoms/icons/Cross.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+const Cross = () => (
+  <svg height="1em" viewBox="0 0 24 24" width="1em">
```
tamlyn: Any reason you're not passing through `props`?

app/components/submission/ReviewerSuggestions/FormSections.js
```diff
@@ -5,6 +5,14 @@ import { Checkbox } from '@pubsweet/ui'
 import ValidatedField from '../../ui/atoms/ValidatedField'
 import CalloutBox from '../../ui/atoms/CalloutBox'
 import Textarea from '../../ui/atoms/Textarea'
+import { FormH3 } from '../../ui/atoms/FormHeadings'
+
+const createCloseHandler = (index, values, fieldName, setFieldValue) => () =>
```
jsms90: We wouldn't have to dynamically generate a closeHandler, if we were to favour composition over configuration.

Having to "configure" `CalloutBox` to have an X icon/not have an X icon is what leads to this kind of extra abstraction to create closing functionality. IMO this is a little unnecessarily convoluted & makes this file slowly more and more difficult to reason about.

app/components/dashboard/Dashboard.js
```diff
@@ -1,15 +1,16 @@
...
 const Dashboard = ({ manuscripts }) => (
   <React.Fragment>
-    <h1>Dashboard Dummy Page</h1>
+    <FormH2>Dashboard Dummy Page</FormH2>
     <ButtonLink data-test-id="submit" primary to="/submit">
       Submit a manuscript
     </ButtonLink>
 
-    <h3>All manuscripts</h3>
+    <FormH3>All manuscripts</FormH3>
```
tamlyn: Is it weird to use the form headings on something that isn't a form?

app/components/ui/atoms/CalloutBox.js
```diff
@@ -2,16 +2,36 @@ import React from 'react'
...
+  <Frame {...props}>
+    {closable && (
+      <CloseButton onClick={onClose} size={3}>
```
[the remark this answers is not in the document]
g-sam: An X in the top right should always be for closing shouldn't it? I feel that *how* the closing is achieved is a concern of the component above this one though.

app/components/submission/ReviewerSuggestions/FormSections.js
g-sam: It comes down to whether you think the panel component should know about the context in which it's being rendered (as part of an array in a form): if not, then you need something like this

client/elife-theme/src/elements/Heading.js
```diff
@@ -0,0 +1,32 @@
...
+
+export default css`
+  /* clear browser-added margin and font-size */
+  margin-top: 0px;
+  font-size: 1em;
+  color: ${th('colorText')};
+`
+export const H1 = css`
+  padding-top: 18px;
+  margin-bottom: 18px;
```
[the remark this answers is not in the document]
g-sam: My thinking here is we should create our base components with the most commonly used margins and override where different (as in FormH2, etc.).

If we do NOT do this, we still have to provide our own css in FormH2, so nothing lost?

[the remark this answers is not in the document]
g-sam: But I am also in two minds about where spacing for form headers should be applied

app/components/ui/atoms/CalloutBox.js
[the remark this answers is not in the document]
jsms90: But you can't _mandate_ that the component above attaches a handler which applies closing functionality.

I could use `CalloutBox` elsewhere and pass in my own `handleClick` to the `onClose` prop, which does something totally different from "closing" e.g. turns the colour of some text red:

```
class myParentComponent extends React.Component {
  constructor(props) {
    super(props)
    this.state = {
      textColour: 'black'
    }
  }
  
  handleClick = () => {
    this.setState({textColor: 'red})
  }

  render() {
    <div>
      <h2 color={this.state.textColor}>Text here</h2>
    <CalloutBox
        closable
        onClose={handleClick}
    >
  }
}
```

Just because you call the prop `onClose`, doesn't mean that's what clicking an X actually does. A more accurate name for this prop would be `onXIconClick`.

This is the inherent problem with separating which component renders the X and which component handles the clicking of the X.

Do you see what I mean?

app/components/submission/ReviewerSuggestions/FormSections.js
g-sam: As a side note Tam poointed out we could simplify this in some respects by using Formik's `FieldArray`. I felt it was actually simpler to keep it this way though, as introducing `FormArray` creates complexities in other respects.

app/components/dashboard/Dashboard.js
g-sam: Yes. This page is 'design: unknown', however

app/components/ui/atoms/CalloutBox.js
g-sam: Possibly. I thought it might be a bit magical. Thoughts one way or the other?

app/components/ui/atoms/CalloutBox.js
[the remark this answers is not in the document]
tamlyn: You can't stop people from misusing your component but making your app do anything other than close as a result of someone clicking an X icon in the top right of a box is silly. 

I think perhaps the issue is that "close" isn't the event that has taken place. So in that sense `onXIconClick` is more accurate but accuracy isn't everything and to me at least `onClose` is more helpful.

I would settle for `onCloseClick` if pushed.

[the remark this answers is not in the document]
g-sam: I see what you mean but surely by the same argument Formik's `onSubmit` should be called `onSubmitButtonClick`?

app/components/ui/atoms/CalloutBox.js
tamlyn: I'm not sure I follow what you mean by "higher order component".

client/elife-theme/src/elements/Heading.js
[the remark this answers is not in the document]
jsms90: Yeah I hear you. Either way, so long as the default headings in pubsweet have no `margin-bottom` applied, that's the main thing :+1: Then how we add margins in eLife is up for grabs & you may be right with this.

Right now our invision doc include 2 different measurements underneath an h1:
- 18px if there's a p tag below (not actually used anywhere in our app yet)
- 36px for our wizard (used in every place that we have an h1 in our app)

Based on the fact that there is _always_ 36px underneath an h1 in all the designs we have for our app, I totally get this choice & you may be right.

I guess all I'm highlighting here is that we are fairly likely to need other measurements underneath an h1 in future. The, as yet unused, 18px being one example. If/when we have other h1s in our app in future, we may well end up removing all the margins for exactly the same reason that we are right now doing this in pubsweet.

For example, if we do use the 18px designs anywhere in future, we will be choosing between:
- Creating 2 components: FormH1 (36px margin) & ParagraphH1 (18px margin)
- Letting the molecule/page decide margins, probably via grid-styled

If, in future, we have another 3rd type of h1 then we will need a 3rd h1 component. And so on and so on. The fact that we would be creating new atoms to wrap an h1, where each one differs _only_ in how much margin they have at the bottom feels a little like a design system code smell to me. You can even feel it in the naming - `FormH1`, `ParagraphH1`, `ButtonH1`, etc. The names indicate that they are combined with other elements...but that's not actually the case. It's not a molecule that's made up of a `p` and an `h1`. It's misleading. For those names to make sense, you would expect to see a molecule that contains and `h1` and a `p`, or an h1 and a form (aka the page).

app/components/ui/atoms/CalloutBox.js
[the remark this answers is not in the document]
jsms90: Ahh yeah, I see what you mean about misusing a component :thinking:

Perhaps my issue is more that this abstraction seems kinda arbitrary. We don't need a boxed component with an X for anything other than the situation where it's combined with a link that generates the box(es) in the first place.

So I don't see any need to remove the existence of the X icon from its intended purpose & it kinda makes the `FormSections.js` code harder to reason about. All for the sake of re-using a border in 2 places. Just feels like the wrong abstraction to me :woman_shrugging: 

It would make more sense to me, if we were building molecules that combined components for encapsulated behaviour i.e. if we built 2 component based on different ways to hide/show:
- checkbox generated boxes - checking generates & closes
- link generated/X closed boxes

The fact that both of these create a border around some fields doesn't feel like an important thing to group together?

And if we grouped the components this way instead, it would make for a page whose render function was more explicit and therefore easier to reason about.

app/components/ui/atoms/CalloutBox.js
jsms90: Oh I'm talking nonsense. This isn't a higher order component at all. It doesn't take in a component and return a component. :flushed:

client/elife-theme/src/elements/Heading.js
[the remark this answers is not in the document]
g-sam: Let's keep as is for now and see if it makes sense to change later on

app/components/ui/atoms/CalloutBox.js
g-sam: We did make a decision a while ago to favour render props, but, yes, not applicable here.

PR #258, opened 2018-06-27
style: update Checkbox component

### LINKED ISSUE #221: Update Checkbox atom to conform to the styleguide
The Checkbox atom should look like the one in the [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985).

Right now we use [Checkbox](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/Checkbox.js) from pubsweet. If updating that component to look like the one in the styleguide via theme variables is not possible, then we should create our own Checkbox atom (e.g. in `app/components/ui/atoms`).

tamlyn:
#### Background
 
- Fixes #221 
- Add Checkmark icon component but use inline data uri until icons are supported in themes
- Reorganise theme override folder structure to reflect object structure
- Ready for review ~~but waiting on [pubsweet!301](https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/301)~~

Visually ~~when linked to the latest `@pubsweet/ui`~~.

### DIFF
--- a/client/elife-theme/src/index.js
+++ b/client/elife-theme/src/index.js
@@ -10,8 +10,8 @@ export default {
   colorSecondary: '#f7f7f7',
   colorFurniture: '#E0E0E0',
   colorBorder: '#E0E0E0',
-  colorBackgroundHue:
-    '#f7f7f7' /* marginally darker shade of the app background so that it can be used to divide the interface when needed */,
+  /* marginally darker shade of the app background so that it can be used to divide the interface when needed */
+  colorBackgroundHue: '#f7f7f7',
   colorSuccess: '#629f43',
   colorError: '#cf0c4e',
   colourWarning: '#e65100',
--- a/app/components/ui/atoms/icons/Checkmark.js
+++ b/app/components/ui/atoms/icons/Checkmark.js
@@ -1,0 +1,9 @@
+import React from 'react'
+
+const Checkmark = props => (
+  <svg height="1em" viewBox="0 0 24 24" width="1em" {...props}>
+    <path d="M9 16.2L4.8 12l-1.4 1.4L9 19 21 7l-1.4-1.4L9 16.2z" />
+  </svg>
+)
+
+export default Checkmark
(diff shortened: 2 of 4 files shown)


diversemix:
Having just merged the other 'fix' to the e2e tests ... is there any benefit to rebasing in the hope that this might start working? ... If it doesn't break `develop` I'm happy to merge anyhow?

tamlyn:
This is already up to date with `develop`. Hmmm. Ah well, let's wait for the gridUnit update to land.

client/elife-theme/src/index.js
```diff
@@ -59,7 +59,5 @@ export default {
   transitionTimingFunction: 'ease',
   transitionDelay: '0',
 
-  cssOverrides: {
-    Button,
-  },
+  cssOverrides,
```
dryhten: This is in conflict to what I did in https://github.com/elifesciences/elife-xpub/pull/248

dryhten: I think this would move faster if we agree on something and then don't change things that are not related to the issue in commits?

client/elife-theme/src/overrides/index.js
```diff
@@ -0,0 +1,3 @@
+import * as ui from './ui'
```
dryhten: Why did you call this overrides? I think it would be better to keep the name `elements` since everyone else (coko, epmc etc) are using elements?

dryhten: https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/coko-theme/src
https://gitlab.coko.foundation/xpub/xpub-epmc/tree/dev/app/epmc-theme

tamlyn: Because then the folder structure matches the object structure.

client/elife-theme/src/index.js
tamlyn: OK. Can we agree on this?

Issue #219, opened 2018-06-19
Update Text area atom to conform to the styleguide


dryhten:
The Text area atom should look like the one in the [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985). We have our own [Text area](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/ui/atoms/Textarea.js).

jsms90:
@tamlyn What's the reasoning for [this line](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/ui/atoms/Textarea.js#L51)?
> {label && <Label htmlFor={this.inputId}>{label}</Label>}

Would there be any objections to me taking this out, in favour of making it a required PropType? I'm assuming that we would never want to create a `TextArea` (or `TextField`) without a label, for accessibility reasons, even if we then chose to hide it visually?

tamlyn:
This component is 95% copy pasted from [`TextField`](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/TextField.js#L61). I think it should be generalised into a single component but the opinion of the crowd was to delay that until later.

jsms90:
So an issue to raise in pubsweet, to get it changed in both places?

jsms90:
Ok so I started work on this, but the decisions that I was making were diverging from [how @dryhten has approached it for `TextField`](https://github.com/elifesciences/elife-xpub/pull/248) & these components need to stay in line with each other.

Seems like it might be worth stopping work on this, waiting for the `TextFiled` changes to be finalised & the PR to be merged, so we can just copy/paste whatever we end up with on that component.

PR #259, opened 2018-06-27
WIP style(TextArea): used grid-styled Flex & Box for layout

jsms90:
Just for reference - should prob be done after `TextField` is finalised

#219


app/components/ui/atoms/Textarea.js
```diff
@@ -47,10 +40,19 @@ class Textarea extends React.Component {
   render() {
     const { label, value = '', readonly, ...props } = this.props
     return (
-      <Root>
-        {label && <Label htmlFor={this.inputId}>{label}</Label>}
-        <Input id={this.inputId} readOnly={readonly} value={value} {...props} />
-      </Root>
+      <Flex flexDirection="column">
+        <Box mb={th('gridUnit')}>
```
tamlyn: Does this work? Shouldn't it be `mb={1}`?

app/components/ui/atoms/Textarea.js
```diff
@@ -45,14 +41,28 @@ class Textarea extends React.Component {
...
-      </Root>
+      <Flex flexDirection="column">
+        <Box mb={1}>
+          <Label htmlFor={this.inputId}>{label}</Label>
+        </Box>
+        <Box>
+          <Input
+            id={this.inputId}
+            readOnly={readonly}
+            type={type}
```
tamlyn: What does `type` do on a textarea?

jsms90: Nothing at all. Thanks @tamlyn 

Made the change :+1:

app/components/ui/atoms/Textarea.js
```diff
@@ -1,19 +1,14 @@
...
-  margin-bottom: ${th('space.3')};
-`
+import { Flex, Box } from 'grid-styled'
 
 const Label = styled.label`
   font-size: ${th('fontSizeBaseSmall')};
-  display: block;
 `
 
+// TODO - remove when this function is added to pubsweet
```
dryhten: This was added to pubsweet!

app/components/ui/atoms/Textarea.js
```diff
@@ -45,14 +41,27 @@ class Textarea extends React.Component {
...
+            value={value}
+            {...props}
+          />
+        </Box>
+      </Flex>
     )
   }
 }
 
+Textarea.propTypes = {
```
dryhten: More props here?

app/components/ui/atoms/Textarea.md
```diff
@@ -0,0 +1,10 @@
+initialState = { value: '' }
+
+```js
+<Textarea
+  label="Label"
+  value={state.value}
```
dryhten: Is this working? What's `state` here? Also, what's `initialState`?

app/components/ui/atoms/Textarea.js
jsms90: Pubsweet doesn't have a `Textarea` component This function only exists within their `TextField`.

As @tamlyn has pointed out in the `SubjectAreaDropdown` PR, it would be nice to have this function separated from the `TextField` component & therefore usable as a utility function across multiple components. But I would suggest that this refactor in pubsweet's repo can wait. We can abstract that function out of `TextField` at the same time as we add a `TextArea` to their repo?

jsms90:
No worries @dryhten, it's not too much work. These were great points & helps make this component cleaner, so thanks for your review :+1: 

All done, lemme know what you think :grin:

app/components/ui/atoms/Textarea.js
```diff
@@ -45,14 +41,24 @@ class Textarea extends React.Component {
     this.inputId = `textarea-${Math.round(Math.random() * 1e12).toString(36)}`
   }
   render() {
-    const { label, value = '', readonly, ...props } = this.props
+    const { label, ...props } = this.props
     return (
-      <Root>
-        {label && <Label htmlFor={this.inputId}>{label}</Label>}
-        <Input id={this.inputId} readOnly={readonly} value={value} {...props} />
```
dryhten: Javascript seems to be case sensitive with variables names, are you sure this works?

app/components/ui/atoms/Textarea.js
```diff
@@ -45,14 +41,24 @@ class Textarea extends React.Component {
     this.inputId = `textarea-${Math.round(Math.random() * 1e12).toString(36)}`
   }
   render() {
-    const { label, value = '', readonly, ...props } = this.props
```
dryhten: Why did you remove default value for `value` here? I assume whoever added it, did it for a reason? Maybe somewhere this gets called without specifying value?

app/components/ui/atoms/Textarea.md
```diff
@@ -0,0 +1,15 @@
+Required props:
+
+* `label`: for accessibility purposes
+* `value` & `onChange`: to make this a controlled React component
+
+Note: other CSS properties for an input element may be passed in (e.g. placeholder) but are not required.
+
+```js
+<Textarea
```
dryhten: It seems you can use `state` and `initialState` in react styleguidist markdown files[1]. Now that you removed the default value for `value` and `initialState` from this file, will this work?

https://react-styleguidist.js.org/docs/documenting.html#writing-code-examples

app/components/ui/atoms/Textarea.js
jsms90: I've added `value` as a required propType, to encourage people using this component by passing in `value` & `onChange` in such a way that this is a controlled React component.

To my mind, it wouldn't make any sense for someone to use `TextArea`, without passing in value dynamically. If it is a required value anyway, there's no need to set `value=''` by default.

Pubsweet's docs specify `initialState: { value = '' } in their styleguidist docs, in an effort to encourage people to overwrite it. See my own description in our markdown file for this component. I hope that this is a good explanation, that will encourage proper usage.

app/components/ui/atoms/Textarea.md
jsms90: See the lines below this one
`onChange` sets the state
`value` passes in the state

It works. Have you tried it?

app/components/ui/atoms/Textarea.js
jsms90: Done. Thanks :smile:

app/components/ui/atoms/Textarea.js
dryhten: I just checked and TextArea doesn't seem to be used without an initial value, so it's fine for now.

Regarding the other thing you mentioned, to me it seems quite the opposite. A TextArea that forces you to pass an empty value every time you need an empty one seems like a weird design decision (I imagine most of the time you'd use an empty text area, and only in few other cases you'd want something already in there).

app/components/ui/atoms/Textarea.md
dryhten: No, that's why I asked you :) I think asking people to checkout to a new branch to test if a PR works is too much work. Everyone should test their own changes to make sure they work before doing a PR (even though that doesn't happen all the time in this repo for some reason!). I was just making sure you tested. If it works, it's all good :)

Issue #228, opened 2018-06-19
Create a Notification Ribbon component (functionality to be added)


mihaildu:
**User Story:**

_As a submitter I want a notification ribbon in order to know whether or not the corresponding author has accepted their responsibility._

**Product requirements/Acceptance criteria:**

Functional:

- When the submitter edits the email in the field on page 1 of the submission wizard and it no longer matches the email associated with the submitter's ORCiD account, a ribbon is displayed at the top of the page after the submitter hits 'Next'.
- This ribbon remains at the top of every page of the wizard until the corresponding author takes an action to validate themselves as such.
- The submitter can use the 'resend' link within the ribbon to send another validation email to the nominated corresponding author.
- If the submitter resends the validation email, the ribbon updates to indicate this.
- The presence of the ribbon should not stop the aubmitter from completing the wizard.

**Relevant links to design(s):**

[Invision Prototype](https://invis.io/CDLI0R5KBF6)
[Styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/)

**Tasks**
- [x] Create notification component - suggested that this be placed in `atoms`

jsms90:
@hdrury1 Could you change this to a feature issue, detailing the user story & product requirements?

After all, this is not just a styling issue, but an issue for creating a whole new component, with associated functionality from issue #53

hdrury1:
Editing the title and description to reflect a feature.

hdrury1:
@dryhten I've removed the email bridge task, as this is beyond the scope of this feature. For now let's stick to the functionality of having a corresponding author verify themselves as such. This is going to require an email token being generated and embedded in the verification email.

hdrury1:
Maybe we split this issue to account for the actual component that has been made already and the functionality that is still to be implemented?

dryhten:
@hdrury1 I don't mind either way. When I created this issue I had in mind only the last task. If you want better modularity I suggest we keep "doing the notification ribbon component" here and move the functional part to a different issue, so we can close this one before I'm off next week.

jsms90:
Yeah sorry guys, I probably confused things here.

My [suggestion above](https://github.com/elifesciences/elife-xpub/issues/228#issuecomment-398725956) was that a new issue get created - a feature issue (user story + requirements + list of sub tasks), to try and help @hdrury1 have a greater overview of what has/hasn't been done yet & how far away a feature was from being finished.

[...]

PR #260, opened 2018-06-27
feat(ui): add notification ribbon atom to ui

dryhten:
re #228

#### Background

This is only the Notification Ribbon atom from the referenced issue.

Locally on styleguide.

### DIFF
--- a/app/components/ui/atoms/NotificationRibbon.js
+++ b/app/components/ui/atoms/NotificationRibbon.js
@@ -1,0 +1,24 @@
+import styled from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+const backgroundColor = ({ type = 'primary', theme }) =>
+  ({
+    error: theme.colorError,
+    success: theme.colorSuccess,
+    primary: theme.colorPrimary,
+  }[type])
+
+export default styled.div`
+  background: ${backgroundColor};
+  color: ${th('colorTextReverse')};
+  padding: ${th('space.2')};
+  font-family: ${th('fontHeading')};
+  font-size: 14px;
+  line-height: 24px;
+  text-align: center;
+  a {
+    text-decoration: underline;
+    color: inherit;
+    cursor: pointer;
+  }
+`
--- a/app/components/ui/atoms/NotificationRibbon.js
+++ b/app/components/ui/atoms/NotificationRibbon.js
@@ -16,9 +16,4 @@ export default styled.div`
   font-size: 14px;
   line-height: 24px;
   text-align: center;
-  a {
-    text-decoration: underline;
-    color: inherit;
-    cursor: pointer;
-  }
 `
--- a/app/components/ui/atoms/NotificationRibbon.md
+++ b/app/components/ui/atoms/NotificationRibbon.md
@@ -1,0 +1,33 @@
+Notification ribbon that can be displayed at the top. It has 3 variations:
+
+Generic/primary ribbon
+
+```js
+<NotificationRibbon type="primary">
+  This is a generic notification ribbon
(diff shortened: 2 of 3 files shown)


app/components/ui/atoms/NotificationRibbon.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default styled.div`
+  background: ${backgroundColor};
+  color: ${th('colorTextReverse')};
+  padding: ${th('space.2')};
+  font-family: ${th('fontHeading')};
+  font-size: 14px;
+  line-height: 24px;
+  text-align: center;
+  a {
```
dryhten: Couldn't find a way to style specific styled components inside other styled components. Suggestions welcomed.

app/components/ui/atoms/NotificationRibbon.md
```diff
@@ -0,0 +1,33 @@
...
+```js
+<NotificationRibbon type="error">
+  This is an error notification ribbon
+</NotificationRibbon>
+```
+
+Anchors also have a special style inside a ribbon
+
+```js
+;<NotificationRibbon type="primary">
```
diversemix: Is the semicolon important?

dryhten: No, I forgot it in there from when I was trying to style `Action` elements inside `NotificationRibbon`.

dryhten: I just removed it.

app/components/ui/atoms/NotificationRibbon.js
g-sam: I think these styles should be applied to the component passed in by *that* component.

app/components/ui/atoms/NotificationRibbon.js
```diff
@@ -0,0 +1,24 @@
+import styled from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+const backgroundColor = ({ type = 'primary', theme }) =>
+  ({
+    error: theme.colorError,
+    success: theme.colorSuccess,
+    primary: theme.colorPrimary,
+  }[type])
```
tamlyn: We now have this helper function in three places. Surely it's time to factor it out into a common place?

app/components/ui/atoms/NotificationRibbon.js
tamlyn: On one level this style rule makes sense: we specify the text colour for this element and we want all other child elements to have the same text colour. But using a naked `a` tag means we have to replicate any styling and behaviour attached to `Link`. What's more resend would probably be implemented using a `button` (via `Action`).

As discussed I think the best option is to create a separate `RibbonAction` component that applies the necessary text colour by extending `Action`.

dryhten: It seems that it's possible to apply style to specific styled components inside other styled components[1]. The problem is that the component we want to restyle is not actually a styled component. I will continue this discussion on coko's mattermost. For now I guess it's better to do the Action restyling in a different component so it's easier to change later on.

[1] - https://www.styled-components.com/docs/advanced#referring-to-other-components

app/components/ui/atoms/NotificationRibbon.js
dryhten: While this takes type as input (as opposed to validationStatus), I agree, there should be a function like `getColor` somewhere in pubsweet that accounts for every possibility (error, success, warning, primary etc).

dryhten:
I should mention that because we are using Actions now, there is this fancy effect when you click on the link in the ribbon -  a white border will show up for a brief period of time.

PR #263, opened 2018-06-28
style: update Menu component

### LINKED ISSUE #223: Update Select Menu molecule to conform to the styleguide
The Select Menu molecule should look like the one in the [styleguide](https://projects.invisionapp.com/share/NZKXOQUV4P3#/screens/302926985).

Right now we don't have any Select Menu component. We could start from [Subject Area Dropdown](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/submission/ManuscriptMetadata/SubjectAreaDropdown.js) from our own repository or the [Menu](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/molecules/Menu.js) from pubsweet.

tamlyn:
#### Background

- Fixes #223 
- Ready for review ~~but do not merge until [pubsweet!304](https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/304) is merged and published~~

Visually ~~when linked to PubSweet master.~~

### DIFF
--- a/client/elife-theme/src/elements/ui/Menu.js
+++ b/client/elife-theme/src/elements/ui/Menu.js
@@ -1,0 +1,51 @@
+import { css } from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+export default {
+  Arrow: css`
+    color: ${th('colorTextSecondary')};
+    transition: none;
+    font-size: ${th('gridUnit')};
+  `,
+  ArrowContainer: css`
+    border-left: none;
+  `,
+  Label: css`
+    margin-bottom: ${th('space.1')};
+  `,
+  Opener: css`
+    &:hover {
+      border-color: ${th('colorBorder')};
+    }
+    ${props =>
+      props.open &&
+      `border-bottom-left-radius: 0; border-bottom-right-radius: 0;`};
+  `,
+  Option: css`
+    height: ${th('space.5')};
+    line-height: ${th('space.5')};
+    padding: 0 ${th('space.2')};
+    font-weight: inherit;
+
+    &:hover {
+      color: ${th('colorTextReverse')};
+      background-color: ${th('colorPrimary')};
+      border: none;
+    }
+  `,
+  Options: css`
+    top: -1px;
+    border-radius: 0 0 ${th('borderRadius')} ${th('borderRadius')};
+  `,
+  Value: css`
+    &:hover {
+      color: ${th('colorText')};
+    }
+  `,
+  Placeholder: css`
+    font-style: normal;
+    &:hover {
+      color: ${th('colorTextPlaceholder')};
+    }
+  `,
+}
--- a/client/elife-theme/src/index.js
+++ b/client/elife-theme/src/index.js
@@ -6,6 +6,7 @@ const gridUnit = 6
 
 export default {
   /* Colors */
+  colorBackground: '#fff',
   colorPrimary: '#0288d1',
   colorSecondary: '#f7f7f7',
   colorFurniture: '#E0E0E0',
--- a/app/components/submission/ManuscriptMetadata/ManuscriptMetadata.js
+++ b/app/components/submission/ManuscriptMetadata/ManuscriptMetadata.js
@@ -1,5 +1,4 @@
 import React from 'react'
-
 import { Menu, Checkbox, Button } from '@pubsweet/ui'
 import { Field } from 'formik'
 import { get } from 'lodash'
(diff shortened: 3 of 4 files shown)


client/elife-theme/src/elements/Menu.js
```diff
@@ -0,0 +1,45 @@
...
+    &:hover {
+      border-color: ${th('colorBorder')};
+    }
+    ${props =>
+      props.open &&
+      `border-bottom-left-radius: 0; border-bottom-right-radius: 0;`};
+  `,
+  Option: css`
+    height: ${th('space.5')};
+    line-height: ${th('space.5')};
```
jsms90: Has there been a decision to use `space` instead of the `fontLineHeightXYZ` variables in our theme?

client/elife-theme/src/elements/Menu.js
```diff
@@ -0,0 +1,45 @@
...
+    border-radius: 0 0 ${th('borderRadius')} ${th('borderRadius')};
+  `,
+  Value: css`
+    &:hover {
+      color: ${th('colorText')};
+    }
+  `,
+  Placeholder: css`
+    font-style: normal;
+    color: ${th('colorText')};
```
jsms90: `colorTextPlaceholder`?

client/elife-theme/src/elements/Menu.js
tamlyn: In this instance line-height is being used for layout (to centre the text in its box) rather than typography.

tamlyn:
All inputs should be 100% width and the parent should control the layout. Also, this change has already been made but I guess you can't see that as the CI is playing up again. Ugh.

hdrury1:
@tamlyn Can you post a screenshot for reference of this change?

tamlyn:
Review deployment now working http://review-menu-overr-m8sm7l.gateway.xpub.semioticsquares.com/submit/metadata

jsms90:
Might be nice to have the scroll bar removed is/when the list isn't long enough to require it
image

Note: this only happens on Firefox

jsms90:
Also, @tamlyn the dropdown icon is still larger than the invision designs.

But we could merge it as is & raise tech debt issues, so @hdrury1 can show to the design team? :woman_shrugging:

jsms90:
Ah, we've just realised that the down arrow is only large on firefox & actually the down arrow looks tiny in Chrome (on my computer only).

Strange things. But seems to be largely browser specific & I have a feeling that pubsweet + firefox doesn't actuallly play that well anyway.

Here's the elifesciences.org homepage on chrome:
image

and here it is on firefox:
image

:woman_shrugging: 

Happy to merge now & raise issues about browser support, wherever necessary

tamlyn:
I'd like to specify that the component and site in general looks fine in Firefox on Mac. It appears to just be Firefox on Linux that is affected by these issues.

client/elife-theme/src/elements/ui/Menu.js
```diff
@@ -0,0 +1,51 @@
+import { css } from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+export default {
+  Arrow: css`
+    color: ${th('colorTextSecondary')};
+    transition: none;
+    font-size: ${th('gridUnit')};
```
dryhten: space?

client/elife-theme/src/index.js
```diff
@@ -6,6 +6,7 @@ const gridUnit = 6
 
 export default {
   /* Colors */
+  colorBackground: '#fff',
```
dryhten: Why did you add this? You don't use it in here (`background-color` is set to `colorPrimary`). We removed it because it was not on `VARIABLES.md` but it is now on the wiki. We don't have it in invision.

tamlyn: It's used in PubSweet components so if we don't have it those components don't render properly. Yet another reason why `VARIABLES.md` should be deleted.

dryhten: I just checked and yes, it's being currently used. Makes sense to add it.

client/elife-theme/src/elements/ui/Menu.js
jsms90: Having 2 variables in our theme which evaluate to the same thing (`gridUnit` and `space.1`) seems less than ideal to me. Also for `calc(...something based on gridUnit...)` vs `space.4`/`space.3`/whatever.

Seems like we need a team-wide convention. Perhaps `gridUnit` is fine for now @dryhten?

tamlyn:
@hdrury1 could you please re-review this? It's blocked until you do because you requested changes.

Issue #266, opened 2018-06-29
Consolidate review deployments into one place


hdrury1:
**User Story:**

_As a member of eLife staff I want a single link in order to see all of the project deveopments in one place._

**Product requirements/Acceptance criteria:**

Functional:

Demo site should show all of the visual changes that have been made (and will be made).
Demo site should encompass all of the functionality changes that have been made (and will be made).
Demo site should display date of last update and confirmation that this is the most recent deployment.

jsms90:
@hdrury1 Tam has just shown me this.

Follow this link, to see our environments in gitlab: https://gitlab.com/elifesciences/elife-xpub/environments.

We have 2 environments: "staging" & "demo"
![image](https://user-images.githubusercontent.com/15656538/42085487-77fc8f4a-7b88-11e8-8188-5e2a6c05ec58.png)


Staging - corresponds directly to what is in our `develop` branch. So click on the little "open" icon (on the right hand side), to see everything that has been merged via our PRs:
![image](https://user-images.githubusercontent.com/15656538/42085499-80d3a2fc-7b88-11e8-9b16-858a310096d4.png)

But, as @tamlyn said, we want to demo what we've done via a different environment, so we can keep pushing to `develop` (staging environment) & not break whatever you're demo-ing while you're demo-ing it :+1:  Hence, getting the "demo" environment in line with "staging" has to be done manually.

To do this, click on the "deploy to.." icon (on the RHS of the "staging" environment
![image](https://user-images.githubusercontent.com/15656538/42085596-cd2da7c4-7b88-11e8-8d0f-867a9fa54d24.png)

@tamlyn has literally just done this.

So now, if you click on the "open" icon on the RHS of the demo environment, you should see our app, with all the currently merged changes. All ready for you to show to Nick & Juliano :tada: 

Hope that helps

hdrury1:
Yes it helps. Except I can't get past the first screen of the wizard (Next doesn't take me to the next page).

jsms90:
@hdrury1 This pre-fill issue isn't happening on the demo site as far as I can see: https://xpub-elife-demo.gateway.xpub.semioticsquares.com/

@tamlyn has raised a PR to fix the 'Next' problem.

Can we close this issue? All it amounts to is saying "we need all the PRs merged" & isn't a separate body of work that needs to be done, unless I'm misunderstanding something about this "second demo site" that we have?

hdrury1:
[...]

PR #278, opened 2018-06-30
WIP: Update docs and add version to the footer

diversemix:
#### Background

Add some more documentation as to how the different environments work.
Add a version number into the footer

#### Any relevant tickets
#266


Dockerfile
```diff
@@ -32,6 +32,7 @@ COPY webpack webpack
 ENV NODE_ENV "production"
 ARG CI_COMMIT_SHA
 ENV CI_COMMIT_SHA ${CI_COMMIT_SHA}
+RUN echo "{ \"version\": \"${CI_COMMIT_SHA}\", \"date\":\"$(date)\" }" > app/version.json
```
g-sam: this is not necessary. cf. https://github.com/elifesciences/elife-xpub/blob/develop/webpack/webpack.production.config.js#L41-L43

tamlyn: Is it confusing to have this information injected into the build at two different points (here and in the webpack build)? I think either is fine but I would prefer to have only one.

docs/deploying/infrastructure.md
```diff
@@ -0,0 +1,39 @@
+# Infrastructure
+
+**Project Name:** elife-xpub
+
+---
+
+All the necessary infrastructure is within [AWS under the elifesciences account](https://512686554592.signin.aws.amazon.com/console)
+
+This is managed using [builder](https://github.com/elifesciences/builder) using the convension where the "Project Name" is appended with the enivronment to give the "Stack Name". This is done in the form:
```
g-sam: convention, environment

app/index.js
```diff
@@ -1,5 +1,27 @@
...
+/*
+The Footer component is a temporary stop-gap while we are still
+undergoing heavy development to ensure we know what we are looking at!
+*/
+const Footer = styled.div`
+  margin: 40px;
+  text-align: right;
+  color: DarkGrey;
+  font-size: 14px;
+  border-top: 1px solid LightGrey;
```
tamlyn: Better to put this at the end of the `Routes` component (outside the `Switch`). This is the entry point file that mounts the app and it's confusing (to me) to have other components here.

app/index.js
```diff
@@ -1,5 +1,27 @@
...
+`
+
+ReactDOM.render(
+  <React.Fragment>
+    <App />
+    <Footer>
+      Date: {date} - Commit: {version}
```
tamlyn: Since this is user facing, I would use a short commit ID (i.e. first 7 characters).

app/version.json
```diff
@@ -0,0 +1 @@
+{}
```
tamlyn: Maybe put some placeholder values in here so that the footer makes a bit more sense when in development.

g-sam: but it should be taken off config, so better to add a default value to the config

Dockerfile
diversemix: This looks to be config and not baked in on build

diversemix: I vote for this method - as config can be changed :) ... as its meant to be!

app/index.js
diversemix: I thought it best to be here and out-of-place so people know that its temporary and should get removed.

app/version.json
diversemix: @tamlyn good idea. @g-sam bad idea :)

Dockerfile
g-sam: I think tam is referring to the fact that we also inject this into the html as a comment, not that it's taken off config.

The sha is added to config and then baked into the build as a comment, or here instead if you wish.

app/index.js
diversemix: ... of course there is the fact it could become a permanent feature :thinking:

Dockerfile
diversemix: I don't understand... where is it injected into the html? I certainly cannot see it on any of the sites.

hdrury1: Injecting into the HTML is not enough. Product and other stakeholders should not be expected to inspect source- this information needs to be readily available and immeadiately visible.

diversemix: Thanks for pointing this out, but we do need this information as it is required for the footer. This is there so that people within eLife who test the product can have something to reference when things go wrong.

g-sam:
There seem to some misunderstandings in the chain of comments on this PR. There are two issues:
 1. we don’t want to bake this information into the app in two places
 2. we can either take this information from config or version.json

The point being:
 1. If we want a footer, the comment should be removed
2. If we want to use version.json, the sha should be removed from config

diversemix:
Let me respond:
1) Why don't we want the SHA in two places? Seems a good idea one for the config and one for the docker container.
2) Which comment should be removed?
3) Why should the SHA be removed from the config?

tamlyn:
- It's useful to have this in the HTML source as well as displayed in the page footer because
   1. If the app completely fails to load you cant see the footer but you can see the source
   2. The footer version will eventually be removed but the source version can remain
- It's confusing to have two independent pieces of code doing the same thing

Therefore, can we agree to update the `HtmlWebpackPlugin` parameters in `webpack.production.config.js` to use the values from `app/version.json`?

g-sam:
Yes!

g-sam:
I have approved at @hdrury1’s request. But I believe the changes suggested by Tam will avoid future confusion.

PR #280, opened 2018-07-02
test: add end-to-end test assertions for manuscript stage transition

tamlyn:
#### Background

- Add a test to assert that the manuscript goes into the QA stage after initial submission
- Refetch the dashboard query on submit so that the data on the dashboard is up to date. It's not yet clear what the best approach is for this.
- Create a page object for the file uploads page test
- Add a wait in the saving test which was sometimes failing because it reloaded the page before the save had completed

With TestCafe.


diversemix:
Didn't know why the e2e tests failed in the pipeline -the failures seemed odd. So I thought I'd run it and it looks like it worked!

tamlyn:
Hmmm, that's a different test flaking now.

PR #291, opened 2018-07-03
fix: remove stage from ManuscriptInput type

tamlyn:
#### Background

The "auto save on unmount" change in #282 introduced a new bug: when the user clicks next, that triggers a mutation and when that succeeds it navigates to the next form page which unmounts the form and triggers an auto save. That's a problem if the previous mutation changed some values on the server as the auto save will overwrite them with the last known values on the client. 

This was caught by the end to end tests introduced in #280 since the change of stage to `QA` was being reverted.

This PR fixes the immediate issue by removing the ability for the client to directly change the `stage` (a change which needs to happen anyway) but this needs to be fixed at a lower level since it could affect other fields too.


diversemix:
Ok this is failing with the other issue: (as per #295 )

```

 ✖ Happy path

   1) Cannot obtain information about the node because the specified
      selector does not match any node in the DOM tree.

      Browser: HeadlessChrome 67.0.3396 / Linux 0.0.0

         89 |  await t.click('[data-test-id=next]')
         90 |
         91 |  // metadata
         92 |  await t
         93 |    .expect(Selector('[name=title]').value)
       > 94 |    .eql(manuscript.title)
         95 |    .click('[role=listbox] button')
         96 |    .click(Selector('[role=option]').nth(0))
         97 |    .click(Selector('label[for=subject-area-select'))
         98 |    .pressKey('enter')
         99 |    .pressKey('down')

```

What say ye? @tamlyn @hdrury1  ... shall I just merge it?

tamlyn:
Currently we have our input types corresponding to the "normal" types i.e. we have `Manuscript` and `ManuscriptInput`. Instead we should have our input types corresponding to our mutations so we have `updateManuscript()` and `UpdateManuscriptInput`. Then it's easy to make sure the input type has only the fields that are relevant to that mutation.

That's a coarse way of preventing writes to things which are supposed to be read-only. A more fine grained approach would be to use authorisation per field. I don't know whether we can get away with only applying authorisation per resolver or whether some mutations will need to support different access controls depending on the role of the user making the request. I guess time will tell.

However that is tangential to the underlying issue here which is that the form data is being saved twice with the later save potentially overwriting changes made by the earlier save. I don't think we currently have a situation where a non-read-only field is modified on the server in such a way that this would cause a problem. But it's unusual behaviour and seems like a likely source of future bugs.

diversemix:
@tamlyn Agree with what you said ... But the fine-grain approach I don't think would work with an authorisation. This seems a little wrong as the context is on the user not on the application. Seems to me knowing the context of the API call is important - this has been solved with different APIs in REST ... but how is this solved in GraphQL ?

PR #293, opened 2018-07-03
fix: add missing params to update mutation

tamlyn:
#### Background

`isAutoSave` wasn't being passed through resulting in emails being sent when the form was auto saved rather than only when the "Next" button was clicked.


diversemix:
I keep trying with test:e2e :(

PR #307, opened 2018-07-05
test: split submission flow into two tests

tamlyn:
#### Background

The "Happy path" test was getting too complex and jeopardising its happiness.

This simplifies the happy path and moves diversions and special cases into other tests. One of these includes a test for the notification ribbon being displayed.

Also create page objects for every page to improve maintainability of the tests.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -76,10 +76,11 @@ test:e2e:
     - cd ${HOME}
     # specify host here else it confuses the linked postgres image
     - export PGHOSTADDR=127.0.0.1
-    - npm run test:e2e -- ---screenshots test/screenshots --screenshots-on-fails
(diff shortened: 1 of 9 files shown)


.gitlab-ci.yml
```diff
@@ -29,7 +29,7 @@ build:
     - echo "Ignore warning! Cannot perform an interactive login from a non TTY device"
     - docker push $IMAGE_ORG/$IMAGE_NAME:$CI_COMMIT_SHA
 
-lint:
+.lint:
```
damnhipster: We shouldn't disable linting in this PR

tamlyn: No, of course. But I'm hammering this pipeline trying to debug the failures and this saves a few precious seconds each time.

Issue #309, opened 2018-07-05
Fetch ORCID user details on login


tamlyn:
**User Story:**

_As an author or submitter I want my details in xpub to be up to date with my details in orcid in order that emails can be sent to me and the corresponding author check can be carried out._

**Product requirements/Acceptance criteria:**

Functional:

- When a user logs in with ORCID, perform an API request to ORCID to fetch the user's email address and save this on the xpub user object.
- Do this even when an existing xpub user is found, updating the previous value. This is because the email address may have changed since their last login.
- This data could be used for the form prefill, bypassing the current API request.

Currently this API request can be quite slow (> 5 seconds). Perhaps this is just an issue with the ORCID sandbox. If not we will need to make this request in the background to avoid slowing down the login too much.

diversemix:
@tamlyn @hdrury1 

Just had a chat to @gmaciocci about this...

The important thing to remember is that during Initial Submission an author and a submitter only have an email address via their relationship to the manuscript. This does not get stored against their profile. Later on when the author creates an profile - we may need this (?).

We do not need this piece of work to be carried out yet. Not sure how this found its way into `In Progress`  let alone `ToDo` - I will leave this open in case I have got the wrong-end-of-the-stick on this.

tamlyn:
I think we do need something like this now:
- The way to detect whether the submitter is different from the corresponding author (and trigger the verification logic) is to compare their email addresses. 
- We don't have a field on the form for submitter email address so I think we need to fetch this from ORCID. 
- The submitter may not use the prefill button at all so we need to fetch their email address separately.

It doesn't have to be on login. Other options:
- Fetch the email on demand each time it is required. 
  - Guaranteed up to date 👍
  - Less robust as network request may fail,  ORCID API token may have expired or been revoked 👎
  - Slows down any operation that requires an email 👎
  - Can't find DB users by email address 👎
  - More complex to implement 👎 
- Fetch on submission start 
  - Doesn't help for user types who never submit anything 👎
- Use webhooks to get notified when ORCID email changes
  - Guaranteed up to date 👍
  - Insufficient on its own so needs to be combined with a way of fetching the initial email 👎

jsms90:
> We do not need this piece of work to be carried out yet. Not sure how this found its way into `In Progress`  let alone `ToDo`

[...]

Issue #240, opened 2018-06-22
Don't send email on form submission


tamlyn:
**User Story:**

As a submitter, I don't want to receive a confirmation email after I complete my initial submission.

**Product requirements/Acceptance criteria:**

From @hdrury1 in https://github.com/elifesciences/elife-xpub/issues/153#issuecomment-397272716:
> Upon form submission, the submitter SHOULD NOT receive an email. Instead, they should be redirected to their dashboard where they will see their most recent submission listed at the top of the page with its status updated to 'Awaiting Quality Checks'.

dryhten:
What's the status of this issue? Is it assigned to anyone? Also, what I'm understanding from this and what Jen did in https://github.com/elifesciences/elife-xpub/pull/315 is that now we only send an e-mail to corresponding author if it's not the same as the submitter, is that true?

jsms90:
Yes, sorry @dryhten. It became clear that it was easier to complete this issue whilst doing #310. So PR #315 solves both issues

jsms90:
Also, to answer your question. Yes, that's exactly correct.

The Confirmation/Success email that we used to send to the submitter every time they clicked 'Submit' (which I believe you were working on before your holiday?) is now being removed by this issue, based on Hannah's request that you can see in the top comment.

A completely separate email is supposed to be sent to the Corresponding Author, if and only if, the Corresponding Author & the Submitter are different people. This was already added by Tam against issue #154 & you reviewed the PR which added it :+1: 

But at the end of last week, the product requirements for the email verification feature changed, so that we're now sending the author this email on 'Submit' rather than 'Next'. Which is why PR #315 removes the code that Tam added & also removes the notification ribbon from the wizard.

To see a full breakdown of the tasks associated with the email verification feature, see issue #272 & to see where the product requirements from this feature were first sent to us, see issue #204.

Happy to chat about this in person, if all these github issues are too confusing?

PR #315, opened 2018-07-06
Email corresponding author on Submit instead of Next

### LINKED ISSUE #310: Email correpsonding author on 'Submit' instead of 'Next'

- Given that we are now only emailing the corresponding author once the submission has been completed, there is no need for a notification ribbon to appear in the wizard.

- The ribbon should appear on the submission once the submitter has hit 'Submit', but there are currently no final designs for this view.

- For the purposes of demonstrating functionality, could we add a temporary status to the submission on the dummy dashboard view to feed back to the user that an email has been sent to the corresponding author?

**Tasks**
- [x] Remove code that adds corresponding author to the manuscript object on clicking 'Next' & move email sending into `finishSubmission` resolver
- [ ] ~Move corresponding author notification ribbon from submission wizard to dashboard~ Remove notification ribbon from submission wizard (or the codebase altogether)?

jsms90:
#### Background
Product requirements now have the corresponding author emailed on Submit, rather than on clicking Next in step 1 of the wizard.

The corresponding author no longer needs to be added to the manuscript object in the middle of the submission process (`updateSubmission` resolver). Differentiating between the submitter email address & the corresponding author email address is to be completed in another ticket - issue #309 

#### What does this PR do?
- Removes code that adds corresponding author to the manuscript object on clicking 'Next' & associated tests
- Moves the sending of an email from the 'updateSubmission' resolver to the `finishSubmission` resolver  
**Note: This email is currently sent every time a user clicks 'Submit'. In order to check whether it should be sent, we first need to store the submitter's email address somewhere** - see issue #309 
- Removes the notification ribbon from the wizard - decision on how/where to display this to the user is still pending
- Removes the 'Confirmation' email that was being sent to the submitter whenever they clicked 'Submit'  (issue #240) - necessary to do within this PR, in order for tests to pass

#### Any relevant tickets
fixes #310 #240 

Added tests for `finishSubmission` resolver:
- check that email isn't sent if submitter = corresponding author
- check that email is sent if submitter & corresponding author are different

### DIFF
--- a/server/submission/resolvers.js
+++ b/server/submission/resolvers.js
@@ -142,10 +142,10 @@ const resolvers = {
           logger.error(`Error sending submitter confirmation email: ${err}`)
         })
 
-      if (ctx.user.email !== manuscript.submissionMeta.author.email) {
+      const user = await User.find(ctx.user)
+      if (user.email !== newManuscript.submissionMeta.author.email) {
         mailer
           .send({
-            from: config.get('mailer.from'),
             to: newManuscript.submissionMeta.author.email,
             text: 'Please verify that you are a corresponding author',
           })
--- a/server/submission/resolvers.js
+++ b/server/submission/resolvers.js
@@ -33,56 +33,6 @@ const staticManifest = `<dar>
 </dar>
 `
 
-function manuscriptHasAuthor(user, manuscript) {
-  return manuscript.manuscriptPersons.some(
-    manuscriptPerson =>
-      manuscriptPerson.role === 'AUTHOR' &&
-      manuscriptPerson.alias.email === manuscript.submissionMeta.author.email,
-  )
-}
-
-async function setupCorrespondingAuthor(user, manuscript) {
-  if (!manuscriptHasAuthor(user, manuscript)) {
-    const manuscriptPerson = await createCorrespondingAuthor(manuscript)
-
-    manuscript.manuscriptPersons.push(manuscriptPerson)
-
-    if (!manuscriptPerson.user) {
-      try {
-        await mailer.send({
-          to: manuscript.submissionMeta.author.email,
-          text: 'Please verify that you are a corresponding author',
-        })
-      } catch (err) {
-        logger.error(err)
-      }
-    }
-  }
-
-  return manuscript
-}
-
-async function createCorrespondingAuthor(manuscript) {
-  const manuscriptPerson = {
-    alias: manuscript.submissionMeta.author,
-    role: 'AUTHOR',
-    metadata: {
-      corresponding: true,
-    },
-  }
-
-  try {
-    manuscriptPerson.user = await User.findByEmail(
-      manuscript.submissionMeta.author.email,
-    )
-    manuscriptPerson.metadata.confirmed = true
-  } catch (err) {
-    if (err.name !== 'NotFoundError') throw err
-  }
-
-  return manuscriptPerson
-}
-
 const resolvers = {
   Query: {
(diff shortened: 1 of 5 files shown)


server/submission/resolvers.js
```diff
@@ -189,16 +130,31 @@ const resolvers = {
...
+          .send({
+            from: config.get('mailer.from'),
+            to: newManuscript.submissionMeta.author.email,
+            text: 'Please verify that you are a corresponding author',
+          })
+          .catch(err => {
+            logger.error(
+              `Error sending corresponding author verification email: ${err}`,
+            )
+          })
```
diversemix: Do you think its worth pulling out a helper function for mailing?

diversemix:
Looks like there maybe a valid reason these tests are failing @jsms90 ... see https://gitlab.com/elifesciences/elife-xpub/-/jobs/80305362

server/submission/resolvers.js
jsms90: Probably not worth it. I refactored this code before I realised that we want to remove the first email anyway - see issue #240 

Since we only have one instance in which we want to send an email (at the moment), I figure we might as well not abstract yet?

jsms90:
Thanks @diversemix! I should have removed those tests, as they're testing functionality that this PR removes :+1:

server/submission/resolvers.test.js
```diff
@@ -278,8 +206,20 @@ describe('Submission', () => {
       const expectedManuscript = lodash.merge(initialManuscript, manuscript)
       expectedManuscript.submissionMeta.stage = 'QA'
       expect(expectedManuscript).toMatchObject(storedManuscript)
+    })
+
+    it('sends Corresponding Author a verification email', async () => {
```
tamlyn: Also add a test that it doesn't send an email if the submitter and author emails are equal?

jsms90: `if (ctx.user.email !== newManuscript.submissionMeta.author.email)` always resolves to true, because `ctx.user.email` is undefined. So, currently, an email is always sent.

Checking both of this situations properly will only be possible (or at least meaningful) once we're saving the submitter's details.

jsms90:
@diversemix @tamlyn The following has been added to this PR since you last reviewed:
- Moves the sending of an email from the 'updateSubmission' resolver to the finishSubmission resolver
**Note: This email is currently sent every time a user clicks 'Submit'. In order to check whether it should be sent, we first need to store the submitter's email address somewhere - see issue #309**
- Removes the notification ribbon from the wizard - decision on how/where to display this to the user is still pending
- Removes the 'Confirmation' email that was being sent to the submitter whenever they clicked 'Submit' (issue #240) - necessary to do within this PR, in order for tests to pass

I've also updated the top comment, in case anyone comes to this afresh, but thought it would be useful for you both to see here

server/submission/resolvers.test.js
tamlyn: In the tests you can set up the user to have the appropriate email and therefore not send the mail. I would say that the fact it doesn't work in the app (and therefore can't be end-to-end tested) is all the more reason to fully unit test it.

tamlyn:
If you rebase onto `develop` again, there's a new end-to-end test which asserts that the ribbon exists. Sorry! This will need to be removed.

jsms90:
Hahaha finally added my comment to that PR to say so, without realising it was already merged. Silly me :upside_down_face:

server/submission/resolvers.test.js
jsms90: Done! :grin:

PR #344, opened 2018-07-11
Add procedure for testing after chat with David

diversemix:
#### Background

Adds the procedure for ui component testing

### DIFF
--- a/docs/developing/ui-component-testing.md
+++ b/docs/developing/ui-component-testing.md
@@ -5,7 +5,7 @@ you have created will be in tip-top shape for the site.
 
 ## Browsers to check
 
-It doesn't have to look exactly the same in every browser, but it should look similar and work properly in current versions of these evergreen browsers:
+It doesn't have to look exactly the same in every browser, but it should look similar and work properly in current versions of these browsers:
 
 ### Desktop
 
@@ -28,5 +28,7 @@ Latest versions of:
 ## Procedure to follow
 
 * Visually check the html of the component.
-* Audibly check the description of the component.
+* Audibly check the description of the component, where possible with:
+  * Voice Over on a Mac while using Safari, or
+  * Chrome Vox Chrome extension using Chrome on any platform
 * Check the CSS is applied correctly.
--- a/docs/developing/ui-component-testing.md
+++ b/docs/developing/ui-component-testing.md
@@ -1,0 +1,32 @@
+# UI Component Testing
+
+This document gives you a process to follow to check that any UI components
+you have created will be in tip-top shape for the site.
+
+## Browsers to check
+
+It doesn't have to look exactly the same in every browser, but it should look similar and work properly in current versions of these evergreen browsers:
+
+### Desktop
+
+* Firefox
+* Opera
+* Chrome
+* Edge
+* IE 11
+* Safari (latest and latest but one)
+
+### Mobile
+
+Latest versions of:
+
+* iPhone: Chrome and Safari
+* iPad: Chrome and Safari
+* Samsung phone: Chrome and Firefox
+* Google phone: Chrome and Firefox
+
+## Procedure to follow
+
+* Visually check the html of the component.
+* Audibly check the description of the component.
+* Check the CSS is applied correctly.


diversemix:
@davidcmoulton Here is the proposal

docs/developing/ui-component-testing.md
```diff
@@ -0,0 +1,32 @@
...
+you have created will be in tip-top shape for the site.
+
+## Browsers to check
+
+It doesn't have to look exactly the same in every browser, but it should look similar and work properly in current versions of these evergreen browsers:
+
+### Desktop
+
+* Firefox
+* Opera
```
tamlyn: Is it really worth testing in Opera considering it uses the same Blink engine as Chrome and it has such a small user bases?

docs/developing/ui-component-testing.md
```diff
@@ -0,0 +1,32 @@
...
+* Chrome
+* Edge
+* IE 11
+* Safari (latest and latest but one)
```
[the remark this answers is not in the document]
davidcmoulton: IE11's not evergreen either.

tamlyn: Are we talking major versions here?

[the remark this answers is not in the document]
diversemix: Shall I just take "evergreen" out?

docs/developing/ui-component-testing.md
```diff
@@ -0,0 +1,32 @@
...
+Latest versions of:
+
+* iPhone: Chrome and Safari
+* iPad: Chrome and Safari
+* Samsung phone: Chrome and Firefox
+* Google phone: Chrome and Firefox
+
+## Procedure to follow
+
+* Visually check the html of the component.
```
tamlyn: I don't understand what this means. Is it about inspecting the DOM? Or just checking the the component renders correctly?

docs/developing/ui-component-testing.md
```diff
@@ -0,0 +1,32 @@
...
+* Edge
+* IE 11
+* Safari (latest and latest but one)
+
+### Mobile
+
+Latest versions of:
+
+* iPhone: Chrome and Safari
+* iPad: Chrome and Safari
```
tamlyn: Is it worth checking Chrome on iOS considering it's just a WebKit view like Safari?

docs/developing/ui-component-testing.md
```diff
@@ -0,0 +1,32 @@
...
+* Samsung phone: Chrome and Firefox
+* Google phone: Chrome and Firefox
```
tamlyn: Is this because you have experienced Chrome/Firefox behaving differently on phones made by different manufacturers?

docs/developing/ui-component-testing.md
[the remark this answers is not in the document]
tamlyn: I was just querying this but was surprised to learn that latest Safari _is_ now evergreen. IE def isn't though.

docs/developing/ui-component-testing.md
```diff
@@ -0,0 +1,34 @@
...
```
tamlyn: There are many browsers (and older versions of browsers) that have higher usage than Firefox mobile's 0.2% user share.

docs/developing/ui-component-testing.md
davidcmoulton: I've observed differences between Chrome and Safari on iOS, even since Chrome changed to use the Webkit view. Main areas are around differing UI responsiveness, and how length calculation assumptions can differ when full screen is involved.

docs/developing/ui-component-testing.md
davidcmoulton: Samsung Chrome is not the same as Google Chrome.

docs/developing/ui-component-testing.md
davidcmoulton: Yes, major versions.

docs/developing/ui-component-testing.md
davidcmoulton: Good point. Replace Firefox with UC Browser.

docs/developing/ui-component-testing.md
diversemix: @davidcmoulton ... my understanding was checking the HTML was well formed in the DOM - what was your intention for this?

docs/developing/ui-component-testing.md
diversemix: @davidcmoulton ...?

jsms90:
I am a little confused by the naming of this file. It says "testing" but the "Procedure to follow" says:
- Visually check the html of the component
- Check the CSS is applied correctly

So is this file referring to manual testing? And are we suggesting that whenever an engineer creates a new component, they should be manually opening every one of these browsers, in web and mobile, & looking it over by eye?

diversemix:
@jsms90 ... that's how it was done in the olden days :grin:

docs/developing/ui-component-testing.md
tamlyn: Ugh, I should have guessed 🙄

diversemix:
This document has been updated after the meeting. @tamlyn / @damnhipster will work on semi-automation.

Issue #328, opened 2018-07-10
change co-submission box


jsms90:
**Designs**
![image](https://user-images.githubusercontent.com/15656538/42562412-92ae4dbe-84f3-11e8-8507-b3e921e49cc4.png)
https://projects.invisionapp.com/share/N3LUQ8S5ARV#/screens/304489195

**Tasks**
- Remove 2nd input field
- Add Action link & functionality which generates new input field
- Maximum of two co-submissions - when 'include' is clicked, the action text ("Would you like to include another co-submission") hides & a second "Article" input field appears
- Second co-submission field is labelled as optional

tamlyn:
Current progress is on https://github.com/elifesciences/elife-xpub/tree/cosubmission

Still to do: 
 - [ ] update tests to use new data model
 - [ ] add the button to create the second field

tamlyn:
Sam's suggestion for validating arrays works:

```
cosubmission: yup
      // validate the array as an object to allow different requirements at different positions
      .object({
        '0': yup.object().shape({
          title: yup.string().required('Article title is required'),
        }),
        '1': yup.object().shape({
          title: yup.string(),
        }),
      })
      // transform array to object for the purpose of validation
      .transform((_, originalValue) =>
        originalValue.reduce(
          (obj, value, index) => ({ ...obj, [index]: value }),
          {},
        ),
      ),
```

diversemix:
Think the schema needs to be: 
```
const schema = yup.array().of(yup.object().shape({
		"title": yup.string().required('Article title is required'),
	})).max(2, "Cannot have more than 2 cosubmissions")
```

PR #358, opened 2018-07-12
Update Cosubmission design

diversemix:
#### Background

Changes the design of the cosubmission entries

#### Any relevant tickets
#328 

Locally


app/components/submission/ManuscriptMetadata/ManuscriptMetadata.js
```diff
@@ -63,45 +45,65 @@ const ManuscriptMetadata = ({ values, setFieldValue, setFieldTouched }) => (
...
-          label="Reference No"
-          name="submissionMeta.cosubmissionId"
-        />
-      </div>
-    </CheckboxGeneratedChild>
+      {values.submissionMeta.cosubmission.map((_, rowIndex) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Box key={rowIndex} mb={2}>
+          <ValidatedField
+            label="Article title"
```
tamlyn: I think we said that the label for the second field should have "(optional)" appended.

diversemix:
@tamlyn would appreciate you look over this again ... I've improved the yup validation (well it wasn't actually working before - hence the Next button not working) ... Also the tests for the second cosubmission - the selectors in `metadata.js` are used by adding a "namedAs" attribute onto the respective components. Not sure this is entirely the right way to do this - please comment :)

app/components/submission/ManuscriptMetadata/ManuscriptMetadataSchema.js
```diff
@@ -9,29 +9,23 @@ const schema = yup.object().shape({
...
+      .nullable(),
+    cosubmission: yup
+      // validate the array as an object to allow different requirements at different positions
+      .array()
+      .of(
+        yup.object().shape({
+          title: yup.string().required('Article title is required'),
+        }),
+      )
+      .max(2, 'Cannot have more than 2 cosubmissions'),
```
tamlyn: This would make sense but as the form is currently designed, this validation message would not be displayed anywhere. This is because it is validating the `cosubmission` key but we only display errors under fields and there is no field on the form with that key. This means the form would fail validation but the user would not know why (as you have already experienced).

app/components/submission/ManuscriptMetadata/ManuscriptMetadataSchema.js
```diff
@@ -9,29 +9,23 @@ const schema = yup.object().shape({
...
-      }),
+      .notOneOf([''], 'Article title is required')
+      .nullable(),
+    cosubmission: yup
+      // validate the array as an object to allow different requirements at different positions
+      .array()
+      .of(
+        yup.object().shape({
+          title: yup.string().required('Article title is required'),
+        }),
```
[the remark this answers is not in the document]
diversemix: Talking to @hdrury1 about this - and it is raising the question about the UI ... there may be a change so putting this back to WIP

app/components/submission/ManuscriptMetadata/ManuscriptMetadata.js
diversemix: Yup @hdrury1 confirmed

hdrury1:
**Further product recs**

- Checkbox label should read 'This manuscript is part of a co-submission'
- Checking this box should generate an input field, label 'Second article title'
- Clicking the 'add another co-submission' link should generate a new input field, label 'Third article title (optional)
- If the checkbox is ticked, there MUST be input in the first field
- Input in the second field is optional, and should not be validated against

**Designs**

Updated invision prototype screens (click the 'include' link): 
https://invis.io/X9LHWJ3BHMY#/304489195_wizard-Screen-3

diversemix:
@tamlyn ... made the updates after doing the debugging we started. Also the changes @hdrury1 requested. Please review.

app/components/submission/Mutations.js
```diff
@@ -0,0 +1,93 @@
...
+        ... on AuthorMetadata {
+          confirmed
+        }
+      }
+    }
+  }
+`
+
+export const GET_CURRENT_SUBMISSION = gql`
+  query CurrentSubmission {
```
tamlyn: This file contains both queries and mutations. I don't think it's worth splitting them into different files. Perhaps `operations.js` would be a better name for the file?

app/components/submission/WithCurrentSubmission.js
```diff
@@ -1,110 +1,69 @@
...
+
+/*
+ * Data Modifiers -
+ *    CosubmissionModifier
+ *
+ * These classes provide a common interface for the main class
+ * 'WithCurrentSubmission' to delgate transforming the data model to the
+ * view model.
+ */
+class CosubmissionModifier {
```
tamlyn: The data modifier concept is good. But given we don't use classes anywhere else in the codebase (except when forced to do so on some React components) I would avoid them here too for consistency. A plain object with three function properties would do the trick.

app/components/submission/Mutations.js
diversemix: sgtm

app/components/submission/WithCurrentSubmission.js
diversemix: Yeah - this is an interesting one. Without sparking a huge conversation - I am very much pro-class sort of a guy :) ... and certainly when looking at other react code - including PubSweet and other xpub apps, I would like to go the same way.... But I get your point!

app/components/submission/Mutations.js
damnhipster: In other places, we name the file `queries.js`, which I think is more apt.

Issue #331, opened 2018-07-11
Turn off Kubernetes cluster during out of work hours


damnhipster:
Turn it off between 8pm and 8am. This should at most halve the cost.

diversemix:
I think this is worth doing with along with #281

hdrury1:
Sounds sensible.

PR #361, opened 2018-07-13
Scaling down the cluster out of hours

damnhipster:
#### Background

Addresses #331 

Uses [pubsweet/kubernetes](https://gitlab.coko.foundation/pubsweet/infra/blob/master/kubernetes/Dockerfile) and [Gitlab CI](https://gitlab.com/elifesciences/elife-xpub/pipeline_schedules) to scale down the Kubernetes cluster at 8pm, and scale up the cluster at 8am on weekdays. 

After the operation, a rolling update will be performed to balance deployments across the cluster, to balance out resource usage, CPU being the main concern. This takes about an hour.

The scheduled jobs can be [run manually](https://gitlab.com/elifesciences/elife-xpub/pipeline_schedules), which has been done.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -121,8 +121,6 @@ review:
   stage: review
   tags:
     - elife-kubernetes-runner
-  except:
-    - schedules
   variables:
     PACKAGE_NAME: xpub-elife
     FORCE_FRESH_DB: "yes"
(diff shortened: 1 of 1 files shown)


g-sam:
conventional commits!

diversemix:
Thanks for looking @giorgiosironi  - there was originally a question of where the code that controlled the cluster should live. That is probably why you were included.

Issue #312, opened 2018-07-06
Allow the server to push messages to the client


tamlyn:
**User Story:**

_As a user I want the application to update as soon new information is available._

**Options**

- [pubsweet-sse](https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/sse) combined with [UpdateSubscriber](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/client/src/components/UpdateSubscriber.jsx).
- GraphQL subscriptions
  - The most common way to do this is with WebSockets and [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws)
  - Another option is SSE and [subscriptions-transport-sse](https://github.com/CodeCommission/subscriptions-transport-sse) 

I would favour GraphQL subscriptions over `pubsweet-sse` in order to be consistent with our technology choices and also because the latter has a bunch of legacy behaviour which we don't need or want. I favour SSE over WebSockets because it is a simpler technology and requires less fiddling with firewall rules etc, however in this instance the corresponding package is much less popular and would need to be carefully audited before use. The two subscription transport packages are meant have the same API so it *should* be easy enough to switch if the need arose.

**Product requirements/Acceptance criteria:**

Functional:
- Implement one of the above options to allow the pushing of messages from the server to the client

diversemix:
@tamlyn I would tend to agree to using GraphQL as its the technology of the API which is where the pub-sub model should be implemented. Do we need an ADR for this?

dryhten:
Update:

I decided to skip pubsweet-sse, since it uses REST API and it would be a nice thing to have GraphQL subscription support in pubsweet as well. Looking at the options for GraphQL subscriptions:

1. subscriptions-transport-sse is using some deprecated API from graphql-subscriptions so it doesn't really work. We could try to fix it but I think that would be too much work.
2. subscriptions-transport-ws - I managed to get this running locally (not in pubsweet though). Right now there is a bug that needs fixing at [!337](https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/337). However, it seems to work just fine in GraphiQL.

dryhten:
I managed to make it work for now with `subscription-transport-ws`. There is still some work to be done to make the code more maintainable and the changes in pubsweet are currently in review.

[...]

PR #368, opened 2018-07-16
feat(submission): add uploadProgress subscription resolver

dryhten:
re #312


server/submission/resolvers.js
```diff
@@ -215,6 +227,18 @@ const resolvers = {
       return manuscript
     },
   },
+  Subscription: {
+    uploadProgress: {
+      subscribe: () => {
+        console.log('new subscription!!')
+        setTimeout(() => {
+          console.log('published new data')
+          pubsub.publish(ON_UPLOAD_PROGRESS, { uploadProgress: 10 })
```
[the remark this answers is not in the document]
dryhten: You actually have to return it like that. This code is here just to debug and see if it works or not, but I tested this with GraphiQL and after 8 seconds I get back the right stuff and it doesn't complain.

This link has some subscription implementation https://github.com/apollographql/graphql-subscriptions

app/components/ui/molecules/FileUpload.test.js
```diff
@@ -51,14 +51,14 @@ it('displays success if conversion.completed is set', () => {
...
 
-it('displays uploading if conversion.converting is set', () => {
+it.skip('displays uploading if conversion.converting is set', () => {
   const dropzoneContentWrapper = makeCheerioWrapper({
     conversion: { converting: true },
   })
   expect(dropzoneContentWrapper.text()).toBe(manuscriptUploading)
 })
 
-it('displays uploading even if there are errors', () => {
```
tamlyn: Did you mean to skip these? If so, please leave a comment to explain.

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -10,6 +14,8 @@ const path = require('path')
 const crypto = require('crypto')
 const Joi = require('joi')
 
+const { ON_UPLOAD_PROGRESS } = asyncIterators
```
tamlyn: Is it confusing that this `ON_UPLOAD_PROGRESS` is a string but [the other thing](https://github.com/elifesciences/elife-xpub/pull/368/files#diff-77d316c88c9f296c72d8fd6f9740d4f6R21) called `ON_UPLOAD_PROGRESS` is a graphql operation?

server/xpub/entities/manuscript/resolvers.test.js
```diff
@@ -281,9 +281,13 @@ describe('Submission', () => {
...
         encoding: 'utf8',
         mimetype: 'application/pdf',
+        size: 73947,
       }
-
-      const manuscript = await Mutation.uploadManuscript({}, { id, file })
+      const manuscript = await Mutation.uploadManuscript(
+        {},
+        { id, file },
+        file.size,
```
tamlyn: Is this right? Shouldn't file size be passed in the second parameter?

app/components/pages/SubmissionWizard/steps/Files/index.js
```diff
@@ -29,28 +35,33 @@ const FileUploadsPage = ({
...
-              setFieldValue('meta.title', data.uploadManuscript.meta.title)
-              setFieldValue(fieldName, data.uploadManuscript.files)
-            })
-          }
-          previewUrl={`/manuscript/${values.id}`}
-          setFieldValue={setFieldValue}
-          {...props}
-        />
+        <Subscription subscription={ON_UPLOAD_PROGRESS}>
+          {({ data: uploadData, loading: uploadLoading }) => (
```
tamlyn: What happens if this request fails? Does the progress just stay at zero? If so, I think that's fine. Just would like to make sure it doesn't fail in any more major way.

server/xpub/entities/manuscript/resolvers.js
dryhten: Not really, one is client side and the other server side. They had similar examples in apollo as far as I remember.

app/components/pages/SubmissionWizard/steps/Files/index.js
dryhten: I just played with this and if the initial subscription fails then the page does not load at all. The error doesn't seem to be captured in the `error` render prop as described here https://www.apollographql.com/docs/react/advanced/subscriptions.html#render-prop

server/xpub/entities/manuscript/resolvers.test.js
dryhten: I don't remember what happened here, pretty sure I tested a while ago. This will fail anyway now because you need to subscribe to uploadProgress before you upload a file. I would skip for now and fix the test later since merging this seems like a higher priority.

app/components/ui/molecules/FileUpload.test.js
dryhten: Yes, same as my previous comment, tests needs updating to account for percentage of file upload. Would skip for now.

app/components/pages/SubmissionWizard/steps/Files/index.js
dryhten: https://github.com/elifesciences/elife-xpub/issues/492

server/xpub/entities/manuscript/resolvers.test.js
dryhten: https://github.com/elifesciences/elife-xpub/issues/493

app/components/ui/molecules/FileUpload.test.js
dryhten: https://github.com/elifesciences/elife-xpub/issues/494

PR #389, opened 2018-07-19
Check the xmlData has an article before parsing

diversemix:
#### Background

Not fall over if Sciencebeam is not available

#### Any relevant tickets

None

Not yet!

### DIFF
--- a/server/submission/resolvers.js
+++ b/server/submission/resolvers.js
@@ -177,10 +177,17 @@ const resolvers = {
         fs.writeFile(manuscriptJatsPath, xmlString),
         fs.writeFile(manuscriptManifestPath, staticManifest),
       ])
-      const title =
-        xmlData.article.front[0]['article-meta'][0]['title-group'][0][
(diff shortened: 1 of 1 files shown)


server/submission/resolvers.js
```diff
@@ -177,10 +177,17 @@ const resolvers = {
...
+
+      let title = ''
+      if (xmlData.article) {
+        const firstArticle = xmlData.article.front[0]
+        const { 'article-meta': articleMeta } = firstArticle
+        const firstMeta = articleMeta[0]
+        const { 'title-group': titleGroup } = firstMeta
+        const firstTitleGroup = titleGroup[0]
+        const { 'article-title': titleArray } = firstTitleGroup
+        ;[title] = titleArray
```
tamlyn: Good thing eslint is here to help you write nice code like this, right?

diversemix: yeah ... was a "learning curve"

Issue #227, opened 2018-06-19
Create "Person Pod" component


dryhten:
## Select Button
**User Story:**

_As a submitter I want to access the people picker from the submission wizard in order to select editor suggestions._

_As an editor I want to access the people picker from the article information screen in order to view submitter suggestions and assign editors and reviewers to the submission._

**Product requirements/Acceptance criteria:**

Functional: 

- Suggestions are required, but exclusions are not and this should be reflected in the text on the button (see wireframes)
- '+' icon on the right-hand side of the button should open the people picker modal.

**Designs**
![image](https://user-images.githubusercontent.com/15656538/42771010-afeff340-891d-11e8-98a7-26452f43024d.png)

---

## Person Pod
**Product requirements/Acceptance criteria:**
- The '+' icon is the only clickable part of the pod, indicated with a pointer change on hover

- Clicking the icon adds a tick to the pod and also adds the person's name as a tag under the search box

**Design(s):**
Component in the [invision styleguide](https://invis.io/D8KXOML3C6N#/302926985_components):
![image](https://user-images.githubusercontent.com/15656538/42823444-7141fffe-89d5-11e8-9f30-57a2c9dcf91d.png)

![group](https://user-images.githubusercontent.com/17023702/42878192-12c33484-8a84-11e8-948f-d4ed62149199.png)

The width of this component is dependent on its parent

**"Chosen" version - states:**
![image](https://user-images.githubusercontent.com/15656538/42822273-8afe02ba-89d2-11e8-9052-ae52b5c1f034.png)
![image](https://user-images.githubusercontent.com/15656538/42820332-cbf4b3f4-89cd-11e8-9c06-b5f6d3af202a.png)
![image](https://user-images.githubusercontent.com/15656538/42821826-590d0888-89d1-11e8-9dac-455ba27fb48e.png)
![image](https://user-images.githubusercontent.com/15656538/42824728-54defaee-89d8-11e8-9620-4774649d4b13.png)

Future states to consider (but not necessarily build yet):
![image](https://user-images.githubusercontent.com/15656538/42824680-3558157a-89d8-11e8-9f59-6e14aff2adf3.png)
![image](https://user-images.githubusercontent.com/15656538/42825366-c6f9baa0-89d9-11e8-8d88-34fae3340ffa.png)
![image](https://user-images.githubusercontent.com/15656538/42824612-1339b868-89d8-11e8-95cb-fe4415add3bc.png)

jsms90:
[...]

PR #390, opened 2018-07-19
feat: people picker selection functionality

### LINKED ISSUE #354: Ability to select person(s) within the picker modal

- Within the picker modal, a user should use the '+' icon inside a person pod to make a selection

- The '+' icon is the only clickable part of the pod, indicated with a pointer change on hover

- Clicking the icon adds a tick to the pod and also adds the person's name as a tag under the search box

- The number of selections required should be indicated and updated to reflect selections made ('2 suggestions required', then '1 suggestion required' then 'Maximum 2 choices')

- A submitter can make selections up to the maximum number, indicated by success text (2 x Senior editors, 2 x Reviewing editors)

- User can deselect by clicking the 'X' on a tag or the tick on a person pod. This restores the pod to its previous state (with '+' icon) and removes the tag.

https://invis.io/PYMEMIF2THG#/305823520_picker-Author-Selection

### LINKED ISSUE #355: Impose limit on number of person selections and display feedback
**Product requirements**

- There is a maximum number of selections that can be made:

2 x suggested Senior editors
2 x excluded Senior editors
2 x suggested Reviewing editors
2 x excluded Reviewing editors

- Feedback should be given to the user (2 more suggestions required, 1 more suggestion required, Maximum 2 choices)
- When the maximum is reached the feedback is shown in success colour
- For exclusions, no 'required' message is displayed, but when the maximum is reached a 'Maximum 2 choices' message is displayed in success colour

**Designs**

[Suggestions required](https://invis.io/9HLUEH9AXRF#/305823519_picker-Author)
[Maximum reached](https://invis.io/9HLUEH9AXRF#/305823520_picker-Author-Selection)

tamlyn:
#### Background

Implement the `PeoplePicker` component which ties together the selected item tags, the selection hint message, the person pod search results and the button bar.

Closes #354 
Closes #355 

~~Uses a dummy person pod implementation pending #227.~~

Visually with styleguide and functionally with Jest.

### DIFF
--- a/app/components/ui/molecules/PeoplePicker.js
+++ b/app/components/ui/molecules/PeoplePicker.js
@@ -93,7 +93,10 @@ const PeoplePickerButtons = ({ isValid, onCancel, onSubmit }) => (
 )
 
 class PeoplePicker extends React.Component {
-  state = { selection: [] }
+  constructor(props) {
+    super(props)
+    this.state = { selection: this.props.initialSelection }
+  }
 
   toggleSelection(person) {
     if (this.isSelected(person)) {
@@ -134,21 +137,24 @@ class PeoplePicker extends React.Component {
     })
   }
 }
+const peopleArrayPropType = PropTypes.arrayOf(
+  PropTypes.shape({
+    id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+    name: PropTypes.string.isRequired,
+  }),
+)
 
 PeoplePicker.propTypes = {
+  initialSelection: peopleArrayPropType,
   minSelection: PropTypes.number,
   maxSelection: PropTypes.number,
   onCancel: PropTypes.func.isRequired,
   onSubmit: PropTypes.func.isRequired,
-  people: PropTypes.arrayOf(
-    PropTypes.shape({
-      id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
-      name: PropTypes.string.isRequired,
-    }),
-  ).isRequired,
+  people: peopleArrayPropType.isRequired,
 }
 
 PeoplePicker.defaultProps = {
+  initialSelection: [],
   minSelection: 0,
   maxSelection: Infinity,
 }
--- a/app/components/ui/molecules/PeoplePicker.js
+++ b/app/components/ui/molecules/PeoplePicker.js
@@ -1,0 +1,159 @@
+/* eslint-disable react/no-unused-prop-types */
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Button } from '@pubsweet/ui'
+import { th } from '@pubsweet/ui-toolkit'
+import { Box, Flex } from 'grid-styled'
+
+import SelectedItem from '../atoms/SelectedItem'
+import PersonPod from '../atoms/PersonPod'
+
+const SelectedItems = ({ selection, onCloseClick }) => (
+  <Flex>
+    {selection.map(person => (
+      <Box key={person.id} mr={1}>
+        <SelectedItem
+          label={person.name}
+          onCloseClick={() => onCloseClick(person)}
+        />
+      </Box>
+    ))}
+  </Flex>
+)
+
+const SuccessMessage = styled.div`
+  color: ${th('colorSuccess')};
+`
+
+const SelectionHint = ({ selection, minSelection, maxSelection }) => {
+  const selectionLength = selection.length
+  if (selectionLength < minSelection) {
+    const numRequired = minSelection - selectionLength
+    return (
+      <div>
+        {numRequired} more suggestion{numRequired === 1 ? '' : 's'} required
+      </div>
+    )
+  }
+  if (selectionLength >= maxSelection) {
+    return <SuccessMessage>Maximum {maxSelection} choices</SuccessMessage>
+  }
+  return null
+}
+
+const PeoplePickerBody = ({
+  isSelected,
+  maxSelection,
+  minSelection,
+  onCancel,
+  onSubmit,
+  people,
+  selection,
+  toggleSelection,
+}) => (
+  <React.Fragment>
+    <Flex justifyContent="space-between">
+      <SelectedItems onCloseClick={toggleSelection} selection={selection} />
+      <SelectionHint
+        maxSelection={maxSelection}
+        minSelection={minSelection}
+        selection={selection}
+      />
+    </Flex>
+    <Flex flexWrap="wrap" mx={-2}>
+      {people.map(person => (
+        <Box key={person.id} p={2} width={1 / 2}>
+          <PersonPod.ChosenPersonPod
+            iconState={isSelected(person) ? 'selected' : 'add'}
+            institution={person.institution}
+            isIconClickable={
+              isSelected(person) || selection.length < maxSelection
+            }
+            name={person.name}
+            onIconClick={() => toggleSelection(person)}
+          />
+        </Box>
+      ))}
+    </Flex>
+  </React.Fragment>
+)
+
+const PeoplePickerButtons = ({ isValid, onCancel, onSubmit }) => (
+  <Flex>
+    <Box mr={3}>
+      <Button onClick={onCancel}>Cancel</Button>
+    </Box>
+    <Box>
+      <Button disabled={!isValid} onClick={onSubmit} primary>
+        Add
+      </Button>
+    </Box>
+  </Flex>
+)
+
+class PeoplePicker extends React.Component {
+  state = { selection: [] }
+
+  toggleSelection(person) {
+    if (this.isSelected(person)) {
+      this.setState({
+        selection: this.state.selection.filter(p => p.id !== person.id),
+      })
+    } else {
+      this.setState({ selection: this.state.selection.concat(person) })
+    }
+  }
+
+  handleSubmit() {
+    if (this.isValid()) {
+      this.props.onSubmit(this.state.selection)
+    }
+  }
+
+  isSelected(person) {
+    return this.state.selection.some(p => p.id === person.id)
+  }
+
+  isValid() {
+    const selectionLength = this.state.selection.length
+    return (
+      selectionLength >= this.props.minSelection &&
+      selectionLength <= this.props.maxSelection
+    )
+  }
+
+  render() {
+    return this.props.children({
+      ...this.props,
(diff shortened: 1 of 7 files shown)


app/components/ui/molecules/PeoplePicker.js
```diff
@@ -0,0 +1,159 @@
...
+    return (
+      <div>
+        {numRequired} more suggestion{numRequired === 1 ? '' : 's'} required
+      </div>
+    )
+  }
+  if (selectionLength >= maxSelection) {
+    return <SuccessMessage>Maximum {maxSelection} choices</SuccessMessage>
+  }
+  return null
```
jsms90: What do you think of having the `PeoplePickerBody` being the one responsible for showing/hiding this message, rather than `SelectionHint` returning an element in some situations and returning nothing in other situations?

I feel like I should be able to look at `PeoplePickerBody`, without knowing anything about how `SelectionHint` works, and understand what kind of form it will take. As it is, I would expect _something_ to be rendered where it says `SelectionHint`.

To me, this is one of the benefits of thinking about building from an Atomic Design approach - composing from the ground up means that each component makes sense in total isolation.

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -0,0 +1,159 @@
...
+    }),
+  ).isRequired,
+}
+
+PeoplePicker.defaultProps = {
+  minSelection: 0,
+  maxSelection: Infinity,
+}
+
+PeoplePicker.Body = PeoplePickerBody
```
jsms90: It's probably good to specify the prop types for these sub components too, right? After all, we're giving people a way of rendering them, separately from the `PeoplePicker` alone

app/components/ui/molecules/PeoplePicker.md
```diff
@@ -0,0 +1,49 @@
...
+    <React.Fragment>
+      <PeoplePicker.Buttons {...props} />
+      <hr />
+      <PeoplePicker.Body {...props} />
+    </React.Fragment>
+  )}
+</PeoplePicker>
+```
+
+### Body only
```
jsms90: Just looking at this component in styleguidist, before looking at the code, this combination of props doesn't really show everything that the `Body` component can do & is responsible for.

Right now, it looks like the icon buttons are always disabled in "Body", which isn't the case, and therefore people don't get to see that "Body" is also responsible for showing/hiding the `SelectedItem` tags & the `SelectionHint` message.

Perhaps we add `isSelected` & `toggleSelection` handlers that change state in the same way that we do in our app, in order to demonstrate this functionality?

app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -0,0 +1,101 @@
...
+  { id: 3, name: 'Chastity' },
+  { id: 4, name: 'Dave' },
+]
+
+const makeWrapper = props =>
+  mount(
+    <ThemeProvider theme={theme}>
+      <PeoplePicker
+        // eslint-disable-next-line react/no-children-prop
+        children={innerProps => <PeoplePicker.Body {...innerProps} />}
```
jsms90: Do we need this? Rather than disabling the eslint rule, we could just explicitly nest the `PeoplePicker.Body` inside the `PeoplePicker` couldn't we?

app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -0,0 +1,101 @@
...
+    const render = jest.fn(() => null)
+    makeWrapper({ children: render })
+    expect(render).toHaveBeenCalled()
+  })
+
+  it('renders one person pod per person', () => {
+    const wrapper = makeWrapper()
+    const buttons = wrapper.find('PersonPod')
+
+    expect(buttons).toHaveLength(4)
```
jsms90: How about `people.length` instead of hardcoding `4`?

app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -0,0 +1,101 @@
...
+  expect(wrapper.find('SelectedItem')).toHaveLength(length)
+
+describe('PeoplePicker', () => {
+  it('calls child render prop', () => {
+    const render = jest.fn(() => null)
+    makeWrapper({ children: render })
+    expect(render).toHaveBeenCalled()
+  })
+
+  it('renders one person pod per person', () => {
```
jsms90: This is probably extremely pedantic, but the wording made me think that you were actually going to check whether there was a single pod for "Annie" pod & a single pod for "Bobby", whereas this test would still pass even if 4 "Annie"s were rendered.

How about "renders as many pods as there are people"? Again, this may feel overly nitpick-y, but if it made me think it was doing something other than it was doing, it might just help someone who is using the failing test's message down the line? :woman_shrugging:

app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -0,0 +1,101 @@
...
+    const wrapper = makeWrapper()
+    const buttons = wrapper.find('PersonPod')
+
+    expect(buttons).toHaveLength(4)
+  })
+
+  it('updates selection', () => {
```
jsms90: Perhaps this could be more specific? Like "clicking the icon adds/removes a selected item from the page"?

Because "updates selection" is actually testing a few things all at once:
- whether the icon changes from `add` -> `remove` -> `add` again
- whether clicking the + button adds a person to the `selected` array
- whether clicking the bin button removes a person from the `selection` array
- whether the appearance of the `SelectedItem`s are correctly connected to the `selection` array

If this test fails, by the name of the description, I would assume this was failing because of the connection between the button click & the selection array. I may not thinking that this test was related to the appearance/disappearance of the `SelectedItem`s

app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -0,0 +1,101 @@
...
+    expectSelectionLength(wrapper, 0)
+
+    getPersonPodButton(wrapper, 0).simulate('click')
+    expectSelectionLength(wrapper, 1)
+
+    getPersonPodButton(wrapper, 0).simulate('click')
+    expectSelectionLength(wrapper, 0)
+  })
+
+  it('disables inputs when maximum is reached', () => {
```
jsms90: Similar to my comment above - this test is relying on a lot of things at once, many of which aren't being tested individually. So the description could potentially be misleading.

For example, this test would pass even if the icon on that particular pod was changing to a Tick, so long as the selection array wasn't being updated & there was no `SelectedItem` on the page.

So the description of this test would be a bit of a false friend, as the "inputs" (button clicks?) could theoretically still be enabled & the icons still changing accordingly.

app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+    getPersonPodButton(wrapper, 0).simulate('click')
+    wrapper
+      .find('PeoplePicker')
+      .instance()
+      .handleSubmit()
+    expect(onSubmit).toHaveBeenCalled()
+  })
+
+  it('shows selected items', () => {
```
jsms90: The test also includes the removal of a selected item too.

Perhaps it would be good to separate this into 2 tests - one for showing & one for hiding? Because it might be nice to add assertions that check whether clicking a person pod adds/removes the _right_ `SelectedItem`?

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -0,0 +1,159 @@
...
+    if (this.isValid()) {
+      this.props.onSubmit(this.state.selection)
+    }
+  }
+
+  isSelected(person) {
+    return this.state.selection.some(p => p.id === person.id)
+  }
+
+  isValid() {
```
jsms90: What do you think of making this name more explicit? Like `hasReachedMaxLimit`?

app/components/ui/atoms/PersonPod.md
```diff
@@ -63,6 +63,20 @@ Selected icon:
 />
 ```
 
+Disabled icon:
```
jsms90: Could you remove this section: https://github.com/elifesciences/elife-xpub/pull/390/files#diff-9002cb506702fd5d27099698b6102d2bR80 or add the new `isIconClickable` prop to that?

app/components/ui/molecules/PeoplePicker.js
tamlyn: Probably wouldn't hurt although if you look at the examples, you're expected to spread the argument from the render prop function into these components. Doing that allows the user to not care what those props are and us (the authors) to change the API in future without breaking most people's usages. I've update the docs to mention this.

app/components/ui/molecules/PeoplePicker.test.js
tamlyn: We do need to do it this way in order to override the `children` prop in some tests. I'll add a comment.

app/components/ui/molecules/PeoplePicker.test.js
tamlyn: Yeah. Initially I was asserting against the internal `selection` state of the component so this made sense. But after having to wrap the whole thing in a `ThemeProvider` I can no longer do that :( Name updated.

app/components/ui/molecules/PeoplePicker.js
tamlyn: No I think this is valid (boom, boom). This is testing whether the picker selection is in a valid state – like a form. Currently that involves checking max and min but in future that could change.

jsms90: Then perhaps the name should change in the future too? :wink:

But seriously, if we're using a name that makes it sound like you can put whatever you want in there, perhaps this is an indication that we should allow people to add their own function to `isValid`?

Rather than this being a stateful component, perhaps we should make `isValid`, `isSelected`, `handleSubmit`, etc all just inputs to a functional component?

app/components/ui/molecules/PeoplePicker.js
jsms90: Sure :+1: 

I was imagining that we were exposing `PeoplePicker.Body`, separately from `PeoplePicker`, so that someone could use the body by itself if they wanted. But I see now that the separation is purely because of `grid-styled` - so that we can put `Flex` & `Box` elements between `PeoplePicker` & its children?

We're not actually expecting/suggesting that anyone uses `PeoplePicker.Body` independently from `PeoplePicker`. Makes more sense now. Thanks.

Issue #351, opened 2018-07-12
Connect people picker modal to app


hdrury1:
**User Stories:**

_As a submitter I want to see a list of editors, their institutions and their subject area(s) in order to choose the most appropriate people to oversee the review of my work or to exclude those who may have a COI._

_As a Deputy/Senior editor, I want to see a list of Senior/Reviewing editors alongside their current availability, expertise and the authors' suggestions in order to assign the most appropriate team to the paper._

_As a Reviewing Editor, I want to see the authors' suggested reviewers to help me decide on who to invite to review._

_As a RE, I want to see how long a reviewer usually takes to write a review as this may impact whether or not I invite them to review a particular paper._

_As an author, I'd like the system to offer me a browsable list of editors I can select from, e.g., by key words, major subject area_

**Product requirements/Acceptance criteria:**

- When a selection has been made, the person pod should replace the Select Button component within the submission form.

- ~~The person pod should be populated with data from eLife API `/people` endpoint.~~ see #401 

- The amount of data displayed in each pod depends on the privileges of the user selecting the person(s):

- _Submitter should see_

First Name, Last Name
Institution
Keywords

_Editor/Staff should see_

First Name, Last Name
Institution
Keywords
Status (see separate ticket)

- ~~If any information is missing, the line should be left blank but not omitted (i.e. overall size of the pod would not be affected)~~ see #404 

- ~~If institution is missing, the line should be replaced with email address (e.g. a reviewer suggested by the submitter that is not a user in the database)~~ see #403 

- Clicking `+` on select pod should bring up people picker
- 'Cancel' button should close the people picker modal with no updates to the Submission wizard/Assignment Tab
- 'Add' should only become active and clickable once the required number of selections has been made
- Clicking 'Add' closes the modal and replaces the previously clicked selection button component with the chosen person pod(s)
- ~~Navigating back in the browser should be the same as clicking 'Cancel'~~ see #402 

**Relevant links to design(s):**

[Invision Prototype](https://invis.io/Z9MDY2CPXE6)

jsms90:
@hdrury1 I've changed the name of this issue because the product requirements describe how the person pod component will interact with the page. This is separate from the creation of the Person Pod atom in the first place. For that work, see issue #227

PR #392, opened 2018-07-19
create PersonPod (minus button clicking functionality - to come in next PR)

jsms90:
### What does this PR do?
- add 3 new icons to our icons folder - plus, rubbish bin & selected tick
- adjust existing `Icon.js` component so that `color` is inherited instead of `black`
- create styled versions of the paragraph element: `RegularP`, `SmallP` (& `SmallAction`) - should these become their own atoms? In elife-xpub or in pusbweet?
- create two versions of the "PersonPod"
  - `ChoosePersonPod`
  - `ChosenPersonPod`

Going with this implementation, a "PersonPod" is simply a styled container, which is responsible for positioning its children.

### Any relevant tickets
#227 

### How has this been tested
Visually for styling

Clicking functionality does not yet work - see below

### Existing issues
Passing props into pubsweet's `Button` component. This affects:
- clicking functionality for button on RHS of pod (`Button.js`)
- clicking functionality for keywords (`Action.js` - which uses `Button.js`)
- font styling for keywords e.g. size

I'm unsure about exporting `PersonPod` as a whole, or quite how this works with `prop-types`? It would make more sense to me, to be exporting just `ChosenPersonPod` & `ChoosePersonPod`, but this seems to break styleguidist :cry:


app/components/ui/atoms/PersonPod.js
```diff
@@ -0,0 +1,117 @@
...
+  justify-content: space-between;
+
+  background: ${th('colorSecondary')};
+  border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorBorder')};
+  border-radius: ${th('borderRadius')};
+  color: ${th('colorText')};
+  font-family: ${th('fontInterface')};
+  font-size: ${th('fontSizeBase')};
+  min-width: 312px;
+  height: 128px;
```
tamlyn: Where do these values come from?

app/components/ui/atoms/PersonPod.js
```diff
@@ -0,0 +1,117 @@
...
+    )}
+    {!isKeywordClickable && <RegularP>{keywords}</RegularP>}
+    {isStatusShown && <SmallP>{status}</SmallP>}
+  </Box>
+)
+
+const ChosenPersonPod = props => (
+  <PersonPod
+    icon={ChangingIcon(props.iconState)}
+    textContainer={ChosenPerson(props)}
```
tamlyn: I find this a bit confusing where ChangingIcon and ChosenPerson are defined as components and start with an uppercase letter, but are used as functions. In this case it makes no functional difference but not all functions are components and not all components are functions.

app/components/ui/atoms/PersonPod.js
```diff
@@ -0,0 +1,117 @@
...
+  />
+)
+
+const Chooser = ({ role, isRequired, ...props }) => (
+  <Flex flexDirection="column" justifyContent="center">
+    <Box ml={2}>
+      {isRequired && <RegularP>Choose {role} (required)</RegularP>}
+      {!isRequired && <RegularP>Choose {role} (optional)</RegularP>}
```
tamlyn: Could you simplify this by only putting the condition on the required/optional text?

app/components/ui/atoms/PersonPod.js
jsms90: Guesswork. But I've just seen Guiliano has replied in Slack - https://elifesciences.slack.com/archives/C6LKMFK5Z/p1531910953000087

app/components/ui/atoms/PersonPod.js
```diff
@@ -0,0 +1,117 @@
+import React from 'react'
+import styled from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+import { Action, Button } from '@pubsweet/ui'
+import { Flex, Box } from 'grid-styled'
+
+import Icon from './Icon'
+
+const RegularP = styled.p`
```
diversemix: What is 'P' ?

app/components/ui/atoms/PersonPod.js
diversemix: I'd vote for `ChangingIcon` -> `podIcon`
`ChosenPerson` -> `podPerson`

app/components/ui/atoms/PersonPod.js
diversemix: Ah - is it for "paragraph"? If so where its used maybe its better to name like `StatusStyle` or `KeywordStyle` ?

app/components/ui/atoms/PersonPod.js
tamlyn: [Pod people!](http://www.darkcrystal.com/galleries_podpeople.php)

image

diversemix: Are you thinking ..... Easter Egg?

app/components/ui/atoms/PersonPod.js
tamlyn: Makes sense for height but does it need a min-width? Seems like it is likely to cause issues in some as yet unforeseen responsive layouts.

app/components/ui/atoms/PersonPod.js
jsms90: Hahahahahahaa, pod people. These films are trippy!

@diversemix So funny that you would say that because `PodIcon` was actually the original name that I gave to it! But I changed it to `ChangingIcon` because using a verb rather than a noun felt like it better described what this bit of code is doing.

I think the lack of clarity in the code is a reflection of the confusion in my brain as I was building it. So thanks so much for this comment @tamlyn, it really helped me pinpoint where my own confusion was coming from! The reason that a verb felt more appropriate is because it's a function rather than a component.

But I was still thinking about `ChangingIcon`/`PodIcon` as though it were a component, rather than a function that creates a component. If you guys could take a look at my latest commit & let me know what you think of these names instead, that would be great :grin:

diversemix: @jsms90 Naming things is always difficult! Looking now.

app/components/ui/atoms/PersonPod.js
jsms90: Yeah I get what you mean @diversemix. But `RegularP` is being used in 3 places, so to use this naming, I would need to create 3 components: `NameStyle`, `InstitutionStyle` and `NonClickableKeywordStyle` all with identical styling on them. (`SmallP` is only used for keywords and could therefore become `KeywordStyle`)

jsms90: I do get the reasoning though - this naming is more explicit in the context of the `PersonPod`.

But if we're going to use these stylings (essentially just a `p` with a specified font size and line height) in other contexts than just the `PersonPod` component, then maybe having a generic name like `SmallParagraph` is better?

If you take a look at the top comment of this PR, I was thinking that maybe we should extract `RegularP` and `SmallP` out into their own atoms? After all, we have `Headings.js` for `h1`, `h2`, etc, which are styled according to our theme. Should we not have the same thing for `p` tags?

I should probably call it `RegularParagraph` instead of `RegularP` though, to avoid the "what is 'P'?" confusion? I agree that this is confusing.

jsms90:
@tamlyn has just found what I've done wrong in writing this component, that explains why the icon wasn't clickable. So there's no issue with pubsweet's button component :+1: & he's changing it whilst doing issue #351 :grin:

PR #398, opened 2018-07-20
refactor person pod, add prop types & tests

jsms90:
#### What does this PR do?
1. Refactor, as per comment in #227:
    - `PersonPod` -> `PersonPodContainer`
    - `ChosenPersonPod` -> `PersonPod`
    - `ChoosePersonPod` -> `PersonPod.SelectButton`

I don't think this is necessarily ideal, as it suggests that `SelectButton` is a sub-component of `PersonPod`, which is not really accurate. Unfortunately, styleguidist does not seem to allow you to export 2 components from the same file? :confused: Suggestions welcome!

2. Specify prop types
3. Add tests
4. Tiny fix - keywords are now a small paragraph rather than regular

#### Any relevant tickets
#227 

Icon clicking & keyword clicking has been tested for `PersonPod`.

Disabling the icon hasn't been tested, since this functionality is coming in a different PR - #390

### DIFF
--- a/app/components/ui/atoms/PersonPod.js
+++ b/app/components/ui/atoms/PersonPod.js
@@ -42,7 +42,7 @@ const StyledPod = styled(Flex)`
   height: 120px;
 `
 
-const PersonPod = ({
+const PersonPodContainer = ({
   isIconClickable,
   onIconClick,
   textContainer,
--- a/app/components/ui/atoms/PersonPod.js
+++ b/app/components/ui/atoms/PersonPod.js
@@ -63,24 +63,23 @@ const PersonPodContainer = ({
   </StyledPod>
 )
 
-const PlusIcon = <Icon size={3}>Plus</Icon>
-const RubbishBinIcon = <Icon size={3}>RubbishBin</Icon>
-const SelectedTickIcon = <Icon size={4}>SelectedTick</Icon>
+const plusIcon = <Icon size={3}>Plus</Icon>
+const rubbishBinIcon = <Icon size={3}>RubbishBin</Icon>
+const selectedTickIcon = <Icon size={4}>SelectedTick</Icon>
 
-const switchIcon = iconState => {
-  switch (iconState) {
-    case 'add':
-      return PlusIcon
+const PodIcon = ({ iconType }) => {
+  switch (iconType) {
     case 'remove':
-      return RubbishBinIcon
+      return rubbishBinIcon
     case 'selected':
-      return SelectedTickIcon
+      return selectedTickIcon
+    case 'add':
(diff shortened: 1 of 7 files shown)


app/components/ui/atoms/PersonPod.js
```diff
@@ -107,12 +117,56 @@ const buildChooserText = ({ role, isRequired, ...props }) => (
...
+  )
+}
+
+PersonPodContainer.propTypes = {
+  onIconClick: PropTypes.func.isRequired,
+  textContainer: PropTypes.element.isRequired,
+  icon: PropTypes.element.isRequired,
+}
+
+switchIcon.PropTypes = {
```
tamlyn: Isn't it `propTypes` with lower case first letter? But also, `switchIcon` is not a component (it's a function that returns a component) so I don't think it makes sense to have prop types at all.

app/components/ui/atoms/PersonPod.js
```diff
@@ -126,18 +133,56 @@ const buildChooserText = ({ role, isRequired, ...props }) => (
...
-const ChoosePersonPod = ({ onIconClick, ...props }) => {
+const SelectButton = ({ onIconClick, ...props }) => {
   const ChooserText = buildChooserText(props)
   return (
-    <PersonPod
+    <PersonPodContainer
       icon={PlusIcon}
       onIconClick={onIconClick}
       textContainer={ChooserText}
     />
```
damnhipster: Rather than passing in `textContainer` as an attribute, you may be able to pass it in as a child

i.e.
```
<PersonPodContainer
  icon={PlusIcon}
  onIconClick={onIconClick}
>
  <ChooserText/>
</PersonPodContainer>
```

`PersonPodContainer` would then use `this.props.children` rather than `this.props.textContainer`.

jsms90: But the `icon` is also a child. So I can't just render all children with `this.props.children`, right?

damnhipster: The icon is an attribute, under `this.props.icon`. It shouldn't be in `this.props.children`.

damnhipster: The idea is that the component has children as it's content, and has attributes which act like modifiers. It makes it easier to understand.

e.g.
```
<Notification dismissable>
  We use cookies on this website. Do you accept our terms & conditions?
  <Accept/>
  <Decline/>
</Notification>
```

jsms90: I'm not sure that I understand what you mean. Why is `textContainer` any different from `icon`?

Issue #347, opened 2018-07-12
Create search/filter box component


hdrury1:
**User Story:**

_As a submitter I want to filter the list of editors in order to narrow down the number I have to choose from._

**Product requirements/Acceptance criteria:**

Functional:

- The search box should be propulated by default with the MSA(s) already selected on page 3 of the submission wizard.
- An entry in the search box should filter down the number of possible selections to choose from. (see ticket)
- A user can filter by MSA, keyword or name. (see  #349)
- Clicking on the search bar enables the user to enter input.
- When a user begins to enter input into the search bar, a dropdown appears with a list of suggested terms based on input. Clicking on one of these suggestions populates the filter box. (see ticket)
- Clicking the search icon or hitting the enter key updates the list of people to reflect the search term(s) entered into the search box.

**Relevant links to design(s):**

https://invis.io/9HLUEH9AXRF#/305823519_picker-Author
https://projects.invisionapp.com/share/FUMJS349R2E#/screens/305823534_staff-Picker-Reviewers-Search

**Relevant files**
[search_icon.zip](https://github.com/elifesciences/elife-xpub/files/2223272/search_icon.zip)

dryhten:
@hdrury1 I have a few questions about this

1. When you click on the search box, do you get a dropdown menu with all the options?
2. When you start typing something, you will get a dropdown list with all the options that match the string you typed. Should this match only happen at the beginning of the string or can it happen anywhere? For example, if I start typing "Mike", do I get all the people who have a first name of Mike or can I can the ones that have a last name of "Mike" too?
3. Do I understand correctly that we can search by name, institution and subject area all in the same search box? If that's the case, is it fine to show all results that matched no matter what field matched? For example, if I search for Max, I would expect to see all the people called Max and also people from the Max Planck Institute.
4. Is anyone supposed to be able to add a new reviewer to the system using `Add new +`? Some people might add gibberish and showing this to other submitters will give a bad experience. Or do we plan to not add the new reviewers entered manually to the global scope? (e.g. save them for other users).
5. As far as I remember, the list of pods won't change as you start typing something, is that correct?

dryhten:
I have another question for this, are we supposed to get the list of people (e.g. reviewers) from an elife api or do we store it in the db ourselves? The same goes for MSA (I assume this is Major Subject Area?) and institutions.

Also, I assume the list of available options that shows up as you type should be done as part of this other issue https://github.com/elifesciences/elife-xpub/issues/349.

[...]

PR #407, opened 2018-07-24
add search box component

dryhten:
re #347


app/components/ui/atoms/icons/Search.js
```diff
@@ -0,0 +1,15 @@
...
+      <g
+        fill="#888888"
+        id="molecules/search-input"
+        transform="translate(-255.000000, -16.000000)"
+      >
+        <path d="M267.5,27 L266.7,27 L266.4,26.7 C267.4,25.6 268,24.1 268,22.5 C268,18.9 265.1,16 261.5,16 C257.9,16 255,18.9 255,22.5 C255,26.1 257.9,29 261.5,29 C263.1,29 264.6,28.4 265.7,27.4 L266,27.7 L266,28.5 L271,33.5 L272.5,32 L267.5,27 L267.5,27 Z M261.5,27 C259,27 257,25 257,22.5 C257,20 259,18 261.5,18 C264,18 266,20 266,22.5 C266,25 264,27 261.5,27 L261.5,27 Z" />
+      </g>
+    </g>
+  </svg>
+)
```
tamlyn: Please put all SVGs through https://svgr.now.sh. For example the above becomes: 
```
<svg viewBox="0 0 18 18" {...props}>
    <path
      d="M12.5 11h-.8l-.3-.3c1-1.1 1.6-2.6 1.6-4.2C13 2.9 10.1 0 6.5 0S0 2.9 0 6.5 2.9 13 6.5 13c1.6 0 3.1-.6 4.2-1.6l.3.3v.8l5 5 1.5-1.5-5-5zm-6 0C4 11 2 9 2 6.5S4 2 6.5 2 11 4 11 6.5 9 11 6.5 11z"
      fill="#888"
    />
</svg>
```

app/components/ui/molecules/SearchBox.js
```diff
@@ -0,0 +1,81 @@
...
+    })
+  }
+
+  getSuggestions = value => {
+    const inputValue = value.trim().toLowerCase()
+    const inputLength = inputValue.length
+    return inputLength === 0
+      ? []
+      : this.props.options.filter(
+          option => option.value.toLowerCase().search(inputValue) >= 0,
```
tamlyn: `String.prototype.search` converts the input to a regular expression which will have unexpected consequences if a search contains any special characters. Did you mean `String.prototype.includes`?

app/components/ui/molecules/SearchBox.js
```diff
@@ -0,0 +1,81 @@
...
+          renderSuggestion={this.renderSuggestion}
+          suggestions={suggestions}
+          theme={theme}
+        />
+      </Root>
+    )
+  }
+}
+
+export default SearchBox
```
tamlyn: How does the people picker (or whatever) hook into this to get the value of the search box?

app/components/ui/atoms/icons/Search.js
dryhten: When was this agreed upon? Is there an ADR about it?

tamlyn: Haha! I mean if you really, really don't like that tool you don't have to use it but it just seems to make sense to commit clean and tidy SVGs.

dryhten: I used that tool and the SVG I get doesn't look like the original icon, at least on the page which is weird. If I save both as svg files and open them with an image editor they both look the same to me, but when I load them on the search bar they look different. I attached images with both cases.
converted-search-svg
search-svg

dryhten: Ok, so it seems to happen because the converter forgot about `stroke='none'` attribute. Open new issue on their repo? :)

dryhten: I think the attributes are a little messed up on the original svg though. It has both `stroke="none"` and `strokeWidth="1"`.

app/components/ui/atoms/icons/Search.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export default props => (
+  <svg viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg">
```
dryhten: I see that https://github.com/smooth-code/svgr has MIT license so I guess we don't have to give credit to them, or should we?

app/components/ui/molecules/SearchBox.js
dryhten: Hm, didn't know about that. I want to know if `option.value.toLowerCase()` has substrings of `inputValue`, and you're right, [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) does the job, thanks!

app/components/ui/atoms/icons/Search.js
tamlyn: It's because you're wrapping the icon in the `Icon` component which applies a stroke to it. By default there is no stroke so the SVG is correct.

app/components/ui/molecules/SearchBox.js
dryhten: I would do that as part of the "integrate search box into the page" issue. I guess we would have to lift the state up? The parent component would keep references to `value` and `suggestions` and pass them down as props.

tamlyn: OK. You don't necessarily need to lift the state. You can just have a `onSubmit` prop which gets called with the current value when `Enter` is pressed or the icon is clicked.

app/components/ui/atoms/icons/Search.js
dryhten: It seems that the Icon component only sets the stroke color. I'm not sure if in the absence of `stroke="none"` this will force a stroke or not. If that's the case then it makes sense what happened, since the original svg has `stroke="none"` which will make it disappear.

What I was saying about the original SVG is that there is no need to set the stroke width if it's nonexistent.

app/components/ui/molecules/SearchBox.js
dryhten: Makes sense, I would still do it as part of that issue.

dryhten:
Oh wow, the pipeline failed for this one for those CSS linting rules :) I will look into it

Issue #401, opened 2018-07-24
Fetch editor data from eLife API


tamlyn:
**Product requirements/Acceptance criteria:**

Functional:
- The person pod should be populated with data from eLife API /people endpoint.

PR #413, opened 2018-07-30
refactor: improve two column layout component

tamlyn:
#### Background

- Allow key to be specified on items
- Add propTypes
- Use `children` prop instead of `items`

#### Any relevant tickets

Spin off from #401 

Visually.


app/components/global/layout/TwoColumnLayout.md
```diff
@@ -0,0 +1,10 @@
+Responsive layout with two columns on wider screens and a single column on
+smaller screens.
+
+```js
+;<TwoColumnLayout>
```
diversemix: Is the semi-colon necessary?

tamlyn: I think in this specific case it isn't necessary as there is nothing before it. But if you were to add a JS statement between lines 4 and 5 it would become necessary.

PR #414, opened 2018-07-30
fix: tweak people picker and person pod display

tamlyn:
#### Background

- Improve person pod display when it has lots of keywords or long institution names
  - Use `text-overflow: ellipsis` as a temporary measure
  - Doesn't support the product requirements of being able to click on the `...`
  - Some `min-width: 0` weirdness due to [flexbox rendering rules](https://css-tricks.com/flexbox-truncated-text/)
- Add commas between subject area keywords in person pod
- Update person pod style guide sample data which was broken in previous PR

#### Any relevant tickets

Another spin off from #401 

Visually.


app/components/ui/atoms/PersonPod.js
```diff
@@ -88,31 +98,40 @@ const PersonText = ({
...
+  const separatedKeywords = keywordList.reduce(
+    (accu, elem) => (accu === null ? [elem] : [...accu, ', ', elem]),
+    null,
+  )
+
+  return (
+    <CollapsibleBox m={2} {...props}>
+      <RegularP>{name}</RegularP>
+      <RegularP>{institution}</RegularP>
+      <SmallP>{separatedKeywords}</SmallP>
```
jsms90: To me, it's not super obvious just by looking at this `return` block, what is actually being returned. Without first looking at what separatedKeywords is & what keywordsList is, I would look at this block and assume that `separatedKeywords` was just a string, sitting inside a paragraph tag. In fact, it's a bunch of `SmallAction`s or `span`s.

I'm wondering if there's a way that we could break this block up into smaller pieces, so that `PersonPodContainer` isn't responsible for all this logic. If we captured each piece of logic in a different component, we could be more explicit on this line.

I'm thinking something like this, so that PersonPodContainer is only responsible for rendering a list of keywords:
```
<CollapsibleBox m={2} {...props}>
      <RegularP>{name}</RegularP>
      <RegularP>{institution}</RegularP>
      <KeywordsList>
     {isStatusShown && <SmallP>{status}</SmallP>}
</CollapsibleBox>
```

Then `KeywordsList` is separate component, responsible for:
1. mapping over `keywords`
1. checking `isKeywordClickable` and rendering either clickable keywords (`<SmallAction>`) or non-clickable keywords (`<p>`)
1. reducing

```
const KeywordsList = (isKeywordClickable, keywords) => {
    keywords
      .map(keyword => isKeywordClickable ? <ClickableKeyword {keyword}> : <NonClickableKeyword {keyword}>)
      .reduce((accu, elem) => (accu === null ? [elem] : [...accu, ', ', elem]), null)
}
```

Then `ClickableKeyword` and `NonClikcableKeyword` are responsible for taking a keyword and rendering a `SmallAction` or `SmallP` respectively & adding the appropriate props.

What do you reckon?

app/components/ui/atoms/PersonPod.js
```diff
@@ -42,6 +48,10 @@ const StyledPod = styled(Flex)`
   height: 120px;
 `
 
+const CollapsibleBox = styled(Box)`
+  min-width: 0;
```
jsms90: Is it worth having a comment here, since both of us found the necessity for this piece of CSS surprising?

app/components/ui/atoms/PersonPod.md
```diff
@@ -66,32 +67,28 @@ Selected icon:
...
-  status="Currently unavailable"
-  isStatusShown={true}
   iconType="add"
   isIconClickable={false}
-  onIconClick={() => console.log('icon clicked')}
 />
 ```
 
-### ChoosePersonPod
+### SelectButton
```
jsms90: Thanks :sweat_smile: that was bugging me.

Could we change the other line too (ChosenPersonPod)? Or maybe just remove it?

jsms90:
Ah I'm too slow :laughing:

Issue #397, opened 2018-07-20
PersonPod Icons should darken on hover


diversemix:
The icons on a PersonPod should darken when the mouse hovers over them. There should be no darkening on the background of the pod - just the button.

I realise this is a little vague - but its here as this is a visual that needs to be specified further, is difficult to explain and is non-obvious from the wireframes.... i.e. don't shoot the messenger!

**Designs**
![group](https://user-images.githubusercontent.com/17023702/42878192-12c33484-8a84-11e8-948f-d4ed62149199.png)

diversemix:
So ... the 48px from rhs is still the button.. hovering over here darkens only the icon.

hdrury1:
@diversemix is correct. Let's shoot for 50% darker icon on hover.

hdrury1:
@jsms90 Pointer should change to a hand on hover too.

PR #420, opened 2018-08-01
change colour of person pod icons on hover

jsms90:
#### What does this PR do?
- use regular HTML `button` element instead of pubsweet's button atom, since we don't need the regular hover behaviour for pubsweet buttons
- change/add padding to RHS of icon, so that spacing between icon and RHS of pod conform to styleguide
- change icon colours on hover using `fill`

Note:
The colour change has to happen when the user hovers over the button (not just the icon itself i.e. the SVG), so we need to change the CSS of the child from within the parent. But the parent (button) doesn't know which child (i.e. which icon) it has inside it. Hence the somewhat ugly `svg > path` & `svg > g > circle`

I considered changing the code structure altogether, to have `StyledPlusButton`, `StyledRubbishButton`, `StyledTickButton`, in order to apply different hover styles based on which icon was inside. But this would cause a lot of duplication.

If anyone has any better ideas, I'm all ears :+1: 

#### Any relevant tickets
#397

### DIFF
--- a/app/components/ui/atoms/PersonPod.js
+++ b/app/components/ui/atoms/PersonPod.js
@@ -37,6 +37,20 @@ const StyledButton = styled.button`
   height: 100%;
   border: none;
   padding: 0 calc(${th('gridUnit')} * 2);
+
+  // careful, this is brittle & relies heavily on the icons' structure
+
+  &:hover {
+    // plus & rubbish icons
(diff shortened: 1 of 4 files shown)


jsms90:
Much more elegant. Silly me. Thanks @tamlyn

Issue #336, opened 2018-07-11
Update suggested reviewers functionality


hdrury1:
**User Story:**

_As a submitter, I would like to suggest appropriate reviewers to assess my submission to eLife_

**Product requirements/Acceptance criteria:**

- Three reviewer name/email rows should be listed by default
- Filling in the first two rows has no effect on the UI
- Focussing on the third row (if the first two rows are already filled) generates a new fouth row below
- Focussing on the subsequent fourth row (if the first three are already filled) generates a fifth row ...and so on, up to a maximum of six rows
- Validation should occur on the first three rows only- the rest are optional input fields
- Currently there is no affordance for deleting rows that have been generated
- Once a sixth row has been generated/filled the user cannot enter any more suggestions

**Relevant links to design(s):**

Screen 1 - two rows filled (https://invis.io/KXMHPB5YRZ8#/305858554_author-Exclusion-Set)
Screen 2 - focus on third row generates a new fourth row (https://invis.io/4GMHPDE89TR#/305873221_author-Add-More-Reviewers)

jsms90:
@hdrury Could we separate this out from the "people picker" epic? It has a fair bit of complexity in it & it's a pretty separate body of work from the people picker modal (even though it exists on the same page of the wizard).

hdrury1:
Done.

diversemix:
@hdrury1 ... for the moment I am not going to do it on focus but when there is data in the field ... There are some horrendous issues with the focus event at the moment.

diversemix:
Requirements have changed for the optional reviewers:
> Morning all. Any queries awaiting me from Friday? peter and tamlyn at a glance, the answer is yes, we should require name+email for the optional reviewer fields. so it’s all or nothing.

https://elifesciences.slack.com/archives/C6LKMFK5Z/p1532939698000180

diversemix:
Further relevant background: 

Tam said:
> There is still *some* validation as we check that any email entered is valid. And I think we said (a long time ago) that it should check that the same email is not entered twice.

https://elifesciences.slack.com/archives/C6LKMFK5Z/p1532940135000069?thread_ts=1532939783.000057&cid=C6LKMFK5Z

GM said:
> Also worth noting that Invalid optional fields should only stop progress to the next stage of a wizard if they’re at least partially filled. If they’re blanked again, they are ignored.

[...]

PR #431, opened 2018-08-02
Add optional reviewers and validate for duplicates

diversemix:
#### Background

Allows up to an addition 3 reviewers - these are optional

#### Any relevant tickets
#336

Not yet - need to change some tests too before final version


diversemix:
Please squash the commits !

app/components/pages/SubmissionWizard/steps/Editors/SuggestedReviewersValidator.js
```diff
@@ -0,0 +1,132 @@
...
+      name: yup.string().required('Name is required'),
+      email: yup
+        .string()
+        .email('Must be a valid email')
+        .required('Email is required'),
+    })
+    this.thingsNotDuplicated = ['name', 'email']
+  }
+
+  checkArrayContents(reviewers, errors) {
```
jsms90: I'm wondering if there's a way to pop all of this `SuggestedReviewsValidator` logic into the component itself? So that there's yup can be responsible only for doing the checking, rather than whether something should be checked?

server/submission/definitions.js
```diff
@@ -18,7 +18,7 @@ const typeDefs = `
       uploadManuscript(id: ID!, file: Upload!): Manuscript!
       finishSubmission(data: ManuscriptInput!): Manuscript!
     }
-    
+
```
jsms90: Is the deletion of all these lines coming from your editor's linting? Personally, I find it easier to read with spaces in between, but either way, we should make it consistent throughout this file

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
     )
   }
 
+  handleSuggestedReviewersChanged = event => {
+    this.props.handleChange(event)
```
jsms90: Unless I'm missing something, I don't think there's a `handleChange` being passed to this component as a prop?

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -41,6 +42,12 @@ const ValidationMessage = ({ message }) => (
 )
 
 class EditorsContent extends React.Component {
+  constructor() {
+    super()
+    this.START_GROW_INDEX = 2
+    this.MAX_REVIEWERS = 6
```
jsms90: Is it necessary for these to be hard coded variables within the constructor?

The concept of _when_ a new line gets rendered could belong to a particular method (e.g. `handleSuggestedReviewersChanged`) & the restriction of "up to 6" could live in the render method?

I don't think `START_GROW_INDEX` is being used anywhere?

dryhten: It's usually not a good idea to declare constants where you are going to use them. They are defined in a constructor or a dedicated place so you can easily change them in the future without going through the code.

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
tamlyn: It's passed down from Formik.

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
     )
   }
 
+  handleSuggestedReviewersChanged = event => {
+    this.props.handleChange(event)
+
+    const MAXIMUM = limits.suggestedReviewers.max
+
+    const itemIsBlank = item => item.name + item.email === ''
+    // const anyBlankMandatory = (revs) => itemIsBlank(revs[0]) || itemIsBlank(revs[1]) || itemIsBlank(revs[0])
```
dryhten: Commented code?

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
...
+
+    const reviewers = this.props.values.suggestedReviewers
+
+    // logic only kicks in for the optional reviewers
+    if (reviewers) {
+      // first count the blanks at the end
+      let numBlanks = 0
+      for (let index = reviewers.length - 1; index > 0; index -= 1) {
```
dryhten: Shouldn't this stop at 2?

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
     )
   }
 
+  handleSuggestedReviewersChanged = event => {
+    this.props.handleChange(event)
+
+    const MAXIMUM = limits.suggestedReviewers.max
```
dryhten: MAXIMUM? More intuitive name like `maxReviewers`? The original name is also intuitive (`limits.suggestedReviewers.max`).

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
...
+
+    const itemIsBlank = item => item.name + item.email === ''
+    // const anyBlankMandatory = (revs) => itemIsBlank(revs[0]) || itemIsBlank(revs[1]) || itemIsBlank(revs[0])
+
+    const reviewers = this.props.values.suggestedReviewers
+
+    // logic only kicks in for the optional reviewers
+    if (reviewers) {
+      // first count the blanks at the end
+      let numBlanks = 0
```
dryhten: What do you think about this?
```
  let numEmpty = 0
  for (let reviewer of [...reviewers].reverse()) {
    if (emptyReviewer(reviewer))
      numEmpty += 1
    else
      break
  }
```

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
...
+      for (let index = reviewers.length - 1; index > 0; index -= 1) {
+        const item = reviewers[index]
+        if (itemIsBlank(item)) {
+          numBlanks += 1
+        } else {
+          break
+        }
+      }
+      // if we have no blanks then add if one if less than max
```
dryhten: This comment is confusing to me :)

diversemix:
Needs to be merged - closing and re-opening another when I have the time

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -246,7 +292,10 @@ class EditorsContent extends React.Component {
           {values.suggestedReviewers.map((_, index) => (
```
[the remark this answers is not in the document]
diversemix: Not sure this is the correct model to follow, maybe you will have to explain more. As the data on the manuscript is not currently a part of state.

app/components/pages/SubmissionWizard/steps/Editors/SuggestedReviewersValidator.js
```diff
@@ -0,0 +1,132 @@
...
+        e.inner.forEach(thrownError => {
+          const error = cloneDeep(thrownError)
+          error.path = `suggestedReviewers.${index}.${error.path}`
+          errors.push(error)
+        })
+      }
+    }
+  }
+
+  static checkArrayForDuplicateKey(key, reviewers) {
```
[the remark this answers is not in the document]
diversemix: Maybe the function would be better named `checkArrayForDuplicateByKey`

app/components/pages/SubmissionWizard/steps/Editors/SuggestedReviewersValidator.js
diversemix: So  the latest PR has all the add/removal done in the UI ... and all the validation in this class.

server/submission/definitions.js
diversemix: yes - it removes unnecessary whitespace - not sure what you mean?

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: removed - thanks

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: No as we use it to see if there is a blank line there

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: done - thanks

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: Feel free to put update the new PR

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: Why?

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -79,6 +86,45 @@ class EditorsContent extends React.Component {
...
+    this.props.handleChange(event)
+
+    const MAXIMUM = limits.suggestedReviewers.max
+
+    const itemIsBlank = item => item.name + item.email === ''
+    // const anyBlankMandatory = (revs) => itemIsBlank(revs[0]) || itemIsBlank(revs[1]) || itemIsBlank(revs[0])
+
+    const reviewers = this.props.values.suggestedReviewers
+
+    // logic only kicks in for the optional reviewers
```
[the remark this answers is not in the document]
diversemix: There are lots of ways to solve this problem :)

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
dryhten: Don't we need 3 suggested reviewers fields present all the time? At least that was my understanding of it. 
Looking at the rest of the code it seems to me that right now we can have a minimum of 2 suggested reviewers fields (since you don't look at the first one no matter what). This seems a little random, I would either keep only 1 suggested reviewer or 3.

diversemix: Yes we do need 3 - with additional optional 3. Did you want to discuss on the merged  #441 , or on slack?

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: Feel free to refactor!

PR #442, opened 2018-08-07
439 integrate search box

dryhten:


dryhten:
As discussed at the morning standup, this should be merged for demo purposes and everything that's left to be done should be added to a github issue.

dryhten:
Ok we discussed this and from a component design perspective it makes sense to have a Modal component with a People Picker inside that has [Search Box, People Pods, Buttons] as part of it. Will merge this in case eLife needs to demo it and will open a new issue to move Search Box inside People Picker.

PR #463, opened 2018-08-09
refactor: reorganise server code by data entity

### LINKED ISSUE #422: Implement ADR 004
Current thoughts are to use the structure in  "ADR 004" (https://github.com/elifesciences/elife-xpub/blob/develop/docs/adr/0004-code-structure.md) where the `data-access.js` returns simple javascript objects from the datastore.

tamlyn:
#### Background

- Implement the spirit (if not the letter) of ADR 004
- Create new `@elifesciences/xpub-server` component which contains the entities folders and some other stuff
- Remove `submission` server component and `db-helpers`
  - There is currently some duplication is the `data-access.js` files which should be addressed
- Split graphql type definitions into separate files
- Stop using `User` model from `pubsweet-server`

#### Any relevant tickets

Closes #422 

Existing tests updated.


server/xpub/entities/manuscript/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+const Manuscript = {
+  MAX_SUGGESTED_REVIEWERS: 6,
+  MIN_SUGGESTED_REVIEWERS: 3,
+
+  find: dataAccess.selectById,
+  all: dataAccess.selectAll,
+  findByStatus: dataAccess.selectByStatus,
+
+  new: () => lodash.cloneDeep(emptyManuscript),
```
damnhipster: `Manuscript.new` seems like it would be better placed to be `Manuscript.create`.

Maybe there should be a `Manuscript` class, where some of these functions return instances of said class. It means we can avoid cloning objects.

```
class Manuscript {
  constructor() {
    // instantiate empty manuscript
  },
  addTeam(roles) {},
  checkPermission() {},
  removeOptionalBlankReviewers() {}
}

const Entity = {
   ...
   create: () => new Manuscript()
   ...
}
```

Is this what you meant by the duplicate code that could be refactored out later?

config/components.json
```diff
@@ -1,6 +1,6 @@
 [
   "@pubsweet-pending/auth-orcid",
-  "@elife-xpub/submission",
+  "@elifesciences/xpub-server",
```
dryhten: Should this be published?

server/xpub/entities/manuscript/helpers/empty.js
```diff
@@ -0,0 +1,41 @@
...
+    articleType: '',
+    subjects: [],
+  },
+  opposedSeniorEditorsReason: '',
+  opposedReviewingEditorsReason: '',
+  opposedReviewers: [],
+  suggestionsConflict: false,
+  files: [],
+  coverLetter: `
+<p><strong>How will your work make others in the field think differently and move the field forward?</strong></p>
```
dryhten: Should this be kept in a file or a different place in the code?

diversemix: That's a great point, and I agree - especially if we use @damnhipster 's suggestion of using a class it could go on there. However, I think as long as the place is consistent for now we can move this later. Good work @tamlyn  - keen to get this in - maybe we can talk about these refinements f2f on Thursday?

config/components.json
tamlyn: Probably doesn't need to be.

Issue #467, opened 2018-08-10
Sort list of People Pods in People Picker alphabetically by default


dryhten:
Should the list of people pods in the people picker be sorted alphabetically by default? (e.g. when the modal is loaded without typing anything)

dryhten:
Oops, forgot to close this issue :)

PR #468, opened 2018-08-10
feat(PeoplePicker): sort list of people in people picker alphabetical…

dryhten:
…ly by default

re #467

### DIFF
--- a/app/components/ui/molecules/PeoplePicker.js
+++ b/app/components/ui/molecules/PeoplePicker.js
@@ -145,19 +145,24 @@ class PeoplePicker extends React.Component {
     this.setState({ searchValue })
   }
 
-  filterPeople() {
+  filterPeople(people) {
     const searchValue = this.state.searchValue.toLowerCase()
-    return this.props.people.filter(person =>
+    return people.filter(person =>
       person.name.toLowerCase().includes(searchValue),
     )
   }
 
   render() {
     const { people, ...otherProps } = this.props
-    const searchOptions = people.map(person => ({ value: person.name }))
+    const sortedPeople = [...people].sort((a, b) =>
+      a.name.localeCompare(b.name),
+    )
+    const searchOptions = sortedPeople.map(person => ({ value: person.name }))
     return this.props.children({
       ...otherProps,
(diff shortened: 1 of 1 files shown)


tamlyn:
The end to end test are failing because they currently select the editors based on index rather than name so the change in sorting means different editors are being selected. When it tries to request info on the new editors, it can't find a network response to replay so it fails. You need to run the tests locally with `REPLAY=record` environment variable then add the resulting files to the commit.

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -145,19 +145,24 @@ class PeoplePicker extends React.Component {
...
+    return people.filter(person =>
       person.name.toLowerCase().includes(searchValue),
     )
   }
 
   render() {
     const { people, ...otherProps } = this.props
-    const searchOptions = people.map(person => ({ value: person.name }))
+    const sortedPeople = [...people].sort((a, b) =>
+      a.name.localeCompare(b.name),
```
damnhipster: Does this run correctly? That following comma looks incorrect...

dryhten: Yes, this runs correctly. That comma was added by either prettier or eslint on the precommit hook, not sure which.

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -145,19 +145,24 @@ class PeoplePicker extends React.Component {
...
+    )
+    const searchOptions = sortedPeople.map(person => ({ value: person.name }))
     return this.props.children({
       ...otherProps,
-      people: this.state.searchValue ? this.filterPeople() : people,
+      people: this.state.searchValue
+        ? this.filterPeople(sortedPeople)
```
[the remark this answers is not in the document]
dryhten: Hm, does it? It's changed anyway in this other PR that should get merged after this one
https://github.com/elifesciences/elife-xpub/pull/469/files#diff-532c4b9d7a19fae6f4736a55158d31b2R224

dryhten:
@tamlyn ok, thanks

Issue #443, opened 2018-08-07
Improve search/filtering functionality for the People Picker


dryhten:
The list of suggestions displayed in the search box should be improved. The matching should happen at the beginning of words and the list should be sorted by relevancy.

dryhten:
Updated issue to reflect the discussion we had at the standup.

hdrury1:
@dryhten Consensus from Product is that the items should be listed by order of relevancy. I know we spoke about this over a call but wanted to record it here too.

dryhten:
@hdrury1 should they be sorted alphabetically too? (e.g. if nothing is typed in the search box)

dryhten:
https://github.com/elifesciences/elife-xpub/issues/467

hdrury1:
If nothing is typed in the box, they should be sorted alphabetically with already selected persons first.

dryhten:
Even if the keywords with the selected people are right under the search box? I was playing with this and it can lead to weird behavior/inconsistencies. What happens if you select something from a page? Should it move to the top right away or when should it move to the top? If you refresh the page/modal with the current implementation you lose all the selections so the only behavior that wouldn't scare the user would be to move them to the top only on new searches which seems a little tedious right now (I would make a new issue for it)

hdrury1:
Let's just have the full list for now and see how that goes down in testing. If we need to tweak the behaviour after this we can do.

dryhten:
Done, right now the matches only happen at the beginning of the words for single word inputs. For inputs containing multiple words it's basically a substring search. The results are sorted by where the match happened (e.g. first word, second word etc).

diversemix:
@hdrury1 Something tells me we should actually document (somewhere) how this search functionality behaves - otherwise we might be in for repeating ourselves :)

PR #476, opened 2018-08-14
feat(PeoplePicker): improve people picker filtering functionality

dryhten:
re #443

### DIFF
--- a/app/components/ui/molecules/PeoplePicker.js
+++ b/app/components/ui/molecules/PeoplePicker.js
@@ -5,6 +5,7 @@ import styled from 'styled-components'
 import { Button } from '@pubsweet/ui'
 import { th } from '@pubsweet/ui-toolkit'
 import { Box, Flex } from 'grid-styled'
+import { cloneDeep } from 'lodash'
 
 import SelectedItem from '../atoms/SelectedItem'
 import PersonPod from '../atoms/PersonPod'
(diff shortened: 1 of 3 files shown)


app/components/ui/molecules/PeoplePicker.js
```diff
@@ -145,11 +146,71 @@ class PeoplePicker extends React.Component {
...
+  filterPeopleSingle = (people, inputValue, field) => {
+    let maxWords = 0
+    people.forEach(person => {
+      const numWords = person[field].match(/[^ ]+/g).length
+      if (numWords > maxWords) {
+        maxWords = numWords
+      }
+    })
+
+    let j
```
diversemix: Where do we initialise `j` ?

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -145,11 +146,71 @@ class PeoplePicker extends React.Component {
...
+    const addMatchIndex = person => {
+      const clonedPerson = cloneDeep(person)
+      clonedPerson.matchIndex = matchIndex[j]
+      j += 1
+      return clonedPerson
+    }
+
+    for (let i = 0; i < maxWords; i += 1) {
+      /* matches for ith word */
+      j = 0
```
diversemix: Maybe you could explain how `j` is used? Not sure I understand, not sure its entirely clear... could we encapsulate this in some way?

diversemix:
Looking good ... but we need to sort that magical `j` :)

app/components/ui/molecules/PeoplePicker.js
dryhten: This `j` made more sense when I had the filtering/mapping functions in place but because of some weird eslint rules (which I didn't want to disable) now it looks all weird. Yeah it guess it's really hard to read like this.

I think a better question here is this: we have to see if the person matches the input value once, when we filter the results, and then again when we render the results because I need to know the index of the match so I can turn the matched part bold. Right now I perform this match test only once (here) and store the match index in the people list so I can use it again in the render function. Does it make more sense to test for the match twice (here and in the render function)? I think so, I can probably put this part in a function and use it in both places. We don't save that much by performing the match test twice since it's just a small list of items. It will be easier to read.

diversemix: I'll delete my other comment as it amounts to the same thing!

diversemix: @dryhten  ... Happy to do the merge if you are planning to do the re-factor? Know we've been busy today - just putting this here as a reminder ;)

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -36,7 +36,8 @@ const SelectionHint = ({ selection, minSelection, maxSelection }) => {
     const numRequired = minSelection - selectionLength
     return (
       <div>
-        {numRequired} more suggestion{numRequired === 1 ? '' : 's'} required
```
dryhten: What happened here? No prettier/eslint on pre-commit before?

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -145,11 +147,57 @@ class PeoplePicker extends React.Component {
...
+        return clonedElem
+      })
+    }
+    return this.filterPeopleSingle(people, inputValue, field)
+  }
+
+  filterPeopleSingle = (people, inputValue, field) => {
+    let maxWords = 0
+    people.forEach(person => {
+      const numWords = person[field].match(/[^ ]+/g).length
```
tamlyn: What does this regex match?

app/components/ui/molecules/PeoplePicker.js
```diff
@@ -145,11 +147,57 @@ class PeoplePicker extends React.Component {
...
+      }
+      return re.test(words[n].toLowerCase())
+    }
+
+    for (let i = 0; i < maxWords; i += 1) {
+      /* matches for ith word, starting at 0 */
+      const ithMatches = people.filter(person => filterNthMatches(person, i))
+      matches = matches.concat(ithMatches)
+    }
+    return matches
```
tamlyn: I don't really understand what any of this is doing.

app/components/ui/molecules/PeoplePicker.js
dryhten: It matches all the words. I haven't done enough testing to be fair, if you know an edge case where this is not working feel free to say :)

app/components/ui/molecules/PeoplePicker.js
dryhten: We discussed this irl, I match after the first word, then after the second and so on.

Issue #255, opened 2018-06-27
Discussion: What should our testing strategy be for the front-end?


hdrury1:
This was a point raised in the course of reviewing https://github.com/elifesciences/elife-xpub/pull/231

> Are you actually commenting on the choice to write integration tests & not really do unit tests? I've got to say, the lack of unit tests feels a little wrong to me too. But I asked about this when I first started working on testing for this component. It seemed to be a commonly held opinion that integration tests would add more value than unit tests & that unit tests may not even be worth writing.

Can we agree on a common approach to front-end testing? This should result in an ADR.

hdrury1:
Closing as we have an aligned testing strategy now.

PR #496, opened 2018-08-17
add Disclosure page as 5th step to submission wizard

### LINKED ISSUE #489: Create data disclosure page
- There should be an additional page added to the wizard to allow the submitter to confirm they are happy with our data disclosure agreement

- The page should pull in the article title, current date and corresponding author's name captured in Steps 1 and 2 of wizard.

- 'Privacy policy' should link to https://elifesciences.org/privacy-notice

- The user should not be able to proceed past this page without ticking the box and entering a value in the name input

- Validation message for the tick box should be 'We are unable to proceed without your consent.'

- An extra step should be added to the progress indicator to account for the extra page.

**Designs**

wizard-screen-5 2x

screen_shot_2018-08-13_at_16 36 39

jsms90:
#### What does this PR do?
Add 5th page to submission wizard - consenting to the disclosure of data

Visuals:
- remove submission action from step 4 of the wizard (i.e. change "Submit" button to "Next" & change which resolver is called)
- add 5th step to `ProgressBar` component
- pull author's first and last name from the values saved by completing step 1 of the wizard
- pull manuscript title from step 3 of the wizard

Behind the scenes:
- add new route: `submit/disclosure`
- add "submitterSignature" & "disclosureConsent" to frontend and backend validation schemas & graphql schema and type definitions

#### Any relevant tickets
fixes #489  

### How has this been tested?
Visually

### Still to be done
- Show the date next to the article type


app/components/pages/SubmissionWizard/steps/Disclosure/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+const RegularP = styled.p`
+  font-size: ${th('fontSizeBase')};
+  line-height: ${th('fontLineHeightBase')};
+`
+
+const SmallP = styled.p`
+  font-size: ${th('fontSizeBaseSmall')};
+  line-height: ${th('fontLineHeightBaseSmall')};
+  color: ${th('colorTextSecondary')};
```
damnhipster: These look like good candidates to be atoms. Let's make sure that we use the atoms anywhere we would be using `RegularP`, `SmallP` and `StyledLink`

app/components/pages/SubmissionWizard/steps/Disclosure/schema.js
```diff
@@ -0,0 +1,11 @@
...
+
+const schema = yup.object().shape({
+  submitterSignature: yup.string().required('Your name is required'),
+  disclosureConsent: yup
+    .bool()
+    .required()
+    .oneOf([true], 'We are unable to proceed without your consent'),
+})
+
+export { schema }
```
damnhipster: It's worth writing a test for this, as this is what is used to decide whether the user is allowed to Submit or not.

app/components/pages/SubmissionWizard/steps/Disclosure/index.js
jsms90: Yeah completely agreed. Especially since we have these elsewhere in the codebase as well (within [PersonPod](https://github.com/elifesciences/elife-xpub/blob/66c00a84bf03c611f873c0cd1e239200aa5cb48f/app/components/ui/atoms/PersonPod.js#L75))

I've raised an issue in pubsweet, as the paragraph components should probably be added in there - https://gitlab.coko.foundation/pubsweet/pubsweet/issues/423

I'm unsure about `StyledLink`, but I should probably be using pubsweet's [`Link` component](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/ui/src/atoms/Link.js) & overriding styling to remove text-decoration.

@hdrury1 Am I right in saying that this is going to be a common pattern across the `elife-xpub` app? That all links within text will be styled like this?
```
color: ${th('colorPrimary')};
text-decoration: none;
```

app/components/pages/SubmissionWizard/steps/Disclosure/schema.js
jsms90: I know that I used to bring this up during standups a fair bit, but at the risk of becoming annoying about this, perhaps now is a good time for us to revisit the creation of a testing strategy? #255 :pray: 

Testing the schema would be a good way of catching validation bugs :+1: More than happy to do this.

But there are plenty of other ways our app could break too, which we're not covering with tests e.g. progressing through the submission wizard correctly. And the layout bug wasn't the first time that breaking changes have made their way into `develop`: https://github.com/elifesciences/elife-xpub/issues?q=is%3Aissue+label%3ABug+is%3Aclosed, which we should probably start being more strict with, since we're coming up to first release?

Then we could start making certain kinds of tests a requirement for each PR, so that we don't keep saying:
> #### How has this been tested?
>
> Visually / manually

app/components/pages/SubmissionWizard/steps/Disclosure/index.js
hdrury1: Correct @jsms90

app/components/pages/SubmissionWizard/steps/Disclosure/schema.js
damnhipster: This is a discussion we should have with everyone else. I suggest creating an issue for this, including the discussion.

jsms90: Yeah I'll comment more thoroughly in #255 a bit later today

jsms90: I've pushed a `schema.test.js` file. Lemme know what you think & if there are other tests you think should be in this PR? :smile:

Issue #473, opened 2018-08-14
Use individual tables instead of "entities"


tamlyn:
**User Story:**

**Product requirements/Acceptance criteria:**

Non-functional:
- Move all queries over to use new SQL schema with tables per entitiy

Depends on https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/368

diversemix:
This is now in the latest version of pubsweet ... we now have this in `develop`

Issue #493, opened 2018-08-17
Fix resolvers 'extracts title from PDF' test


dryhten:
Right now this test is skipped. It was supposed to be done as part of https://github.com/elifesciences/elife-xpub/pull/368 but it's not something fast to be fixed and the PR had to be merged for various reasons. Before `uploadManuscript` is called a `Subscribe.uploadProgress` subscription should be done.

diversemix:
The upload test is being removed as part of #497  ... this will need to be re-added.

tamlyn:
Done some time ago.

PR #497, opened 2018-08-18
feat: store data in individual database tables

tamlyn:
#### Background

- Store manuscripts, teams and files in their own tables (keep users in `entities`)
- Use `knex` for query building
- Change some columns to arrays instead of JSONB due to a quirk in `pg`'s query formatting.
- Use `,` instead of `.` in column names (e.g. `meta,title`) due to dots being special in knex

#### Any relevant tickets

Partially addresses #473 (users and identities still to do)

Existing tests updated, new unit tests written.


diversemix:
Well after a long day of trying to get the tests to work... there seems to be a few issues that need to be sorted. 

- The queries do not work if the tables they are joining on are empty
- There are "random" errors that a given table does not exist.
- There are "random" errors that a given relations do not exist.
- deadlocks are detected 

Ideas:

- In the middle of trying to create a framework for `data-access.js`
- Improve errors lots of them have the stack: ```
      at Connection.Object.<anonymous>.Connection.parseE (node_modules/pg/lib/connection.js:553:11)
      at Connection.Object.<anonymous>.Connection.parseMessage (node_modules/pg/lib/connection.js:378:19)
      at Socket.<anonymous> (node_modules/pg/lib/connection.js:119:22)
```

**Comment by diversemix** [comment]
I think I've found the problem (at 11:33pm!) ... needs a fix in `create-tables.js` like:
```
      let deleteSQL = ""
      await Promise.all(
        // TODO this is dangerous, change it
        rows.map(row => deleteSQL += `DROP TABLE "${row.tablename}" CASCADE;`),
      )
      await db.query(deleteSQL)
```
or something a bit more elegant!

File: server/xpub/entities/manuscript/resolvers.test.js [hunk 1]
```diff
@@ -275,29 +283,29 @@ describe('Submission', () => {
...
-      expect(manuscript).toMatchObject({
-        id,
-        meta: {
-          title:
-            'The Relationship Between Lamport Clocks and Interrupts Using Obi',
-        },
-        files: [{ filename: 'manuscript.pdf' }],
-      })
-    })
+    // it.skip('extracts title from PDF', async () => {
```
damnhipster: Warnings don't fail the build, so you don't need to comment this out

server/xpub/entities/manuscript/resolvers.test.js
```diff
@@ -28,10 +32,12 @@ describe('Submission', () => {
     const user = await User.save(userData)
     userId = user.id
     mailer.clearMails()
+    initializing = false
```
damnhipster: We shouldn't need to have an `initializing` parameter. Jest will wait for the async function to complete before running the tests.

damnhipster: You can see details [here](https://jest-bot.github.io/jest/docs/setup-teardown.html) and [here](https://github.com/facebook/jest/issues/1256)

diversemix: You're right we shouldn't. However I'd like to leave them in - that way if anyone introduces code that makes the tests fail because of async init or tear down then we will easily know.

server/xpub/entities/manuscript/resolvers.test.js
diversemix: But it does stop you making a commit ;P

damnhipster: Ah I didn't realise, I may not be running the commit hook locally... I suggest deleting the test and addressing it in #493

diversemix: Good plan!

server/xpub/entities/manuscript/resolvers.test.js
damnhipster: It may be worth using a promise to make it clearer that we're waiting for the database to be initialised. Using an assertion says to me that we're testing the database initialisation.

i.e.
```
 let initializeDb;

 const initialize = async () => {
    replaySetup('success')
    await Promise.all([
      fs.remove(config.get('pubsweet-server.uploads')),
       createTables(true),
     ])
    const user = await User.save(userData)
    userId = user.id
    mailer.clearMails()
 }

 beforeEach(async () => {
    initializeDb = initialize();
 });

 ...

 it('Returns empty object when there are no manuscripts in the db', async () => {
      await initializeDb();
      const manuscript = await Query.currentSubmission({}, {}, { user: userId })
      expect(manuscript).toBe(null)
 })
```

diversemix: Ooo yeah I like that :+1: 
.. you want to add it? @damnhipster

tamlyn: This `initializing` stuff is unnecessary noise. All it is asserting is that the `beforeEach` function completed before the test started but this is guaranteed by Jest already. 

If a bug in the initialisation logic is introduced (e.g. forgetting to await a statement) this won't catch it because the `beforeEach` still completed and set `initializing` back to `false`. The most likely way that assertion could fail is if an exception is thrown in the `beforeEach`. But in that case it's confusing because the initialisation failed so `initializing` should really be `false`.

The promise solution is bypassing Jest's automatic waiting in favour of explicit manual waiting. Surely that's likely to lead to more bugs, not fewer.

server/jest-setup.js
```diff
@@ -1,27 +1,10 @@
...
-  const pool = new pg.Pool({ database: mockTestDbName })
-  pool.on('error', err => {
-    if (err.message !== 'terminating connection due to administrator command') {
-      logger.error(err)
-    }
-  })
-  pool.testDbName = mockTestDbName
-
-  return pool
-})
```
tamlyn: Removing this stuff means we can no longer parallelise test runs (need to run jest with `--runInBand` to avoid errors on multi-core machines). But we have so few tests that for the time being it's only a couple of seconds faster when running in parallel. The gain in simplicity is probably worth the loss in performance.

diversemix: It wasn't the simplicity ... it was the fact that `jest.mock` doesn't work. There are three places `db.js` is used in the project and the jest mock would only work for one of them. I did try to explicitly mock out the paths but to no avail. Do you know how to ensure that jest mocks everything under `node_modules` ? @tamlyn

server/xpub/entities/manuscript/resolvers.test.js
diversemix: I'd like to leave it in there... it helped me find the issue in pubsweet with the asynchronous drop tables. So it would catch issues with pubsweet? Please correct if I am wrong/you disagree.

server/jest-setup.js
tamlyn: No, I just had a look too and gave up. It's a hacky solution which is good when it works but hard to debug when it doesn't. That's what I means about simplicity. At least it's easier to reason about what's going on when you have a single database.

diversemix: Agreed,  just a shame it doesn't work. I just saw your commit (e615fc1) to put in the "runInBand" flag, looks good. I was wondering where is the best place to put in the tests for `selectById` (and status) to test the queries.... At the `data-access` level it gets tested to make sure it returns the correct object id ... but it doesn't check exactly what is returned. I'm thinking this should go in the `resolve.test` ...? ... what's the thinking @damnhipster  @tamlyn ?

server/xpub/util.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+const entityToRow = (manuscript, columns) =>
+  columns.reduce((entity, columnName) => {
+    const camelKey = keyToCamelCase(columnName)
+    const value = lodash.get(manuscript, camelKey)
+    return { ...entity, [columnName]: value }
+  }, {})
+
+module.exports = { rowToEntity, entityToRow }
```
damnhipster: These functions look like they're better off living under `server/xpub/entities/util.js`, as they're specifically for entities.

server/jest-setup.js
tamlyn: With Hem's help I realised that the root of the problem here was that `pubsweet-server` was present multiple times in the dependency tree. This is a problem because each one gets its own database pool and it introduces subtle problems as seen here. Ensuring that all pubsweet dependencies are always upgraded together avoids this.

I've re-enabled parallel test execution for now. It will become more valuable as the number of tests increases.

diversemix: Oh wow! That is fantastic! So how was the dependency tree fixed ... we'll talk tomorrow - great work again :+1:

server/xpub/entities/file/data-access.js
```diff
@@ -0,0 +1,59 @@
...
+  'filename',
+  'label',
+  'type',
+  'mime_type',
+  'size',
+  'url',
+]
+
+const dataAccess = {
+  async selectById(id) {
```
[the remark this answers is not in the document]
diversemix: ... the issue with that is we need to create `files` and `teams` within that script. But I guess this is a reflection on the hierarchy of objects. Maybe we directly use the corresponding entity's `data-access` functions to do that?

diversemix:
I'm trying to run the tests on this branch ... and I started with a fresh database and got:
``` 
error: relation "migrations" does not exist
```

along with #539  

Is anyone else seeing this?

server/xpub/entities/file/index.js
```diff
@@ -0,0 +1,21 @@
+const lodash = require('lodash')
+const dataAccess = require('./data-access')
+
+const empty = { teamMembers: [] }
```
diversemix: This is a bug - fixing as part of the testing changes going it.

diversemix:
@tamlyn I've updated with the tests and made that minor fix to the definition of `empty` for `File`

server/xpub/entities/file/data-access.test.js
```diff
@@ -0,0 +1,76 @@
...
+  manuscriptId = await manuscriptTestData.createSingleManuscript()
+  testId = await createAFile(manuscriptId, 'testFile')
+  console.log('Created testId', testId)
+
+  return testId
+}
+
+describe('FileAccessLayer', () => {
+  beforeEach(async () => {
+    testId = await initializeDatabase()
```
tamlyn: You're already assigning `testId` to the outer variable in `initializeDatabase`.

Issue #507, opened 2018-08-20
Demo environment fails (websockets)


damnhipster:
This is due to WebSockets being requested over http when the page is https.

```
client.js:426 Mixed Content: The page at 'https://demo--xpub.elifesciences.org/' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://demo--xpub.elifesciences.org/subscriptions'. This request has been blocked; this endpoint must be available over WSS.
SubscriptionClient.connect @ client.js:426
client.js:426 
```

diversemix:
Potentially, we need to change the salt-stack used to deploy to the demo site. This is stored here:

https://github.com/elifesciences/elife-xpub-formula/blob/master/salt/elife-xpub/config/etc-nginx-sites-enabled-xpub.conf

This could contain the port forwarding as illustrated: http://nginx.org/en/docs/http/websocket.html

diversemix:
The above suggested update to the config of nginx has been attempted - but does not work. @damnhipster will also try on the YLD infra, I will also try updating on the live `demo` site to see what the issue is. 

This has highlighted another issue - one of the container size being so large that we cannot do a test deployment locally in a VM - as the network keeps dropping. This also might be due to the office environment too. There is a fix for this in `245-prod` to dramatically reduce the size of the container. Going to raise a separate issue for this.

diversemix:
New issue #517

damnhipster:
Websocket requests can be made using wss:// rather than ws:// on a secure page. The demo and staging environments on Kubernetes support this out of the box. What still needs to be done is changing the client to use wss:// rather than ws:// if the page is requested over https. This logic sits within pubsweet: https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/370

damnhipster:
I've raised #518 to roll back the changes, as it'll require a change in pubsweet-client to support `wss://`.

diversemix:
Reverting is a good idea in this instance. I'll mark this as blocked ... while you get an MR for those changes (https://gitlab.coko.foundation/pubsweet/pubsweet/commit/16ab81c940a5cce58d1988499c63c0541786e47e ) ... and we get a new release :+1:

damnhipster:
Waiting on https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/380 to be released. This should allow us to bring back the changes introduced in #368

diversemix:
[...]

PR #518, opened 2018-08-22
Revert "Merge pull request #368 from elifesciences/312-push-messages"

damnhipster:
#### Background

This reverts the Websocket subscription, as it doesn't work when the app is served over https. This is due to the subscription url being prefixed with `ws://` in [pubsweet-client](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/client/src/components/Root.js#L37)

#### Any relevant tickets
#368, #507

CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "joi-browser": "^13.4.0",
     "lint-staged": "^7.2.2",
     "prettier": "^1.8.2",
-    "pubsweet": "^2.2.8",
+    "pubsweet": "^2.3.0",
     "react-styleguidist": "^7.1.0",
     "replay": "^2.1.4",
     "string-replace-loader": "^2.1.1",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -139,7 +139,7 @@
     "joi-browser": "^13.4.0",
     "lint-staged": "^7.2.2",
     "prettier": "^1.8.2",
-    "pubsweet": "^2.3.0",
+    "pubsweet": "^2.2.8",
```
diversemix: Don't think we should be downgrading these - probably what is causing the failures in the pipeline.

server/dar-server/package.json
```diff
@@ -12,6 +12,6 @@
     "express": "^4.16.2"
   },
   "peerDependencies": {
-    "pubsweet-server": "^8.0.0"
+    "pubsweet-server": "^7.0.0"
```
diversemix: Don't think we should be downgrading these - probably what is causing the failures in the pipeline.

Issue #509, opened 2018-08-21
Update input for excluding reviewers


hdrury1:
**Product recs**

- Clicking on 'Would you like to exclude a reviewer' should generate a hide/show panel that contains inputs for all three potential exclusions at once

- There should be one reason to account for all exclusions, which fits the data model

- Clicking the cross should hide the panel

- If the panel has been opened, there should be input in both reviewer 1 name and reviewer 1 email text boxes (validation message: 'Name required' and 'Email required')

- There should be no validation against the optional fields UNLESS one has input in, i.e. if user has entered the name of optional reviewer 2 they must also include an email and vice versa

- ~To ensure consistency throughout the application, please update labels of optional fields to "XXX (optional)" (e.g. "Excluded reviewer 1 name (optional)")~ - separate issue has been raised for this #508 

**Designs**

![wizard-mobile-reviewers 2x](https://user-images.githubusercontent.com/17023702/44392662-a0d01100-a52a-11e8-99d6-696a12e8e64a.png)

Issue #508, opened 2018-08-21
Update reviewer suggestion inputs


hdrury1:
- Slight tweaks to the labelling of the text inputs should ensure they are more obviously linked in the mobile view of the application

- Labelling for optional 4th, 5th and 6th reviewers should be:
Reviewer 4 name (optional), Reviewer 4 email (optional)

**Designs**
<img width="782" alt="screen shot 2018-08-21 at 10 02 34" src="https://user-images.githubusercontent.com/17023702/44392263-b4c74300-a529-11e8-8d65-7d0d26af11e7.png">

Issue #459, opened 2018-08-09
Ensure consistency in text throughout application


hdrury1:
**User Story:**

_As a marketer I want xPub content to follow eLife house style in order to reflect a common brand._

**Our Press Officer has proofread the demo and has made the following requests for minor textual adjustments:**

**In the 'Files' section**

- When you upload your manuscript, you get a success message asking if you want to 'Preview or replace your Manuscript'. 'Manuscript' doesn't need to be capitalised in this context. (I also noticed that it's capitalised in the error message that comes up if there's a problem, which again it doesn't need to be.)

**In the 'Submission' section**

- When you're asked to specify the article type, it currently gives you the option of 'Research article' or 'Feature article'. 'Article' should be capitalised in this context, as these are proper nouns we use to describe these types of papers.

- When you tick the box to say that 'This manuscript is part of a co-submission', a question comes up asking if you would like to 'include another cosubmission'. This is currently missing a question mark at the end, and 'cosubmission' should be hyphenated to keep it consistent with the tick-box text.

**In the 'Editors' section**

- When you're asked who should review your work, the text in the box for both Senior Editors and Reviewing Editors says 'Choose Editor'. 'Editor' in this context should be lower case, as we don't capitalise the word when referring to editors generally.

- When you're asked if you'd like to 'exclude a senior editor' or 'exclude a reviewing editor', both 'senior editor' and 'reviewing editor' should be capitalised.

- And in the heading that says "Suggest Reviewers", "reviewers" shouldn't be capitalised.

PR #524, opened 2018-08-23
WIP: Textual changes to match new wireframes and house style

hdrury1:
#### Background

Closes down #508 and #459. Also lays foundation for closing down #509.

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
+++ b/app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
@@ -2,8 +2,7 @@ import { cloneDeep } from 'lodash'
 import React from 'react'
 import { Box } from 'grid-styled'
 import { Action, ErrorText } from '@pubsweet/ui'
-import MoreButton from '../../../../ui/molecules/MoreButton'
-import { FormH3 } from '../../../../ui/atoms/FormHeadings'
+import { FormH3, FormH4 } from '../../../../ui/atoms/FormHeadings'
 
 import {
   Declaration,
@@ -37,6 +36,23 @@ const OptionalExclude = ({
       ?
     </Box>
   )
+/*
+ * If using the OptionalExclude component you need to extend the function
+ * below, to define when its visible.
+ */
+const OptionalExcludeVisibilites = (nextProps, prevState) => ({
+  boxVisibility: {
+    opposedSeniorEditors:
+      !!nextProps.values.opposedSeniorEditors.length ||
+      prevState.boxVisibility.opposedSeniorEditors,
+    opposedReviewingEditors:
+      !!nextProps.values.opposedReviewingEditors.length ||
+      prevState.boxVisibility.opposedReviewingEditors,
+    opposedReviewers:
+      !!nextProps.values.opposedReviewers.length ||
+      prevState.boxVisibility.opposedReviewers,
+  },
+})
 
 const ValidationMessage = ({ message }) => (
   <div aria-live="polite">{message && <ErrorText>{message}</ErrorText>}</div>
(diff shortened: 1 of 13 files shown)


app/components/pages/SubmissionWizard/steps/Editors/FormSections.js
```diff
@@ -20,17 +19,17 @@ export const SuggestedReviewer = ({ changeHandler, index }) => (
     <ValidatedField
       label={
         index < limits.suggestedReviewers.min
-          ? 'Suggested reviewer name'
-          : 'Optional suggested reviewer name'
+          ? `Suggested reviewer ${index + 1} name`
+          : `Suggested reviewer ${index + 1} name (optional)`
```
jsms90: @diversemix was "suggested" left in here on purpose? The designs just say "Reviewer 1 name", "Reviewer 2 email", etc:
screen shot 2018-08-21 at 10 02 34

diversemix: Good catch ... I'd leave it for now and @hdrury1 can make a call once she is back.

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
```diff
@@ -37,6 +36,23 @@ const OptionalExclude = ({
       ?
     </Box>
   )
+/*
+ * If using the OptionalExclude component you need to extend the function
+ * below, to define when its visible.
+ */
+const OptionalExcludeVisibilites = (nextProps, prevState) => ({
```
jsms90: I'm not sure I understand why this has been abstracted out from `getDerivedStateFromProps`?

diversemix:
I agree I don't think this is ready. I'll mark it as WIP, its something I don't mind picking up next week if you have other stuff to do @jsm90

jsms90:
@diversemix It's ok, I'm happy to work on this as I've just been looking through the PR so my head is very much on this body of work.

Going to raise a PR that focuses on the copy changes alone. I would rebase this branch to keep @hdrury1's commits but the copy changes have been made on top of the refactor. So I'm just going to add it to a new branch so we can get those 2 issues closed off today :+1: 

Then I'll have a stab at #509 while this is fresh in my mind.

app/components/pages/SubmissionWizard/steps/Editors/EditorsContent.js
diversemix: This has been done to make the code more maintainable

jsms90: Yeah I see that the comment above says this makes it extendable. I'm not necessarily saying I disagree. I just genuinely don't understand.

What scenario does this cater for, that couldn't be done by adding directly into `getDerivedStateFromProps`?

diversemix: Let's talk about it IRL ... it basically explicitly ties the functionality of `getDerivedStateFromProps` to the `OptionalExclude` component by name. I'd like this done my making a class and using a design pattern for state management - as it terns out React makes this complicated. Would like to talk to Tam about it - incase he knows any React patterns that would make this pretty.

jsms90: Yeah we can definitely talk about it IRL if you prefer. I was just wondering about the reasoning, so that I could tell whether to pop this into the implementation I'm working on?

From what I can tell, your intention for this is about making state manipulation class-based? But this feels a little like fighting against React itself, since the premise of the library is that UI is conceived of as a function of state. So by design, the manipulation of state is intended to remain functional.

Happy to talk about changes like this with @tamlyn when he comes back, but for now I would suggest that we stay with a more functional approach?

jsms90:
This work has been separated out:
- Issues 508 & 459 have been closed off by PR #528
- Issue 509 has been completed (pending review) via PR #531 

Closing this PR

Issue #481, opened 2018-08-15
Fix Gitlab mirroring to speed up/enable the build pipeline


damnhipster:
We've had several occasions where none of the builds have been running for hours, which has slowed down code reviews and demos.

The issue is with Gitlab's code mirroring feature, which we use to copy the code from Github to Gitlab, so we can use Gitlab CI.

The issue has affected lots of users of Gitlab CI:
https://gitlab.com/gitlab-org/gitlab-ee/issues/7243
https://gitlab.com/gitlab-com/infrastructure/issues/4764

Gitlab hasn't identified the cause as of 15/08/2018. In the meantime, we should write our own mirroring script to be scheduled every minute.

- [x] create script to clone from Github, and push to Gitlab, every minute
- [x] understand whether script can be run alongside Gitlab's mirroring feature
- [x] run script using Docker
- [x] push Docker container to the Kubernetes cluster
- [x] add Gitlab key to the Kubernetes deployment for push access
- [x] create docker repository for the mirroring scripts - _created https://hub.docker.com/r/xpub/git-mirror/_
- [x] host docker image on https://hub.docker.com/r/xpub/git-mirror/
- [x] create Kubernetes templates and add them to the codebase - moved to #481

damnhipster:
Closing in favour of #530 for the remainder of the work.

PR #529, opened 2018-08-24
Added scripts to mirror from Github to Gitlab

damnhipster:
#### Background

Addresses #481


scripts/gitlab-mirror/mirror
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+TARGET=$1
+MIRROR=$2
+DELAY=${3:-60}
+SSH_KEY=${4:-/root/.ssh/gitlab-mirror}
```
diversemix: A neater way (IMHO) is to use  `~/.ssh/config`  to specify keys on a per host basis. Although slightly less flexible it does separate the transport layer from the script... just my preference ;)

scripts/gitlab-mirror/mirror
```diff
@@ -0,0 +1,21 @@
...
+ssh-add $SSH_KEY
+
+while true
+do
+mkdir -p repo/
+git clone --mirror $TARGET repo/
+cd repo
+git push --mirror $MIRROR
+cd ..
+rm -rf repo
```
diversemix: Perhaps we can use `git remote update` rather than removing the folder each time? (as per: https://stackoverflow.com/questions/6150188/how-to-update-a-git-clone-mirror )

damnhipster: Good find, will try it out

scripts/gitlab-mirror/mirror
damnhipster: `~/.ssh/config` is necessary for this to work, and is currently meant to be included as part of the volume mount. I'll tidy this part of the code up so to make it clearer, and maybe avoid using a mount if I can.

They key was only there because it wasn't being added to the ssh-agent on startup for some reason. I'll do some googling as to the correct method of adding ssh keys to containers.

scripts/gitlab-mirror/commands.sh
```diff
@@ -0,0 +1,25 @@
+function build() {
+  docker build -t xpub/git-mirror:latest .
+}
+
+function docker-run() {
+  ssh_dir=$HOME/Projects/elife/.ssh
+  docker run \
+  --name mirror \
+  -v $ssh_dir:/root/.ssh \
+  -it xpub/git-mirror:latest \
```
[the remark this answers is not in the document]
damnhipster: These commands are meant to be run locally and not to be automated. `-it` allows the user to see the output when running locally. I'll document them in the README instead and I can rename the file to `examples.sh`.

scripts/gitlab-mirror/mirror
```diff
@@ -22,6 +22,6 @@ git push --mirror $MIRROR
 # Update the mirror
 while true
 do
-git remote update
+git remote update --prune
```
diversemix: Oooo

damnhipster: Should be safe :) If we've deleted a branch on Github, it shouldn't be a problem if the pipeline on Gitlab gets deleted too

PR #569, opened 2018-09-07
WIP: Mock out css as per docs

diversemix:
I'm not entirely convinced this is going to work. It certainly appeared to fix the problem, but removing the solution did not show the problem again. So I have my doubts.

Ref: https://jestjs.io/docs/en/webpack#handling-static-assets

### DIFF
--- a/__mocks__/styleMock.js
+++ b/__mocks__/styleMock.js
@@ -1,0 +1,3 @@
+// __mocks__/styleMock.js
+
+module.exports = {}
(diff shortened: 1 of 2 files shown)


damnhipster:
I'm able to reproduce the problem consistently, and this change wasn't able to fix it. Jest still attempts to parse the imported file even when it's mocked out.

diversemix:
OK .. happy to  delete the branch and this PR?

diversemix:
Closing as @damnhipster  has confirmed it does not work

PR #570, opened 2018-09-07
fix(SearchBox): remove matchIndex from multi words match

dryhten:
this was left here by accident


dryhten:
This is here from when I was precomputing match index for suggestions. There was a discussion in this PR https://github.com/elifesciences/elife-xpub/pull/476 about it.

The outcome was to create a new function (`getMatchIndex`) and use that every time we render suggestions in the drop down list, instead of precomputing them.

Issue #438, opened 2018-08-06
Fix search icon for SearchBox


dryhten:
The search icon in the SearchBox has the following problems:
1. It should be clickable, having the same effect as hitting enter. Right now it doesn't do anything.
2. Typed text shouldn't go underneath it. Right now it does.

hdrury1:
Note that #540 affects these requirements slightly.

jsms90:
Just going to update the top comment based on slightly altered design

PR #573, opened 2018-09-07
WIP Search button for people picker

jsms90:
#### What does this PR do?
- Create new SearchButton component
- Replace search icon with this button
- Add clicking functionality, which filters the pods

#### Any relevant tickets
#438 

Will add to test file once it's created, but writing existing tests is In Progress right now as well

### Why this is still WIP
Currently battling with positioning

At the moment I've got the search button in line with the input field, but the dropdown menu is too wide:
image

Or the dropdown menu is too narrow:
image


jsms90:
Ah! I removed `position: relative` from the `Root` element when I took out the IconContainer that had `position: absolute`. I didn't twig that the dropdown list also had absolute positioning on it.

So I've added `position: relative` back in to fix the issue I mentioned above. But also rebasing now, so this isn't gonna be obvious from the commit history. Just mentioning in case anyone was curious

Issue #470, opened 2018-08-13
Write tests for Search Box


dryhten:
We need tests for the [Search Box](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/ui/molecules/SearchBox.js) component.

PR #575, opened 2018-09-10
test(SearchBox): add tests for SearchBox component

dryhten:
re #470

### DIFF
--- a/app/components/ui/molecules/SearchBox.test.js
+++ b/app/components/ui/molecules/SearchBox.test.js
@@ -12,30 +12,30 @@ const searchOptions = [
   { value: 'second option' },
 ]
 
-function filterFunction(options, searchValue, field) {
-  if (!searchValue) return options
+const mockFilter = jest.fn((options, searchValue, field) => {
+  if (!searchValue) return []
 
   const inputValue = searchValue.trim().toLowerCase()
-  if (!inputValue) return options
+  if (!inputValue) return []
 
   return options.filter(option =>
     option[field].toLowerCase().includes(inputValue),
   )
-}
+})
 
-function getMatchIndex(inputValue, option) {
+const mockMatchIndex = jest.fn((inputValue, option) => {
   const re = new RegExp(escapeRegExp(inputValue))
   const match = re.exec(option.toLowerCase())
   if (match) return match.index
   return -1
-}
+})
 
 function makeWrapper() {
   return mount(
     <ThemeProvider theme={theme}>
       <SearchBox
-        filterFunction={filterFunction}
-        getMatchIndex={getMatchIndex}
+        filterFunction={mockFilter}
+        getMatchIndex={mockMatchIndex}
         onSubmit={jest.fn()}
         options={searchOptions}
       />
@@ -51,35 +51,30 @@ describe('SearchBox component tests', () => {
   })
 
   it('has search icon', () => {
-    expect(wrapper.find('#search-icon')).toHaveLength(1)
+    expect(wrapper.find('[data-test-id="search-icon"]')).toHaveLength(1)
   })
 
-  it("doesn't show any suggestions on empty input", () => {
-    wrapper.find('input').simulate('change', { target: { value: '' } })
-    expect(wrapper.find(SearchBox).instance().state.suggestions).toEqual([])
-  })
-
-  function compareSuggestions(searchValue) {
-    const expectedSuggestions = filterFunction(
-      searchOptions,
-      searchValue,
-      'value',
-    )
-    wrapper.find('input').simulate('change', { target: { value: searchValue } })
-    const receivedSuggestions = wrapper.find(SearchBox).instance().state
-      .suggestions
-    expect(receivedSuggestions).toEqual(expectedSuggestions)
-  }
-
-  it('typing something will show suggestions based on filtering function, test 1', () => {
-    compareSuggestions('option')
-  })
-
-  it('typing something will show suggestions based on filtering function, test 2', () => {
-    compareSuggestions('f')
-  })
-
-  it('typing something will show suggestions based on filtering function, test 3', () => {
-    compareSuggestions('xyz')
-  })
+  it.each([[1, 'option'], [2, 'f'], [3, 'xyz'], [4, '']])(
+    'typing something will show suggestions based on filtering function, test %i',
+    (index, searchValue) => {
+      /**
+       * these tests make the assumption that suggestions will
+       * be correctly rendered from the state of the component
+       *
+       * in the future this should be changed to check for the
+       * list of rendered suggestions directly
+       */
+      const expectedSuggestions = mockFilter(
+        searchOptions,
+        searchValue,
+        'value',
+      )
+      wrapper
+        .find('input')
+        .simulate('change', { target: { value: searchValue } })
+      const receivedSuggestions = wrapper.find(SearchBox).instance().state
+        .suggestions
+      expect(receivedSuggestions).toEqual(expectedSuggestions)
+    },
+  )
 })
--- a/app/components/ui/molecules/SearchBox.test.js
+++ b/app/components/ui/molecules/SearchBox.test.js
@@ -1,0 +1,85 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import { ThemeProvider } from 'styled-components'
+import { escapeRegExp } from 'lodash'
+import theme from '@elifesciences/elife-theme'
+
+import SearchBox from './SearchBox'
+
+const searchOptions = [
+  { value: 'first option' },
+  { value: 'final option' },
+  { value: 'second option' },
+]
+
+function filterFunction(options, searchValue, field) {
+  if (!searchValue) return options
+
+  const inputValue = searchValue.trim().toLowerCase()
+  if (!inputValue) return options
+
+  return options.filter(option =>
+    option[field].toLowerCase().includes(inputValue),
+  )
+}
+
+function getMatchIndex(inputValue, option) {
+  const re = new RegExp(escapeRegExp(inputValue))
+  const match = re.exec(option.toLowerCase())
+  if (match) return match.index
+  return -1
+}
+
+function makeWrapper() {
+  return mount(
+    <ThemeProvider theme={theme}>
+      <SearchBox
+        filterFunction={filterFunction}
+        getMatchIndex={getMatchIndex}
+        onSubmit={jest.fn()}
+        options={searchOptions}
+      />
+    </ThemeProvider>,
+  )
+}
+
+const wrapper = makeWrapper()
+
+describe('SearchBox component tests', () => {
+  it('renders an input field', () => {
+    expect(wrapper.find('input')).toHaveLength(1)
+  })
+
+  it('has search icon', () => {
+    expect(wrapper.find('#search-icon')).toHaveLength(1)
+  })
+
+  it("doesn't show any suggestions on empty input", () => {
+    wrapper.find('input').simulate('change', { target: { value: '' } })
+    expect(wrapper.find(SearchBox).instance().state.suggestions).toEqual([])
+  })
+
+  function compareSuggestions(searchValue) {
+    const expectedSuggestions = filterFunction(
+      searchOptions,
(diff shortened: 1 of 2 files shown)


dryhten:
I was not able to get the list of suggestions from the component itself (e.g. after typing something in the search box there was no div with the list of suggestions, even after calling `component.update()`). Instead I just looked at the component state to see the list of suggestions that should be displayed in the dropdown list. This makes the assumption that the component will correctly display the suggestions from its state. If anyone knows how to get the list of suggestions from the rendered element, feel free to contribute.

app/components/ui/molecules/SearchBox.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+function getMatchIndex(inputValue, option) {
+  const re = new RegExp(escapeRegExp(inputValue))
+  const match = re.exec(option.toLowerCase())
+  if (match) return match.index
+  return -1
+}
+
+function makeWrapper() {
+  return mount(
```
jsms90: There's no theme prop being passed in to `Autosuggest` any more, so I reckon you can just do a shallow render here, as I don't think you need the `ThemeProvider`?

app/components/ui/molecules/SearchBox.test.js
```diff
@@ -0,0 +1,85 @@
...
+  it('typing something will show suggestions based on filtering function, test 1', () => {
+    compareSuggestions('option')
+  })
+
+  it('typing something will show suggestions based on filtering function, test 2', () => {
+    compareSuggestions('f')
+  })
+
+  it('typing something will show suggestions based on filtering function, test 3', () => {
+    compareSuggestions('xyz')
```
jsms90: Shouldn't this fail?

If our options are:
```
  { value: 'first option' },
  { value: 'final option' },
  { value: 'second option' },
```

then shouldn't typing in "xyz" actually show no suggestions at all? :confused:

jsms90: Ah no, I see, it's the title of the test that's misleading.

It says "typing something will show suggestions based on filtering function", but the assertion is actually checking whether the received suggestions are the same as the expected suggestions. Since we expect no results & we receive no results, the tests passes.

So perhaps the issue is just the naming of the test?

app/components/ui/molecules/SearchBox.test.js
```diff
@@ -0,0 +1,85 @@
...
+    const receivedSuggestions = wrapper.find(SearchBox).instance().state
+      .suggestions
+    expect(receivedSuggestions).toEqual(expectedSuggestions)
+  }
+
+  it('typing something will show suggestions based on filtering function, test 1', () => {
+    compareSuggestions('option')
+  })
+
+  it('typing something will show suggestions based on filtering function, test 2', () => {
```
jsms90: Is there any added value in running this test 3 times?

client/elife-theme/src/icons/material/Search.js
```diff
@@ -3,6 +3,7 @@ import React from 'react'
 const Search = props => (
   <svg
     height="24"
+    id="search-icon"
```
jsms90: Could we make this `data-test-id`?

app/components/ui/molecules/SearchBox.test.js
dryhten: The `theme` is used in the `AutosuggestWrapper`, which is part of the `SearchBox`. I didn't spend that much time digging up in enzyme to try to figure out a weird way to make it work with `shallow` (since it doesn't, e.g. there's no straightforward way to reach the `input` field) as long as it works just fine with `mount`.

app/components/ui/molecules/SearchBox.test.js
dryhten: I think the title is true, but maybe the wording is not the best. However, I wouldn't consider this of high importance, if a test fails you usually jump to the test after its description and see what's up.

app/components/ui/molecules/SearchBox.test.js
dryhten: Yes, in retrospect I'm thinking I wrote too few tests for this :) What if there is a bug in the component so that when there's a match, only 2 results are suggested all the time? What if there is a bug where the filtering function is applied in a wrong way and it only matches first words? My take is that the fewer assumptions you make about the component you're testing, the better.

client/elife-theme/src/icons/material/Search.js
dryhten: Well, you're right that I use it only in tests right now but I usually see no harm in adding ids to svg elements, just so I can target them if I need to (e.g. for applying animations).

app/components/ui/molecules/SearchBox.test.js
jsms90: Hmm I'm not so sure. The title "typing something will show suggestions" would imply to me that the test covers whether something is being rendered. But as you said above, we haven't quite got this working with enzyme. So, as it is, the suggestions are _not_ shown. In fact, this is only testing whether the state is updated & I reckon the name should reflect that?

I would say that we need to figure out how to make this work with enzyme, otherwise we're not able to actually test that the dropdown functions as it's supposed to. But I struggled with this exact thing in enzyme myself, when writing the SubjectAreaDropdown tests, so I totally get why you've opted to test the state change alone & not whether it renders.

app/components/ui/molecules/SearchBox.test.js
jsms90: Hmm, I'm still not sure that these are actually testing different pieces of the code.

First you simulate a change event on the input field. That's going to fire [the `onChange` handler in `react-autosuggest`](https://github.com/moroshko/react-autosuggest/blob/master/src/Autosuggest.js#L530-L549), which then calls the `onSuggestionsFetchRequested` prop that we've passed in.

I guess we're not trying to test whether the 3rd party component functions as it should. So these tests are all checking whether
- we have configured Autosuggest correctly in the render function (i.e. we have `onSuggestionsFetchRequested={this.onSuggestionsFetchRequested}`)
- we've written [our `onSuggestionsFetchRequested` function](https://github.com/elifesciences/elife-xpub/blob/c9cbd8a8dd633a7eb79d14d30e9e827392eb9e10/app/components/ui/molecules/SearchBox.js#L96-L110) correctly

More specifically, it's testing whether
- we have first called the `filterFunction` inside `onSuggestionsFetchRequested`
- whether state has then been updated with the return value

The only way any of these 3 tests would fail, is if we hadn't done the above correctly. So unless I'm missing something, I can't see a situation in which some of these 3 tests would pass and some would fail?

app/components/ui/molecules/SearchBox.test.js
dryhten: I am not familiar with all the tests that we have, but the ones I've written/seen needed a `ThemeProvider`, yes.

app/components/ui/molecules/SearchBox.test.js
jsms90: That being said, I do agree that we need more tests in general, as we're not covering all of the functions in the `SearchBox` component. But I imagine that the bugs you're talking about are probably best caught by other tests?

For example, the following bug would be caught by testing the filterFunction itself:
> What if there is a bug where the filtering function is applied in a wrong way and it only matches first words?

As for this one
> What if there is a bug in the component so that when there's a match, only 2 results are suggested all the time?

that could be introduced by any number of things being implemented incorrectly:
- the render function in `SearchBox`
- the `renderSuggestions` function in `SearchBox`
- the onSuggestionsFetchRequested function in `SearchBox`
- the filterFunction in `PeoplePicker`

But! If there is an issue is with `onSuggestionsFetchRequested`, then all 3 tests will fail. If there is an issue in another place in the code, then all 3 tests will still fail. Right?

So, to find the cause of that bug, we would need tests for each of these functions.

client/elife-theme/src/icons/material/Search.js
jsms90: I would probably suggest that we select it via another method than hard coding an `id` into the icon. For example, what if there were multiple search icons on a single page? Using an `id` as a selector for app functionality (rather than tests) may not be the best idea.

app/components/ui/molecules/SearchBox.test.js
dryhten: The way I'm testing things is more of black box testing[1]. Relying on implementation details when testing will cause you to catch less bugs. This is also the reason why testers are part of a different team than the developers most of the time. When I test I don't care about `react-autosuggest`, `onChange` handler or any of that. All I know is that I have a component called `SearchBox` and I know how it should behave and maybe how it should look like in the browser. Then I write tests to validate that behavior and look. If you want to write tests based on implementation details then be my guest (not that I approve of it), but I'm happy with the tests I've written.

[1] - https://en.wikipedia.org/wiki/Black-box_testing

client/elife-theme/src/icons/material/Search.js
dryhten: Highly unlikely that we have multiple search icons on a single page, but I agree with you, I'll change this to `data-test-id`.

app/components/ui/molecules/SearchBox.test.js
jsms90: Updating the name doesn't change what you're testing. It just more accurately describes what this test currently does.

If you think that testing whether the state has been updated is a bad test, then the test needs to be changed, no? Because that's exactly what it's doing, isn't it?

app/components/ui/molecules/SearchBox.test.js
```diff
@@ -0,0 +1,85 @@
...
+    )
+    wrapper.find('input').simulate('change', { target: { value: searchValue } })
+    const receivedSuggestions = wrapper.find(SearchBox).instance().state
+      .suggestions
+    expect(receivedSuggestions).toEqual(expectedSuggestions)
+  }
+
+  it('typing something will show suggestions based on filtering function, test 1', () => {
+    compareSuggestions('option')
+  })
```
damnhipster: This test looks very confusing to me, and it's not clear what the test output will be, or what's being tested here. I'd recommend putting the input and the expected outputs into the statement, or using [test.each](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout)

app/components/ui/molecules/SearchBox.test.js
```diff
@@ -0,0 +1,85 @@
...
+    option[field].toLowerCase().includes(inputValue),
+  )
+}
+
+function getMatchIndex(inputValue, option) {
+  const re = new RegExp(escapeRegExp(inputValue))
+  const match = re.exec(option.toLowerCase())
+  if (match) return match.index
+  return -1
+}
```
damnhipster: `filterFunction` and `getMatchIndex` are mocks, and it would be useful to mark them as such. Having them declared as above, with normal function names, gives the impression that they're helper functions to use within the test.

I recommend doing either of the following approaches:
```
function mockFilterFunction() {...}
```

```
let mockFilterFunction;

beforeEach(() => {
  mockFilterFunction = jest.fn();
});
...
it('shows the correct suggestions when "option" is typed in', () => {
  mockFilterFunction.returns(['expected', 'values'])
  const wrapper = mount(<SearchBox filterFunction={mockFilterFunction} />)
  expect(wrapper.instance().state.suggestions).toEqual(['expected', 'values'])
})
```

app/components/ui/molecules/SearchBox.test.js
dryhten: Not if you make a reasonable assumption and plan on changing it for the better in the future. I've seen plenty of similar assumptions in tests before, I'm just the only one drawing attentions towards it so others are aware.

app/components/ui/molecules/SearchBox.test.js
dryhten: I think the function `compareSuggestions` has a confusing name too, but I guess `test.each` is better suited here.

app/components/ui/molecules/SearchBox.test.js
dryhten: Makes sense. I'll just go with mock function implementation https://jestjs.io/docs/en/mock-functions#mock-implementations

PR #585, opened 2018-09-12
fix: prevent users from accessing or modifying others manuscripts

### LINKED ISSUE #580: Configure application permissions
**User Story:**

_As a submitter I want to only have access to information pertaining to my submissions in order to track their progress._

Functional:

- There is only one user group for the purposes of the initial submission wizard
- Each user should only receive updates to their dashboard pertaining to their own submissions

tamlyn:
#### Background

- Update authsome mode to deny all requests. Authsome is only used by built in PubSweet endpoints and resolvers so this has the effect of disabling these. 
- Update data access functions to only fetch manuscript for current user ID. This means a user can only gain access to a manuscript if they also created it.

Suggestions of alternate approaches welcome.

#### Any relevant tickets

Closes #580 

Jest integration tests for resolvers and data access tests updated.

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -43,49 +43,48 @@ const resolvers = {
 
       return manuscripts[0]
     },
-    async manuscript(_, { id }) {
-      return Manuscript.find(id)
+    async manuscript(_, { id }, { user }) {
+      return Manuscript.find(id, user)
     },
-    async manuscripts() {
-      return Manuscript.all()
+    async manuscripts(_, vars, { user }) {
+      return Manuscript.all(user)
     },
   },
 
   Mutation: {
     async createSubmission(_, vars, ctx) {
+      if (!ctx.user) {
+        throw new Error('Not logged in')
+      }
       const manuscript = Manuscript.new()
       manuscript.createdBy = ctx.user
       return Manuscript.save(manuscript)
     },
 
-    async deleteManuscript(_, { id }, ctx) {
-      const originalManuscript = await Manuscript.find(id)
-      Manuscript.checkPermission(originalManuscript, ctx.user)
+    // TODO restrict this in production
+    async deleteManuscript(_, { id }, { user }) {
+      await Manuscript.find(id, user)
 
       await Manuscript.delete(id)
       return id
     },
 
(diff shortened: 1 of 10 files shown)


server/xpub/entities/manuscript/resolvers.js
```diff
@@ -43,73 +43,68 @@ const resolvers = {
...
-      return Manuscript.all()
+    async manuscripts(_, vars, { user }) {
+      return Manuscript.all(user)
     },
   },
 
   Mutation: {
     async createSubmission(_, vars, ctx) {
+      if (!ctx.user) {
+        throw new Error('Not logged in')
```
dryhten: Log something too?

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -43,73 +43,68 @@ const resolvers = {
...
+      await Manuscript.find(id, user)
 
       await Manuscript.delete(id)
       return id
     },
 
-    async updateSubmission(_, { data }, ctx) {
-      const originalManuscript = await Manuscript.find(data.id)
-      Manuscript.checkPermission(originalManuscript, ctx.user)
+    async updateSubmission(_, { data }, { user }) {
```
dryhten: If you used destructuring here for user, why don't you use it everywhere else as well?

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -43,73 +43,68 @@ const resolvers = {
...
     },
-    async manuscript(_, { id }) {
-      return Manuscript.find(id)
+    async manuscript(_, { id }, { user }) {
+      return Manuscript.find(id, user)
     },
-    async manuscripts() {
-      return Manuscript.all()
+    async manuscripts(_, vars, { user }) {
+      return Manuscript.all(user)
```
dryhten: I assume this will change in the future based on the user role as well?

server/xpub/entities/manuscript/resolvers.js
tamlyn: Errors thrown in resolvers are logged to the console as well as being returned in the response.

dryhten: I remember having some issues with this in the past when I first started logging stuff, but can't remember exactly what.

PR #588, opened 2018-09-12
feat: connect MECA export to submission form

### LINKED ISSUE #533: remove corresponding author email from finishSubmission resolver
The corresponding author verification feature has been removed from first release:
image

Right now, [we are sending an email](https://github.com/elifesciences/elife-xpub/blob/develop/server/xpub/entities/manuscript/resolvers.js#L102-L114) when the author gets to the end of the submission wizard. This and any other associated code should be removed.

### LINKED ISSUE #566: Hook up form submission to MECA export

- Trigger the export from the finishSubmission mutation resolver
- Mutation should not await the result of the export
- If the export fails 
  - log the error
  - send an email to staff (address tbd)
  - set the manuscript `status` to `FAILED_EXPORT` (or something like that)

tamlyn:
#### Background

Closes #566 
Also closes #533 

Jest test of resolver plus check that a file was uploaded to the mock SFTP server in the E2E happy path flow.

### DIFF
--- a/config/test.js
+++ b/config/test.js
@@ -32,6 +32,7 @@ module.exports = {
       username: 'test',
       password: 'tset',
     },
+    notificationEmail: 'test@example.com',
     remotePath: '/test',
   },
   mailer: {
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -1,5 +1,4 @@
 const config = require('config')
-const User = require('../user')
 const {
   getPubsub,
   asyncIterators,
@@ -13,6 +12,7 @@ const fs = require('fs-extra')
 const path = require('path')
 const crypto = require('crypto')
 const Joi = require('joi')
+const mecaExport = require('@elifesciences/xpub-meca-export')
 
 const { ON_UPLOAD_PROGRESS } = asyncIterators
 
@@ -79,7 +79,7 @@ const resolvers = {
       return manuscript
     },
 
-    async finishSubmission(_, { data }, { user }) {
+    async finishSubmission(_, { data }, { user, ip }) {
       const originalManuscript = await Manuscript.find(data.id, user)
 
       const manuscriptInput = Manuscript.removeOptionalBlankReviewers(data)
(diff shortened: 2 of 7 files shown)


server/xpub/entities/manuscript/resolvers.js
```diff
@@ -105,19 +105,29 @@ const resolvers = {
...
 
-      const user = await User.find(ctx.user)
-      if (user.email !== modifiedManuscriptInput.author.email) {
-        mailer
-          .send({
-            to: modifiedManuscriptInput.author.email,
-            text: 'Please verify that you are a corresponding author',
+      mecaExport(manuscript.id, ctx.user, ctx.ip)
+        .then(() =>
+          logger.info(`Manuscript ${manuscript.id} successfully exported`),
```
jsms90: @hdrury1 Do we definitely not want to send an email to the author, confirming that their submission has been sent?

Normally I would agree that this is redundant, since they can just look at their dashboard. But if it's the case that some authors aren't necessarily "digital natives" or if their primary workflow is geared around using their emails, perhaps they would prefer to use their emails to track their submissions rather than being required to log in to our app?

Just a thought. What do you reckon?

test/submission.e2e.js
```diff
@@ -126,6 +130,12 @@ test('Happy path', async t => {
     .eql(manuscript.title)
     .expect(dashboard.stages.textContent)
     .eql('QA')
```
jsms90: Do we perhaps want this to change to "Exporting to eJP", until the promise has resolved & then show "QA" afterwards? That way we would be making the difference between finishing the submission wizard & finishing the submission (including getting it into eJP) more obvious to our users. At the moment we're just relying on them seeing a failure email, to let them know if it hasn't been successful, but our dashboard itself gives them a false sense of security that all is well.

@hdrury1, I assume all of this has been considered in the dashboard designs & that's a different epic. But just highlighting it here in case?

config/test.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
       username: 'test',
       password: 'tset',
     },
+    notificationEmail: '',
```
jsms90: Is it worth putting a dummy email address in here? Just so we see an email being sent from our mail client to the right address?

test/submission.e2e.js
tamlyn: Yes, it definitely shouldn't be "QA" when the stage is called Quality Check 😀 

A status of 'Submitting' followed by 'Submitted' would probably make most sense to users. It's very unlikely a user would see the 'Submitting' stage though as it will only last a second or two at most to be followed by either 'Submitted' or 'Failed'. 

Just to be clear though, the failure email would go to staff, not authors.

config/test.js
tamlyn: Yep. Amended.

test/submission.e2e.js
jsms90: Hahaha. Yeah all makes total sense.

I know the dashboard designs are still pending & this version is just a dummy. Was just wondering whether it was worth amending that now, if nothing else, just as a reminder to ourselves? Since the "QA" thing is actually in our e2e tests as well as sitting on the dashboard page.

Anyway, not necessarily important. I'm mostly pointing this stuff out as a reminder to ourselves. Happy for all this to be merged now :+1:

server/xpub/entities/manuscript/resolvers.js
hdrury1: It's been considered at length, but ultimately decided against for now. Obviously this is subject to change post-first release depending on real user feedback.

PR #589, opened 2018-09-12
remove Manuscript preview page from our app

### LINKED ISSUE #560: Update dropzone molecule to match wireframe and remove preview link
- An additional instruction should be added to the default state of the molecule (i.e. what the user sees when first loading the page)

- We should remove the link to 'preview' the manuscript until we have determined the product requirements of this feature.

**Design**

screen shot 2018-09-06 at 12 54 49

jsms90:
Planning to do a little refactoring so I'm not using `<br />` before taking this off WIP status

#### What does this PR do?
- Remove Manuscript page
- Remove preview link from FileUpload component
- Remove the clickable link from the dashboard
- Minor refactor to replace `Instruction` with `Paragraph` component in `FileUpload`

Note: design for "success" now looks like this
image

#### Any relevant tickets
fixes #560 

Removed tests that refer to this page & use the link

### DIFF
--- a/app/components/pages/Dashboard/Dashboard.js
+++ b/app/components/pages/Dashboard/Dashboard.js
@@ -1,11 +1,11 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Link } from 'react-router-dom'
 import { Box } from 'grid-styled'
 import { Button } from '@pubsweet/ui'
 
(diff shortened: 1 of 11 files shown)


jsms90:
@hdrury1 Do you think the second line should be grey? In the same way as the second line of the new error message & upload message?

PR #592, opened 2018-09-13
fix: pass user in manuscript delete mutation

### LINKED ISSUE #591: cannot delete manuscript from dashboard
**Describe the bug**
Clicking delete from the dashboard doesn't remove it, so you can't start the wizard from scratch again.

**To Reproduce**
1. Half fill in a manuscript
2. Click on 'Dashboard'
3. Click on 'Delete'
4. Nothing happens - manuscript remains on the dashboard

**Expected behavior**
Half completed manuscript should be removed & clicking 'Submit a manuscript' should allow you to start the wizard from scratch

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Firefox

tamlyn:
#### Background

Fixes #591 

Dashboard functionality is deliberately not covered by tests because it's throw away code. But obviously that was a silly decision.

### DIFF
--- a/server/xpub/entities/manuscript/index.js
+++ b/server/xpub/entities/manuscript/index.js
@@ -24,8 +24,8 @@ const Manuscript = {
   all: dataAccess.selectAll,
   findByStatus: dataAccess.selectByStatus,
 
-  delete: async id => {
-    const manuscript = await dataAccess.selectById(id)
+  delete: async (id, user) => {
+    const manuscript = await dataAccess.selectById(id, user)
     await Promise.all(manuscript.files.map(file => FileManager.delete(file.id)))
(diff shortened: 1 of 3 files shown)


jsms90:
@tamlyn is there any chance that we're going to remove the ability to delete a manuscript from our codebase altogether? Otherwise this might be an opportune moment to write tests?

But...I'm wondering what it would even mean for a someone to click 'Delete', so maybe it might be canned? What do you think?

tamlyn:
I think we will not allow manuscripts to be removed from the database in production. There may be a concept of 'Delete' which actually just sets a flag on the manuscript. 

I've added tests for the delete mutation 👍

Issue #557, opened 2018-09-06
Update limit of editor suggestions


hdrury1:
- When suggesting _either_ Senior Editors or Reviewing Editors, the user should be able to make a maximum of 6 choices

- The behaviour when deleting a selection from `submit/editors` should remain the same

- The minimum of 2 suggestions should remain the same

- The validation message `Maximum two choices` should be updated to `Maximum six choices`

jsms90:
How does this affect the designs?

Do you want the page to look something like this?
![image](https://user-images.githubusercontent.com/15656538/45483566-384f0b00-b749-11e8-811a-6b63a5da61ba.png)
![image](https://user-images.githubusercontent.com/15656538/45483621-5caae780-b749-11e8-9ee8-e66a85149c18.png)
![image](https://user-images.githubusercontent.com/15656538/45483623-5d437e00-b749-11e8-8a8d-239fe9ed8c7b.png)
![image](https://user-images.githubusercontent.com/15656538/45483792-e1960100-b749-11e8-9b37-ba11442eca07.png)
![image](https://user-images.githubusercontent.com/15656538/45483634-6cc2c700-b749-11e8-8110-afd56e5be708.png)
![image](https://user-images.githubusercontent.com/15656538/45483590-48ff8100-b749-11e8-874d-8c5aa9df7616.png)
![image](https://user-images.githubusercontent.com/15656538/45483623-5d437e00-b749-11e8-8a8d-239fe9ed8c7b.png)
![image](https://user-images.githubusercontent.com/15656538/45483623-5d437e00-b749-11e8-8a8d-239fe9ed8c7b.png)
![image](https://user-images.githubusercontent.com/15656538/45483623-5d437e00-b749-11e8-8a8d-239fe9ed8c7b.png)
![image](https://user-images.githubusercontent.com/15656538/45483841-0ab69180-b74a-11e8-859a-8428bf32c26b.png)

![image](https://user-images.githubusercontent.com/15656538/45483740-bca18e00-b749-11e8-9330-c4957713077c.png)

jsms90:
If the user:
- has already selected 6 senior editors from the People Picker
- comes back to this page from the picker
- clicks the trash icon on one pod

I assume it should look like this?
![image](https://user-images.githubusercontent.com/15656538/45483566-384f0b00-b749-11e8-811a-6b63a5da61ba.png)
[...]

Issue #448, opened 2018-08-08
People Picker position of information text


diversemix:
**Describe the bug**
The people picker has the information text on the same level as the blue "tags" - the information should be beneath these

**Screenshots**
In this case "Maximum 2 choices" should be beneath the blue people at the top.
 
![pp-wrapping](https://user-images.githubusercontent.com/8524685/43831833-54bc7e48-9afd-11e8-8d82-46653b8f5f19.png)

tamlyn:
Is this only at the smallest (single column) breakpoint? And with a standard 24px margin above and below?

hdrury1:
Closing due to duplicated ticket

PR #593, opened 2018-09-13
WIP extend maximum suggested editors to 6 instead of 2

jsms90:
#### What does this PR do?
- allow authors to suggest up to 6 senior editors & 6 reviewing editors, instead of max 2

Note: validation message shows the new max no. automatically

#### Any relevant tickets
#557 

- add min, max limits to server-side validation
- choose 6 senior editors & 6 reviewing editors in e2e test


jsms90:
@hdrury1 ok so this is currently working i.e. product requirements satisfied :+1: 

But design wise it doesn't look great.

---
With 4 people chosen:
image

---

But then 5 people:
image

---

And then with 6, when we get the message back:

image

---

Basically this isn't a new issue - it always existed on mobile, but it's more obvious now that we're allowing more than 2 reviewers.

I notice that this is actually covered by another ticket - #448. How about we merge this as it is and then, assuming you also feel that this is a significant design issue, I can pick up #448 next?

jsms90:
Getting an error in the tests within CI but seemed to be fine locally before I pushed up.

Seems to be an issue with using eLife's `/people` endpoint
image

Investigating now

tamlyn:
Did you forget to check in the new Replay recordings (since you are fetching new editors)?

jsms90:
@tamlyn Yep, exactly it. Trying to do that now. But actually the e2e tests do seems to be having problems locally (different problems) & I just didn't realise.

The whole happy path test fails for me right at the start, when it tries to fetch author details from Orcid :confused: 
image

So I can't even get to the People Picker part.

jsms90:
Oh no, that only happens when I add `REPLAY=record`

When I run the e2e test with `REPLAY=replay`, the whole thing passes fine. People Picker and all.

hdrury1:
@jsms90 Just in response to your comment, yep I'm happy for this to be merged as is once the tests pass.

PR #594, opened 2018-09-13
Update file upload design and error messaging

### LINKED ISSUE #513: Update error message for manuscript upload
- User should be able to re-attempt upload if it fails (file is too big, wrong format etc.)

fontSizeBase
colorText
colorTextSecondary
colorError (Oops!, icon, border)
colorPrimary (link to upload again)

**Design**

drop-zone-error

### LINKED ISSUE #597: Show error message when user tries to upload wrong manuscript type
**User Story:**

_As a submitter I want to be notified when I upload the wrong file type for manuscript._

Functional:
The dropzone molecule should show a helpful error message that the file type I uploaded as a manuscript is not accepted.

dryhten:
Closes #513
Closes #597 
Addresses client side portion of #595

### DIFF
--- a/app/components/ui/molecules/FileUpload.js
+++ b/app/components/ui/molecules/FileUpload.js
@@ -115,9 +115,9 @@ const DropzoneContent = ({ conversion, formError, dropzoneOpen }) => {
     return (
       <React.Fragment>
         <StyledUploadSuccessIcon />
-        <Paragraph data-test-id="dropzoneMessage">Success!</Paragraph>
-        <Paragraph>
-          <Action onClick={dropzoneOpen}>Replace</Action> your manuscript.
+        <Paragraph data-test-id="dropzoneMessage">
+          Success! <Action onClick={dropzoneOpen}>Replace</Action> your
+          manuscript.
         </Paragraph>
       </React.Fragment>
     )
--- a/app/components/ui/molecules/FileUpload.js
+++ b/app/components/ui/molecules/FileUpload.js
@@ -5,7 +5,6 @@ import styled from 'styled-components'
 import Dropzone from 'react-dropzone'
 import { ErrorText, Action } from '@pubsweet/ui'
 import { th } from '@pubsweet/ui-toolkit'
-import { get } from 'lodash'
 
 import Icon from '../atoms/Icon'
 import Paragraph from '../atoms/Paragraph'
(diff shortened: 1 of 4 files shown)


jsms90:
Just need to change [this line](https://github.com/elifesciences/elife-xpub/blob/349678b857bec58e9e5da190877b333965096100/app/components/ui/molecules/FileUpload.test.js#L24) back in the test file to get CI passing

app/components/ui/molecules/FileUpload.js
```diff
@@ -84,18 +88,16 @@ const DropzoneContent = ({ conversion, formError, dropzoneOpen }) => {
...
-          {errorMessage}. Try to <Action onClick={dropzoneOpen}>upload</Action>{' '}
-          your manuscript again.
-        </ErrorText>
+        <Paragraph data-test-id="dropzoneMessage">
+          <DropzoneErrorText>Oops!</DropzoneErrorText> Unable to upload
+          manuscript. Please <Action onClick={dropzoneOpen}>try again.</Action>
+          <StyledParagraph>
+            Files should be in .docx or .pdf format and no larger than 100MB.
+          </StyledParagraph>
+        </Paragraph>
```
jsms90: Sorry if this is a WIP because you don't want comments yet? But just noticed that this `<Paragraph>` tag is sitting around the `StyledParagraph`, so you've got a `p` within a `p`

dryhten: Hm, I guess we should change to a div then? This was fine when it was a div and not a paragraph and I do want to group both messages together so it's easier to search for the the entire message (e.g. in tests) and wrapping everything in another div is too much nesting.

tamlyn:
Hope you don't mind @dryhten but I took this over in your absence since it overlapped significantly with #597 which I was working on.

I stopped using the `Paragraph` atom since that uses the serif `fontReading` whereas the designs are all sans-serif (sorry if I missed something there). The messages are still rendered as `p` tags though. 

There's a little bit of weirdness in using `setFieldError` in that the message disappears on the next form validation (e.g. when blurring the cover letter editor). Another issue is that it favours `conversion.error` (i.e. server errors) over form errors. That means if you upload a valid file and it fails on the server, then you try to upload an invalid file and it fails on the client, the server error continues to be displayed rather than the new form error.

I think these are acceptable edge case bugs for now but happy to be convinced otherwise. The alternative is to create a stateful component which tracks errors from various sources and can choose to show the most recent error, for example.

jsms90:
Doesn't seem to work for me at all - on Ubuntu + Firefox or Chrome
Doesn't work for Tam on Mac + Firefox.

So this currently only works for Mac + Chrome (and maybe other browsers? :woman_shrugging:)

PR #596, opened 2018-09-13
fix: persist opposed reviewers reason

tamlyn:
#### Background

This fixes the issue I was seeing on iOS Safari preventing submission of the form. 

- Add opposed_reviewers_reason to SQL schema.
- Add test to ensure all manuscript properties are persisted.
- Remove opposedReviewers property now that these are handled through teams.

### DIFF
--- a/server/xpub/entities/manuscript/helpers/empty.js
+++ b/server/xpub/entities/manuscript/helpers/empty.js
@@ -9,7 +9,6 @@ const emptyManuscript = {
   },
   opposedSeniorEditorsReason: '',
   opposedReviewingEditorsReason: '',
-  opposedReviewers: [],
   opposedReviewersReason: '',
   suggestionsConflict: false,
   files: [],
--- a/server/xpub/entities/manuscript/resolvers.test.data.js
+++ b/server/xpub/entities/manuscript/resolvers.test.data.js
@@ -1,0 +1,77 @@
+exports.userData = {
+  username: 'testuser',
+  email: 'test@example.com',
+  orcid: '0000-0003-3146-0256',
+  oauth: { accessToken: 'f7617529-f46a-40b1-99f4-4181859783ca' },
+}
+
+exports.badUserData = {
+  username: 'baduser',
+  email: 'bad@example.com',
+  orcid: '0000-0001-0000-0000',
+  oauth: { accessToken: 'badbadbad-f46a-40b1-99f4-4181859783ca' },
+}
+
+exports.manuscriptInput = {
+  author: {
+    firstName: 'Firstname',
+    lastName: 'Lastname',
+    email: 'mymail@mail.com',
+    aff: 'Institution Inc',
+  },
+  cosubmission: ['More interesting observations on chickens'],
+  coverLetter: 'my cover letter',
+  disclosureConsent: true,
+  meta: {
+    articleType: 'research-article',
+    subjects: ['cancer-biology'],
+    title: 'My manuscript',
+  },
+  opposedReviewers: [{ name: 'Reviewer 4', email: 'reviewer4@mail.com' }],
+  opposedReviewersReason: 'Reviewer 4 is my wife',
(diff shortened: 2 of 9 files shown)


server/xpub/entities/manuscript/helpers/empty.js
```diff
@@ -9,7 +9,6 @@ const emptyManuscript = {
   },
   opposedSeniorEditorsReason: '',
   opposedReviewingEditorsReason: '',
-  opposedReviewers: [],
```
jsms90: Why is this being removed?

server/xpub/entities/manuscript/resolvers.test.data.js
```diff
@@ -0,0 +1,77 @@
...
+  },
+  previouslyDiscussed: 'Discussed this with bob',
+  previouslySubmitted: ['An important finding about ants'],
+  cosubmission: ['More interesting observations on chickens'],
+  submitterSignature: 'Sneha Berry',
+  disclosureConsent: true,
+}
+
+exports.manuscriptMatch = {
+  cosubmission: ['More interesting observations on chickens'],
```
jsms90: :rofl:

server/xpub/entities/manuscript/resolvers.test.js
```diff
@@ -151,99 +142,59 @@ describe('Submission', () => {
...
       )
 
       expect(returnedManuscript.status).toBe('QA')
 
       const storedManuscript = await Manuscript.find(id, userId)
-      expect(storedManuscript.status).toBe('QA')
-      expect(storedManuscript.meta.title).toBe('My manuscript')
+      expect(storedManuscript).toMatchObject({
+        ...manuscriptMatch,
+        status: 'QA',
```
jsms90: Let's not introduce 'QA' in more places :laughing:

server/xpub/entities/manuscript/resolvers.test.data.js
```diff
@@ -0,0 +1,77 @@
...
+    aff: 'Institution Inc',
+  },
+  previouslyDiscussed: 'Discussed this with bob',
+  previouslySubmitted: ['An important finding about ants'],
+  cosubmission: ['More interesting observations on chickens'],
+  submitterSignature: 'Sneha Berry',
+  disclosureConsent: true,
+}
+
+exports.manuscriptMatch = {
```
jsms90: Is there any reason why all of the keys in this object are in a different order? It might be a little simpler to follow the logic if it were in the same order as the object above?

Could we maybe call this `expectedManuscript` or `expectedManuscriptResult` or something similar? I was kind of expecting everything in this file to be data that we were testing with, as opposed to also being the results that we expect to be returned. The name might help with that? And then the test would read
`expect(actualManuscript).toMatchObject(expectedManuscript)`

server/xpub/schema/migrations/1536842027-opposed-reviewers.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE manuscript
+  ADD COLUMN opposed_reviewers_reason TEXT,
+  DROP COLUMN opposed_reviewers;
```
jsms90: Same as above, why is this being removed & the same time as we're adding a reason into the manuscript? :confused:

server/xpub/entities/manuscript/resolvers.test.data.js
tamlyn: That's just the way they came out. Makes sense to reorder and rename.

server/xpub/entities/manuscript/helpers/empty.js
tamlyn: Opposed reviewers are now being stored in teams along with all other suggestions (they we left out in the initial work because their shape was different and it would have needed special handling). So there's no need for an explicit property on the manuscript or column in the table. They still exist as a separate property on the manuscript input because that matches the shape of the form.

Issue #587, opened 2018-09-12
Write tests for Search Box icons


dryhten:
Currently we don't have any test for the Search Box icons. There is an X icon which should clear everything you typed in the box, and a search icon that updates the list of person pods in the people picker (similar to hitting enter).

- [x] Write tests for the X icon
- [x] Write tests for the search icon

dryhten:
Blocked by https://github.com/elifesciences/elife-xpub/issues/547

hdrury1:
Is this unblocked now?

jsms90:
Yup

Issue #540, opened 2018-09-03
Add option to clear search box


hdrury1:
**User Story:**

_As a submitter I want to be able to easily clear my previous search term(s) in one action, rather than have to manually delete each character_

**Product requirements/Acceptance criteria:**

Functional:

- As soon as the search box is not empty, a cross icon should appear to the right of the text field

- Clicking the cross should clear the field

- An empty field should not have a cross icon

- Clearing the search box should also update the returned Person Pods (i.e. all Pods are returned in alphabetical order)

**Relevant links to design(s):**

![screen shot 2018-09-03 at 13 26 33](https://user-images.githubusercontent.com/17023702/44987251-79436480-af7e-11e8-9c52-79071f6f3e80.png)

**Tasks:**
- [x] Create button component & add to search box
- [x] Add handler so that clicking the cross button clears the input field
- [x] Fix positioning, so that button sits on top of `react-autosuggest`'s input field
- [x] Hook up clicking functionality to PeoplePicker so that it also updates the person pods

PR #599, opened 2018-09-13
test(SearchBox): add icon tests

dryhten:
re #587

### DIFF
--- a/app/components/ui/molecules/PeoplePicker.test.js
+++ b/app/components/ui/molecules/PeoplePicker.test.js
@@ -163,5 +163,12 @@ describe('PeoplePicker', () => {
       searchFor('s')
       expect(searchWrapper.find('PersonPod')).toHaveLength(0)
     })
+
+    it('filters the people pods when you click on the search icon', () => {
+      const input = searchWrapper.find('input')
+      input.simulate('change', { target: { value: 'annie' } })
+      searchWrapper.find('[data-test-id="search-icon"]').simulate('click')
+      expect(searchWrapper.find('PersonPod').length).toBeGreaterThan(0)
+    })
   })
 })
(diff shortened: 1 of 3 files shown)


app/components/ui/molecules/PeoplePicker.test.js
```diff
@@ -163,5 +163,12 @@ describe('PeoplePicker', () => {
       searchFor('s')
       expect(searchWrapper.find('PersonPod')).toHaveLength(0)
     })
+
+    it('filters the people pods when you click on the search icon', () => {
+      const input = searchWrapper.find('input')
+      input.simulate('change', { target: { value: 'annie' } })
+      searchWrapper.find('[data-test-id="search-icon"]').simulate('click')
+      expect(searchWrapper.find('PersonPod').length).toBeGreaterThan(0)
```
jsms90: This just checks whether any results are displayed at all, no?

Wouldn't this test pass, even if the results hadn't been filtered or changed at all, so long as there were still person pods in the grid?

dryhten: Yeah, totally forgot all person pods are displayed by default. I'll change it.

dryhten:
Does clicking on the X icon clear the content of the search box only or does it submit as well? I'm not that familiar with the functionality requirements for it.

dryhten:
Ok, I found the initial issue https://github.com/elifesciences/elife-xpub/issues/540, I'll add a test for it then.

app/components/ui/molecules/SearchBox.test.js
```diff
@@ -77,4 +77,15 @@ describe('SearchBox component tests', () => {
       expect(receivedSuggestions).toEqual(expectedSuggestions)
     },
   )
+
+  it('does nothing when clicking on X icon and nothing is typed', () => {
+    wrapper.find('[data-test-id="cross-icon"]').simulate('click')
+    expect(wrapper.find('input').props().value).toEqual('')
+  })
+
+  it('clears thes input field when x icon is clicked after typing something', () => {
```
jsms90: Teeny typo "clears the**s** input field"

jsms90:
@dryhten Any thoughts on the `searchFor` naming I mentioned above?

jsms90:
> not add something ridiculous

Oh @dryhten, is there really any need to be so inflamatory? We currently have two ways of triggering a search, only one of which is named `searchFor`. I simply thought it might help for readability. If you don't think it's important, that's fine. It was only a suggestion :woman_shrugging:

Issue #595, opened 2018-09-13
Enforce 100MB file size limit on manuscripts


dryhten:
There should be a limit on the file size of manuscripts. According to this issue it should be 100MB https://github.com/elifesciences/elife-xpub/issues/513. However, even if this value changes in the future, we need to check for a maximum file size in several places and we currently don't. These places are:

- [x] On the frontend, in the `onDrop` function of the dropzone molecule: this will allow the app to fail faster if the file is too big.
- [x] On the backend in the `uploadManuscript` resolver.
- [x] The nginx config should be updated to allow for larger files to be uploaded. The limit could be bigger than 100MB if we decide at some point that we want to allow for file types other than documents to be uploaded.

There is still something left to be done for the backend check documented here https://github.com/elifesciences/elife-xpub/issues/641. I'm going to close this issue for now since I opened the other one.

PR #606, opened 2018-09-17
feat(server): add file size check in uploadManuscript resolver

dryhten:
re #595

This is the second task from #595


dryhten:
`MAX_FILE_SIZE_MB` should be stored in one place only (for both frontend and backend)

dryhten:
Yes, I was about to suggests config. I realized that we also need at least a test for this. I'll write it.

PR #610, opened 2018-09-17
fix: improve stream handling in upload resolver

tamlyn:
#### Background

Follow up to #607 in light of the fact that [you shouldn't mix](https://nodejs.org/api/stream.html#stream_choose_one_api_style) `.pipe()` and `.on('data')`. Instead this uses a transform stream to handle the upload progress reporting and then passes the data on untouched for use in the file save stream.


dryhten:
LGTM

PR #640, opened 2018-09-20
fix: remove double submit and race condition

tamlyn:
- One test was clicking *Next* twice causing validation errors to be displayed prematurely.
- Added regression test to check that validation errors are not displayed.
- Asserting on the result of the SFTP upload occasionally failed due to the time it took to upload, now it automatically retries.

### DIFF
--- a/test/submission.e2e.js
+++ b/test/submission.e2e.js
@@ -3,13 +3,13 @@ import config from 'config'
 import startSshServer from '@elifesciences/xpub-meca-export/test/mock-sftp-server'
 import replaySetup from './helpers/replay-setup'
 import {
-  dashboard,
-  wizardStep,
   author,
+  dashboard,
+  disclosure,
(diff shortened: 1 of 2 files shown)


test/submission.e2e.js
```diff
@@ -22,6 +22,22 @@ const manuscript = {
 }
 
 const getPageUrl = ClientFunction(() => window.location.href)
+const autoRetry = async (fn, timeout = 5000) => {
+  const delay = 100
+  const start = Date.now()
+  while (true) {
```
diversemix: Should we have a timeout here ... even if its 10 mins to prevent this hanging adinfinitum?

tamlyn: The function either returns on line 31 or throws on line 34. I think the only way it could hang is if `fn` doesn't return but that seems quite unlikely.

Issue #546, opened 2018-09-04
Prevent the server from sending all push back messages at once


dryhten:
**Describe the bug**
When you try to upload a file with the dropzone molecule, the websocket server will send progress messages at once, so it stays at 0% and then it suddenly jumps to 100%.

**To Reproduce**
1. Go to Files page (assuming connection to websocket was OK)
2. Upload a (larger) manuscript
3. See that after 0% it jumps really fast to 100%

**Expected behavior**
The server should send back each individual message when it's published, instead of sending them all at once.

diversemix:
I suggest we see if we can turn on `TCP_NODELAY` ... which effectively turns off Nagle's algorithm.

http://nginx.org/en/docs/http/ngx_http_core_module.html#tcp_nodelay

dryhten:
Can you test this fast enough? If this solves the issue there is no need to hide upload progress.

diversemix:
Ah! The default is `on`  ... so maybe this is not the issue. Just going to check on the actual server.

dryhten:
We tested this without nginx and the behavior was weirder: a random number of messages were sent initially, then nothing for a while and finally all the remaining messages were sent at once. I had 1-2 tries where this was working perfectly which makes it even weirder.

Another idea we had is that maybe the messages are sent too fast (5-10ms apart), so spreading them to at least 500ms apart might work. The PR for this is here https://github.com/elifesciences/elife-xpub/pull/645

dryhten:
Just tested https://github.com/elifesciences/elife-xpub/pull/645 on [kubernetes](http://review-add-delay-399fcd.gateway.xpub.elifesciences.yld.io) and it doesn't seem to solve the problem. Now the browser only receives the last message with 100%.

dryhten:
We finally got the hang of this. It turns out that the file was first uploaded to nginx and while this was happening the docker process running node.js was not receiving anything. Only after the entire file was uploaded to nginx it was passed to the docker process, resulting in the upload progress going from 0 to 100% really fast. To enable streaming `proxy_request_buffering` had to be turned off for nginx, which was done here https://github.com/elifesciences/elife-xpub-formula/pull/33/.

dryhten:
@giorgiosironi should this be reverted? https://github.com/elifesciences/elife-xpub-formula/pull/28

giorgiosironi:
[...]

PR #645, opened 2018-09-20
feat(graphql-subscriptions): wait for at least 500ms

dryhten:
wait for at least 500ms before sending upload progress

just a shot in the dark

re #546

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -164,6 +164,10 @@ ${err}`,
             }
             startTime = process.hrtime()
           }
+          logger.info(
+            'Sent websocket message to graphql-postgres-subscriptions with uploadProgress ',
+            uploadProgress,
+          )
           pubsub.publish(`${ON_UPLOAD_PROGRESS}.${user}`, {
             uploadProgress,
           })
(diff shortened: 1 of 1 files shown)


server/xpub/entities/manuscript/resolvers.js
```diff
@@ -150,14 +150,24 @@ ${err}`,
...
+          if (uploadProgress < 100) {
+            const elapsedTime = process.hrtime(startTime)
+            // MS_IN_S = 1e3, NS_IN_MS = 1e6
+            const elapsedTimeMs = elapsedTime[0] * 1e3 + elapsedTime[1] / 1e6
+            if (elapsedTimeMs < 500) {
+              return done(null, chunk)
+            }
+            startTime = process.hrtime()
+          }
           pubsub.publish(`${ON_UPLOAD_PROGRESS}.${user}`, {
```
diversemix: @dryhten  Maybe we also want to  throttle the messages to 500ms apart? ... this would prevent any flooding too?

dryhten: You mean in general for all websocket messages using something like a decorator or specifying this somewhere in config? Maybe the PR description was not good enough, but this waits for 500ms between each message, not just the first one.

PR #647, opened 2018-09-21
feat: store manuscripts in S3

### LINKED ISSUE #520: Store Manuscript uploads into S3
@hdrury1 - Not sure if this is required for release - but I'm raising this as I think it would be diligent to ensure that xpub keeps a copy of what was submitted - as well as exporting to eJP.

tamlyn:
#### Background
- Save manuscripts file contents in S3
- Restructure config to share AWS credentials between packages.
- Refactor MECA package to not access the database. This avoids circular dependency between xpub-server and xpub-meca-export packages.

#### Any relevant tickets

Closes #520 

Integration tests with mock S3 server, end to end tests updated.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -39,17 +39,22 @@ module.exports = {
     from: 'dev@example.com',
     path: `${__dirname}/mailer`,
   },
+  aws: {
+    credentials: {
+      region: '',
+      accessKeyId: '',
+      secretAccessKey: '',
+    },
+    s3: {
+      s3ForcePathStyle: true,
+      params: {
+        Bucket: 'demo-elife-xpub',
+      },
+    },
+  },
   meca: {
     s3: {
-      params: {
-        Bucket: '',
-      },
-      connectionOptions: {
-        s3ForcePathStyle: true,
-        region: '',
-        accessKeyId: '',
-        secretAccessKey: '',
-      },
+      remotePath: 'meca-archive/',
       disableUpload: false,
     },
     sftp: {
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -99,7 +99,8 @@ const resolvers = {
       manuscript.status = ManuscriptManager.statuses.MECA_EXPORT_PENDING
       await ManuscriptManager.save(manuscript)
 
-      const content = 'dummy content'
+      const sourceFile = ManuscriptManager.getSource(manuscript)
+      const content = await FileManager.getContent(sourceFile)
       mecaExport(manuscript, content, ip)
         .then(() => {
           logger.info(`Manuscript ${manuscript.id} successfully exported`)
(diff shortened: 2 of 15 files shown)


config/default.js
```diff
@@ -30,17 +30,22 @@ module.exports = {
   darServer: {
```
diversemix: I saw you removed the dar stuff... should we remove this config item too?

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -91,21 +80,22 @@ const resolvers = {
...
+      mecaExport(manuscript, content, ip)
         .then(() =>
           logger.info(`Manuscript ${manuscript.id} successfully exported`),
         )
         .catch(async err => {
           logger.error('MECA export failed', err)
-          await Manuscript.save({
+          await ManuscriptManager.save({
             id: manuscript.id,
             status: 'FAILED_MECA_EXPORT',
```
diversemix: If this is where `FAILED_MECA_EXPORT` is set on the manuscript - where is succeeded equivalent set ?

config/default.js
diversemix: I saw you removed the dar stuff .... should this go too?

server/xpub/entities/manuscript/resolvers.js
tamlyn: Currently nowhere. I think there's a separate piece of work to make more sense of the statuses. Currently it gets set to `QA` on finishSubmission and will stay that way if this doesn't fail.

config/default.js
tamlyn: Yes!

tamlyn: Actually, the DAR server package which uses this is still in there. I'll remove it in a separate PR.

diversemix: Agreed

diversemix:
I'm seeing content in AWS: https://s3.console.aws.amazon.com/s3/buckets/demo-elife-xpub/manuscripts/?region=us-east-1&tab=overview

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -94,27 +83,28 @@ const resolvers = {
...
         manuscriptInput,
       )
 
-      manuscript.status = Manuscript.statuses.MECA_EXPORT_PENDING
-      await Manuscript.save(manuscript)
+      manuscript.status = ManuscriptManager.statuses.MECA_EXPORT_PENDING
+      await ManuscriptManager.save(manuscript)
 
-      mecaExport(manuscript.id, user, ip)
+      const content = 'dummy content'
```
damnhipster: Is this going to be changed?

tamlyn: Whoa now! Yes, that needs to be changed.

jsms90:
@tamlyn and I are just rebasing on `develop`. Seeing errors in the e2e tests. Sometimes :rofl:

We reckon it's because the tests which upload the MECA files to S3 & the tests which upload the manuscript to S3 are running in parallel (default behaviour of jest) & both trying to connect a `mock-s3rver` on the same port.

After randomising the port number & running a bunch of times, we don't seem to be able to replicate the error :+1: But just wanted to record this in case we see it again in future.

Issue #659, opened 2018-09-24
Disable SFTP test


hdrury1:
In order to avoid the test failures temporarily and increase work progress.
The issue being avoided is: #657

diversemix:
This was done and now tests use SFTP appropriately

PR #660, opened 2018-09-24
Disable the test as a temporary workround

diversemix:
Implemented to stop the need to retry tests. See #659

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.test.js
+++ b/server/xpub/entities/manuscript/resolvers.test.js
@@ -191,18 +191,10 @@ describe('Submission', () => {
 
       const storedManuscript = await Manuscript.find(id, userId)
 
-      // check the object returned is what is expected, less the status
-      const expected = lodash.cloneDeep(expectedManuscript)
-      delete expected.status
-
       expect(storedManuscript).toMatchObject({
-        ...expected,
+        ...expectedManuscript,
+        status: expect.stringMatching(/^MECA_EXPORT_(SUCCEEDED|PENDING)/),
       })
-
-      // Now check status. Becaue of the race condition, check either valid value
-      expect(storedManuscript.status).toEqual(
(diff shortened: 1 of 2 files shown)


server/xpub/entities/manuscript/resolvers.test.js
```diff
@@ -189,6 +189,9 @@ describe('Submission', () => {
         Manuscript.statuses.MECA_EXPORT_PENDING,
       )
 
+      // Put in a delay to reduce risk of race condition
+      await new Promise(resolve => setTimeout(resolve, 100))
```
tamlyn: I think better would be to assert that the value is either `SUCCEEDED` or `PENDING` since we don't actually care and that covers all possibilities.

server/xpub/entities/manuscript/resolvers.test.js
```diff
@@ -190,11 +190,19 @@ describe('Submission', () => {
...
       expect(storedManuscript).toMatchObject({
-        ...expectedManuscript,
-        // TODO this might cause a race condition
-        status: Manuscript.statuses.MECA_EXPORT_SUCCEEDED,
+        ...expected,
       })
+
+      // Now check status. Becaue of the race condition, check either valid value
+      expect(storedManuscript.status).toEqual(
+        expect.stringMatching(/^MECA_EXPORT_SUCCEEDED|^MECA_EXPORT_PENDING/),
```
tamlyn: You should be able to just put this line inside the `toMatchObject` expectation. Avoids cloning and deleting properties.

tamlyn: I mean:

```
      expect(storedManuscript).toMatchObject({
        ...expectedManuscript,
        status: expect.stringMatching(/^MECA_EXPORT_(SUCCEEDED|PENDING)/),	
      })
```

PR #661, opened 2018-09-24
Specify postgres:10.4 for local usage

giorgiosironi:
See https://github.com/elifesciences/builder/pull/393 for AWS support reasons: going back to `10.4` from the current default `10.5`.

No idea where to change what is used in building PRs to this repository.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
       PGUSER: $USER
 
   postgres:
-    image: postgres:10
+    image: postgres:10.4
     ports:
       - 5432:5432
     environment:


jsms90:
> No idea where to change what is used in building PRs to this repository.

Depends what kind of development setup you're using - see pubsweet docs: https://gitlab.coko.foundation/pubsweet/pubsweet/wikis/Development:-setup

If people are developing without docker, then it's obviously whatever version of Postgres they have installed on their machine. If with docker, then I think I'm right in saying it'll be [this line in the docker compose file](https://github.com/elifesciences/elife-xpub/blob/develop/docker-compose.yml#L21)

Issue #636, opened 2018-09-20
Discussion: using profiles service instead of users table


tamlyn:
There are a few issues that need to be worked out before this work can be specced.

1. The SQL schema expects user IDs to be UUIDs but eLife profile IDs are 8 character (base 36?) strings.
   - We could change the SQL schema to `TEXT` instead of `UUID`?
   - Or convert the eLife IDs into UUID format (e.g. convert to base 16 then pad with zeros)?
   - Or keep a local users table which just maps UUID to elife ID?

2. How will this work in development?
   - Is there a staging profiles service environment which we could each have API keys for?
   - How will developers log in to xpub? Is there a staging journal server connected to the staging profiles service?

diversemix:
I'm liking the third option of (1) ... to map a UUID to an eLife ID ... reason being is that its quite flexible we could support other system's ids for users. The other two options feel like "here be dragons" :)

diversemix:
@giorgiosironi / @thewilkybarkid (point 2) Could you give us API keys for profile service in the staging environment?

diversemix:
My reply to point (2) is... could we mock out the profiles service (do we already have a mock for it) ... for devs?

giorgiosironi:
Staging is implemented by an environment called `continuumtest`:
- https://continuumtest--journal.elifesciences.org/
- https://continuumtest--profiles.elifesciences.org/ping
They will use the ORCID sandbox, which is what https://demo--xpub.elifesciences.org/ uses.

If you want to stub the `profiles` service, [`api-dummy` can be used](https://github.com/elifesciences/api-dummy/blob/develop/test/OAuth2Test.php). It's so limited (only one user, no HTML pages) that it should only be used in development and/or for running automated tests.

giorgiosironi:
The API keys for `profiles` will be the same across environments, changing only the URL to use. Should be [added to the formula](https://github.com/elifesciences/elife-xpub-formula/blob/master/salt/elife-xpub/init.sls#L28-L30) with fake pillar values, then they can be populated with real one.

giorgiosironi:
Reference from `journal`: https://github.com/elifesciences/journal-formula/blob/master/salt/pillar/journal.sls#L31-L32 two values, id and secret. Let's do this together tomorrow.

tamlyn:
[...]

PR #662, opened 2018-09-24
WIP: Using ORCID sandbox by default

damnhipster:
#### Background

This allows us to continue using ORCID's sandbox for the demo environment.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -27,6 +27,13 @@ module.exports = {
     from: 'dev@example.com',
     path: `${__dirname}/mailer`,
   },
+  'auth-orcid': {
+    // get the oauth credentials from another developer or
+    // create new ones at https://orcid.org/content/register-client-application-sandbox
+    clientID: '',
+    clientSecret: '',
+    sandbox: true,
+  },
   meca: {
     s3: {
       params: {
--- a/config/default.js
+++ b/config/default.js
@@ -28,8 +28,6 @@ module.exports = {
     path: `${__dirname}/mailer`,
   },
   'auth-orcid': {
-    // get the oauth credentials from another developer or
-    // create new ones at https://orcid.org/content/register-client-application-sandbox
     clientID: '',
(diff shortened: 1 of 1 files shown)


config/default.js
```diff
@@ -27,6 +27,11 @@ module.exports = {
     from: 'dev@example.com',
     path: `${__dirname}/mailer`,
   },
+  'auth-orcid': {
+    clientID: '',
+    clientSecret: '',
+    sandbox: true,
```
giorgiosironi: I'd prefer to have the possibility to configure any URL, as in
https://github.com/elifesciences/profiles-formula/blob/master/salt/pillar/profiles.sls#L11
where we can configure the real ORCID, the sandbox ORCID, or a [dummy](https://github.com/elifesciences/orcid-dummy). `https://orcid.org/` is probably fine-grained enough

tamlyn: Good point. But bear in mind the whole ORCID login functionality is being removed imminently, see #636

diversemix: @giorgiosironi ... Is this okay then as its just for the short-term?

config/default.js
```diff
@@ -27,6 +27,11 @@ module.exports = {
     from: 'dev@example.com',
     path: `${__dirname}/mailer`,
   },
+  'auth-orcid': {
+    clientID: '',
+    clientSecret: '',
```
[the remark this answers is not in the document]
diversemix: The way `default.js` is used is more like a template... the values are overridden in the other config files such as `local.js` and `development.js`

diversemix:
"Work in Progress" ... i.e. there is probably more to come, or don't merge just yet. 

I approved it as I was responding to an email that had the title without the "WIP"  :/

damnhipster:
Addressed by https://github.com/elifesciences/elife-xpub-deployment/pull/10/files#diff-4e5e90c6228fd48698d074241c2ba760R22

PR #666, opened 2018-09-25
Change font weight of button to be consistent with eLife

diversemix:
As per discussions with @chugginselifesciences  and @davidcmoulton  ... the font weight in a button should be 500

### DIFF
--- a/client/elife-theme/src/elements/ui/Button.js
+++ b/client/elife-theme/src/elements/ui/Button.js
@@ -38,6 +38,7 @@ export default css`
   min-width: calc(${th('gridUnit')} * 16);
   padding: calc(${th('gridUnit')} * 2);
   font-size: 14px;
+  font-weight: 500;
   text-transform: uppercase;
 
   ${props => props.small && small};


jsms90:
@diversemix @chugginselifesciences I don't think this is actually making any difference

Before:
image

After:
image

I believe this is what is happening - [the exact weight can't be found, so the browser is falling back to `400`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Meaning_of_relative_weights#Fallback_weights)

diversemix:
Yeah, @chugginselifesciences found this independently too ... we need to look into it with @davidcmoulton 
 at some point.

hdrury1:
FTR, looks different again in the review environment (running in Firefox)

screen shot 2018-09-26 at 14 01 43

jsms90:
@diversemix I believe you would have to specify which font you want (including the weight) when you first import it.

Anyway, I get that this issue is totally not a priority right now. So perhaps we should remove this line for now & raise a new issue to specify the problem that needs solving? I'm not 100% clear what this was supposed to fix. Is it only buttons that you wanted to add a different weight to? Or everywhere that we use `Noto Sans`?

hdrury1:
@jsms90 What it says on the tin- just font weight for buttons, in order to match the journal site (designs have always mirrored the UI of the journal). Except the weight of the journal site doesn't seem to change in the inspector either, so this is a wider question. You're absolutely right though, not a priority. I think we're okay to leave it in for now rather than taking the time to remove it again.

Thanks!

diversemix:
@hdrury1 Do you think its work raising a general issue to check consistency with eLife (we could refer to this useful discussion to in it) ... but something to pick up later?

PR #667, opened 2018-09-25
Disable ORCID sandbox in production

damnhipster:
#### Background

Dependant on the following: 
- https://github.com/elifesciences/elife-xpub-deployment/pull/12
- https://github.com/elifesciences/elife-xpub-formula/pull/36
- https://github.com/elifesciences/builder-configuration/pull/3

Production has ORCID secrets set to allow it to work with the sandbox disabled. Without this, production fails.

- Verified `config` locally to ensure `NODE_CONFIG_ENV` can be used to load environment specific config files within `/config`

### DIFF
--- a/config/production.js
+++ b/config/production.js
@@ -1,7 +1,10 @@
 module.exports = {
+  auth-orcid: {
+    sandbox: false
+  },
   meca: {
     sftp: {
-      disableUpload: true,
-    },
-  },
+      disableUpload: true
+    }
+  }
 }
--- a/config/production.js
+++ b/config/production.js
@@ -1,5 +1,5 @@
 module.exports = {
-  auth-orcid: {
+  'auth-orcid': {
     sandbox: false
   },
   meca: {
--- a/config/ci.js
+++ b/config/ci.js
@@ -1,0 +1,10 @@
+module.exports = {
+  'auth-orcid': {
+    sandbox: true,
+  },
+  meca: {
+    sftp: {
+      disableUpload: true,
+    },
+  },
+}
--- a/config/demo.js
+++ b/config/demo.js
@@ -1,5 +1,5 @@
 module.exports = {
-  auth-orcid: {
+  'auth-orcid': {
     sandbox: true
   },
   meca: {
--- a/config/demo.js
+++ b/config/demo.js
@@ -1,0 +1,10 @@
+module.exports = {
+  auth-orcid: {
+    sandbox: true
+  },
+  meca: {
+    sftp: {
+      disableUpload: true
+    }
+  }
+}
--- a/config/prod.js
+++ b/config/prod.js
@@ -1,10 +1,10 @@
 module.exports = {
   'auth-orcid': {
-    sandbox: false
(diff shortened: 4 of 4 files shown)


config/production.js
```diff
@@ -1,7 +1,10 @@
 module.exports = {
+  auth-orcid: {
```
tamlyn: This is not valid JavaScript.

giorgiosironi:
I'm not opposed to environment-specific configuration files, but I think this is not the right place as it violates the portability of Docker images:

- to add a new environment, we have to rebuild the container image
- to change a setting in a single environment, we have to rebuild the container image
- if there is anything secret, it can't be put in here, hence needing an external file or environment variable anyway

My suggestion is to extract this file as a template in the formula, building it with pillar values.

diversemix:
I think we have to requirements in tension then... those that @giorgiosironi has pointed out (above) and the needs of the developer to be able to easily change the configuration. 

My pov is that while the app is still relatively immature it would make sense to sacrifice the "portability" for ease and transparency of configuration. We can then move towards this once the app matures. Interested in others thoughts though ...

tamlyn:
I think Peter is right. It's also worth noting that you don't **have** to build a new image to create a new environment or override config as the `NODE_CONFIG` environment variable can be used to override any config baked into the image. And whatever happens we will have *some* config baked into the image otherwise you end up drowning in a sea of environment variables.

giorgiosironi:
> https://github.com/elifesciences/elife-xpub/blob/develop/config/custom-environment-variables.js

This looks promising to use the image as-is and provide all the environment variables that are needed from the formula. The configuration files in the image itself can then be fully overridden in new environments.

In the call later I'd make the case to call these variants `testing` and `production` (or similar) so that they can be used in multiple environments, if like in this PR they don't seem to contain specific values like `demo--xpub.elifesciences.org`.

giorgiosironi:
> you don't have to build a new image to create a new environment or override config as the NODE_CONFIG environment variable can be used to override any config baked into the image

As far as I understand this is true the first time that happens; but as soon as an additional environment is created that uses `NODE_CONFIG` the config format cannot be changed (or expanded with new values) anymore in the image, otherwise it breaks that environment?

I think this would hold for the https://github.com/elifesciences/elife-xpub/blob/develop/config/custom-environment-variables.js approach instead, as long as they have a default.

giorgiosironi:
- add a standard name like `pr-formulas.js` config file that can be used by PR builds of `elife-xpub-formula`

giorgiosironi:
* in addition to the environment-specific config files, the custom environment variables file can be used to pass in secrets (and anything) from the formula that would override what is in the `elife-xpub` repository.

giorgiosironi:
Hold on, just found out the file should be named `prod.js`:
https://github.com/elifesciences/builder-configuration/pull/3/files#r220573032

giorgiosironi:
Rename to `config/prod.js` to avoid a translation of names from Salt environments to Xpub environments; that way they coincide.

PR #668, opened 2018-09-26
Profiles service integration

tamlyn:
#### Background

- Get user information from profiles service
- Use user and identities tables to store mapping between elife ID and PubSweet UUID
- Create login mock for development and e2e tests
   - Currently configured to log in with my eLife profile ID
- Remove auth-orcid component

#### Any relevant tickets

Closes #665 

#### How has this been tested?

Unit, integration and end to end tests updated and some new ones created

Issue #642, opened 2018-09-20
MECA file needs renaming on upload


diversemix:
See Clara's comment 
> Another point that's been brought up is the need for a method to indicate when file transfers are complete, so that we don't attempt to import partway through a transfer. One option is for you to send the file initially as something other than '.meca' / '-meca.zip' --- perhaps '.meca.incomplete' or something similar --- and then rename the file to the correct name once the transfer finishes. This is probably the simplest option, but let me know if you have other thoughts on the matter.

PR #674, opened 2018-09-27
MECA rename

diversemix:
#### Background
Uploads to SFTP site as a different name, then rename when done - to prevent eJP from prematurely processing

* NOTE *  this is branched from `520-manuscript-s3` #668 ... so please merge after this.

#### Any relevant tickets

#642 

Locally with mock sftp server


diversemix:
Do you think the best plan is to merge this into `520-manuscript-s3` now and leave this PR as open and blocked... or .... do we just leave it and merge it with the other branch when the time comes?

Issue #625, opened 2018-09-20
differentiate between different types of file upload error & update UI


jsms90:
**Product requirements/Acceptance criteria:**

- Error message should change depending on nature and origin of error (user or server)

- User should use the 'try again' link to open their native file browser, or else be able to drag and drop onto the molecule to replace the file

a) User error type = incorrect format/too big

**Design**

<img width="853" alt="screen shot 2018-10-04 at 14 49 11" src="https://user-images.githubusercontent.com/17023702/46478434-e0a23d80-c7e4-11e8-87f9-35330c26fd52.png">

b) Server error

**Design**

<img width="855" alt="screen shot 2018-10-04 at 14 51 37" src="https://user-images.githubusercontent.com/17023702/46478508-0b8c9180-c7e5-11e8-892f-bab8a9d8694c.png">

Note that 'Contact Us' should lead to the corresponding page in xPub

c) User error type = no manuscript uploaded

jsms90:
Note that the way errors are current handled produces the following problems (identified in [this PR](https://github.com/elifesciences/elife-xpub/pull/594#issuecomment-421392135):
> There's a little bit of weirdness in using `setFieldError` in that the message disappears on the next form validation (e.g. when blurring the cover letter editor). Another issue is that it favours `conversion.error` (i.e. server errors) over form errors. That means if you upload a valid file and it fails on the server, then you try to upload an invalid file and it fails on the client, the server error continues to be displayed rather than the new form error.
>
> I think these are acceptable edge case bugs for now but happy to be convinced otherwise. The alternative is to create a stateful component which tracks errors from various sources and can choose to show the most recent error, for example.

jsms90:
I've captured the bugs in another ticket about refactoring into a stateful component. Going to tackle that issue before this one, so moving this back into To Do

jsms90:
@hdrury1 Now that we have many different types of error state, should we perhaps visually differentiate between these, in a way that's more obvious to the user if/when they change?

The suggestion during demo was to highlight these in red:
- That file is not supported.
- Your file is too large.
- Unable to upload your file.

Just to attract the eye to what the problem is, rather than just the "Oops!"

But I'm unsure whether that really solves the situation in which this component changes from one type of error to another :woman_shrugging: 

Either way, just wanted to record the question, so we don't forget.

jsms90:
Also, question from a product perspective around whether we should differentiate between different types of server-side error:
[...]

Issue #549, opened 2018-09-04
Fix issues with file upload feature


dryhten:
The upload progress is not completely done yet. 

The list of tasks that still need to be done:

- [x] #544 Delay connection to websocket until token is present in local storage
- [x] #548 Don't show manuscript upload progress for now
- [x] #546 Prevent the server from sending all push back messages at once
- [x] #551 Show upload progress on dropzone molecule again
- [x] #538 Upload Manuscript flaky
- [x] #546 Prevent server from sending all pushback messages at once
- [ ] #493 Fix resolvers 'extracts title from PDF' test
- [ ] #492 Subscribe error handling
- [ ] #652 Upload Progress for 90M file pauses (Bug)
- [ ] #653 Upload Progress does not resume after navigating away (Bug)
- [ ] #639 Refactor component to be stateful
- [ ] #625 Differentiate between upload errors
- [ ] #646 Fix border color when reuploading manuscript (Bug)
- [ ] #641 Perform proper check for file size limit for manuscripts on the backend (Bug)
- [ ] #839 Sort ScienceBeam extraction once and for all

hdrury1:
@diversemix I would like to sit down and figure out which of these are MVP tasks.

jsms90:
I would suggest that we probably shouldn't go to production with some of these bugs e.g. #646 means that a user sees errors when there are none & #653 could make it very unclear which file has been uploaded / whether a file has been uploaded (see @diversemix's comment on the issue).

If an author isn't convinced that the right manuscript has been uploaded, they may not trust the app enough to submit through it. So this feature, more than most, seems like a key part of the app to ensure is as bug-free as possible?

It might be worth going over the user journeys for this particular feature from scratch, considering each possible action and what a user should see when they move from one situation to another? I would be more than happy to sit down with you & @chugginselifesciences and go through the flow some time, if that would help?

diversemix:
I should have updated this... but @hdrury1 and I decided that we'd look at this with the multiple file upload feature after our current set of work... i.e. Production Readiness and integration with eLife

jsms90:
Yeah, it's in the roadmap next to the multiple file upload. To be built after the current work & after the People Picker. :+1: I just meant for the actual product/design phase of the File Upload (before it's then ready to make it into 'Now', in terms of build).

[...]

PR #680, opened 2018-09-28
Showing user an error when uploading a manuscripts to S3 fails

damnhipster:
#### Background

No longer show a successful upload when the manuscript fails to upload to S3. This would then fail the MECA ZIP file export.

Locally

screen shot 2018-09-28 at 12 34 33

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -194,7 +194,7 @@ ${err}`,
       })
 
       const saveFileStream = stream.pipe(progressReport)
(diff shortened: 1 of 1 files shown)


tamlyn:
@hdrury1 I think changing the second line of text is a different issue. It's client side and this bug fix is all server side. Also it needs a bit more specification: should it be displayed for any error including, for example, pressing next without uploading a manuscript?

jsms90:
Yeah, #625 is the issue for changing the error messages.

But the epic for finalising the manuscript upload (#549) has been deprioritised for now. So at the moment, the way the manuscript upload works has not changed since the demo.

Issue #671, opened 2018-09-27
create more comprehensive in-app documentation


jsms90:
In-app docs should contain the following:
- [x] add documentation for initial setup
- [x] add architecture summary and details
  - [x] link to architecture summary within the app's README.md
- [x] add details about configuring the app (don't forget having an S3 bucket)
- [x] link to the [eLife playbook](https://github.com/elifesciences/continuum-playbook) from any infrastructure/deployment sections
- [x] update the developers docs - to include how to run a fake s3 container.

PR #686, opened 2018-09-28
Config documentation

damnhipster:
Relates to #671 

This adds documentation for the configuration methods we've approved this week.

### DIFF
--- a/docs/configuration/README.md
+++ b/docs/configuration/README.md
@@ -1,0 +1,7 @@
+# Configuration
+
+`elife-xpub` can be configured using multiple different methods. The configuration methods, and their precedence, are as follows:
+
+- `config/default.js` is used as the base configuration
+- `config/${environment}.js` is used for environment specific configuration, which can be stored publically. The environment is specified by setting `NODE_CONFIG_ENV` to the appropriate environment.
+- `config/custom-environment-variables.js` is used for environment specific secrets. This gets populated using configuration within [builder-configuration](https://github.com/elifesciences/builder-configuration), and injected into the app using [elife-xpub-formula](https://github.com/elifesciences/elife-xpub-formula) and [elife-xpub-deployment](https://github.com/elifesciences/elife-xpub-deployment).


jsms90:
Perhaps we could just link to [the npm package's own docs](https://github.com/lorenwest/node-config/wiki/Configuration-Files)?

That would clarify that the "File Load Order" comes from `config` itself & isn't something we've chosen or altered. It would also be simpler for someone who wants to add a new config file to the project:
- they know how it's going to be used by `config` (i.e. the order)
- they'll be aware of other options that exist in `config` (e.g. `{xyz-instance}`)
- they'll know how `config` handles environment variables

Basically just demystifies where all of this is coming from & clarifies what it us and what's the 3rd party library.

If we also want to specify which files we currently use in our app, because we feel that it would be hard for someone to map what's in our app to what's in [the whole list in the official docs](https://github.com/lorenwest/node-config/wiki/Configuration-Files#file-load-order), then perhaps we could:
- specify that you need to add in a `local-development.js` file on your own machine & then specify _what_ needs to be added here while you're developing (i.e. S3 bucket ... not sure there's anything else any more)
- mention the non-environment related files: `authsome.js`, `mailer.js` & `components.json` and explain what they're used for (?)  
Perhaps authsome and mailer are self-explanatory, but I don't think `components.json` is obvious
- call it `{deployment}.js` rather than `{environment}.js`, simply to keep things the same as they are in the official docs (?)


Perhaps the `{deployment}.js` line could link directly to our documentation about which environments we actually use? That way, it will be really clear for someone who reads our docs and then looks in our app, that the files that fall under the `{deployment}.js` umbrella are: `default.js`, `prod.js`, `ci.js`, `demo.js` & `development.js` (vs authsoe, mailer, etc).

What do you think?

jsms90:
Also might be worth just popping this info straight into the ["initial steup"](https://github.com/elifesciences/elife-xpub/tree/develop/docs#initial-setup) portion, rather than having another nested folder?

Think of clicking around docs the same way that you think of clicking around an app - the fewer clicks you can make in order to complete the task you came for, the better. Right?

giorgiosironi:
> Perhaps we could just link to the npm package's own docs

I agree if there's an authoritative source, it should be referenced here; some additional information such as the links to eLife's repositories is still required.

PR #690, opened 2018-10-01
fix: unsafe piping of streams in upload

tamlyn:
#### Background

Our internet was very slow in the office on Friday and it exposed a bug in the way streams were handled in the upload resolver. Because connecting to S3 took so long, it allowed the stream to be emptied before it had a chance to also pipe the contents to sciencebeam making sciencebeam fail with a `no content` error.

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -192,20 +192,19 @@ ${err}`,
           )
         }
       })
-
-      const saveFileStream = stream.pipe(progressReport)
-      await FileManager.putContent(
-        fileEntity,
-        saveFileStream,
-        { size: fileSize },
-      )
-
       // also send source file to conversion service
       const convertFileRequest = request.post(config.get('scienceBeam.url'), {
         qs: { filename },
         headers: { 'content-type': mimetype },
       })
+
+      // note: if a stream is piped to multiple destinations, ensure the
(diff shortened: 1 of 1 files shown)


server/xpub/entities/manuscript/resolvers.js
```diff
@@ -192,20 +192,19 @@ ${err}`,
...
-
       // also send source file to conversion service
       const convertFileRequest = request.post(config.get('scienceBeam.url'), {
         qs: { filename },
         headers: { 'content-type': mimetype },
       })
+
+      // note: if a stream is piped to multiple destinations, ensure the
+      // pipes are set up in the same tick, or else chunks can be lost
+      const saveFileStream = stream.pipe(progressReport)
```
diversemix: Might be missing something - but how does this ensure it happens in the same tick?

tamlyn: By "tick" I mean [event loop tick](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/). The problem before was that we had an `await` between the two `pipe`s. Now we can guarantee that both `stream.pipe`s will be called before the stream starts flowing. A different stream is passed to `FileManager.putContents` so we don't need to worry about that one.

tamlyn:
Something still not quite right here. Investigating.

tamlyn:
Closing in favour of #697

Issue #673, opened 2018-09-27
Configure Mailer for failure alerts/notifications


damnhipster:
**User Story:**

_As a DevOps engineer I want to be notified of failures in Production in order to avoid downtime on the xPub submission system._

**Product requirements/Acceptance criteria:**

Functional:
- Successfully send emails in Production when there are failures on upload/export

dryhten:
Just to make sure, this is about both configuring `nodemailer` to use the right smtp server and sending the actual mails when the log level is greater than warning for example?

damnhipster:
Mailer is only used when we have export failures, which is when we'll want to send the mail out.

diversemix:
I'd like to understand how this will be configured in `prod` and `staging` before we consider the issue "closed" 

I don't think should prevent us from merging this - as `localhost` seems sensible for now, but currently I don't have a clear idea on how we expect this to work. I don't think we deploy any SMTP container (e.g. `namshi/smtp`). Again maybe @giorgiosironi  could help us :grin:

dryhten:
Ok, then there is nothing else to do other than configuring the mailer to use the right smtp. The code that sends the e-mail on export failure is already in develop https://github.com/elifesciences/elife-xpub/blob/develop/server/xpub/entities/manuscript/resolvers.js#L118.

I guess there is no point in addressing the question "Should we send an e-mail to DevOps every time we log something of a particular level? (e.g. log level less than 1 which is warn & error)" right now.

@diversemix I would assume the production vm will have postfix running as well but maybe @giorgiosironi is better to answer this.

giorgiosironi:
We usually go for:

- `end2end`: configure the application to talk to a fake mail server that captures emails and allow end2end tests to check them.
- `staging`: talk to Postfix, mails end up in https://mailtrap.io/ or similar
- `prod`: talk to Postfix, mails are delivered for real with AWS SES

dryhten:
This was done in https://github.com/elifesciences/elife-xpub/pull/694 but it was never tested on the demo environment because it was not straightforward to do. The consensus was that if it doesn't after it gets deployed we will open a new issue.

PR #694, opened 2018-10-01
feat(mailer): configure nodemailer with the right smtp server

dryhten:
re #673

### DIFF
--- a/config/mailer.js
+++ b/config/mailer.js
@@ -1,10 +1,6 @@
 module.exports = {
   transport: {
-    host: 'smtp.ethereal.email',
-    port: 587,
-    auth: {
-      user: 'g3jjujasexmtz4dl@ethereal.email',
-      pass: 'QF2bP2FcHa5cVCVPPT',
-    },
+    host: 'localhost',
+    port: 25,
   },
 }


config/mailer.js
```diff
@@ -1,10 +1,6 @@
 module.exports = {
   transport: {
-    host: 'smtp.ethereal.email',
-    port: 587,
-    auth: {
-      user: 'g3jjujasexmtz4dl@ethereal.email',
-      pass: 'QF2bP2FcHa5cVCVPPT',
-    },
+    host: 'localhost',
```
diversemix: Do we want to configure this differently on a per-environment basis? Or are these settings ok for all? @giorgiosironi - what would you recommend?

dryhten:
This needs to be tested in a production-like environment before removing the `WIP` tag.

diversemix:
@dryhten ... I tested this on `demo` with the "mail" command and I did receive the mail.

dryhten:
@diversemix I did that a couple of times as well. I was thinking we need to test from the app that `nodemailer` will actually send the mail to `postfix` with these credentials. In theory it should work, postfix has a `mynetworks` config for trusted networks which is set to `127.0.0.0/8` so everything coming from localhost should be trusted and no authentication should be required, but you never know what might go wrong.

Issue #650, opened 2018-09-21
Decide for a minimum time delay between upload progress messages


dryhten:
When an user uploads a manuscript we send back the upload progress via websockets. If we send back the progress every time we receive a chunk of data (roughly 64K) then we end up sending a message every 5-10ms which seems like a waste of resources. Right now we have a minimum delay of 500ms between 2 consecutive messages. We have to decide if we need this delay or what its value should be.

Just a side note: no matter what we decide to do there is still some work here to be done e.g. the minimum delay should be stored somewhere in config and not hardcoded in the resolver.

diversemix:
Based on a TV framerate of 50fps ... this is 20ms ... I would say it should not be less than that. I am happy with 500ms.

diversemix:
Closing as this we have decided on 200ms and all has been fine.

PR #697, opened 2018-10-01
fix: don't use streams in upload resolver

tamlyn:
#### Background

We've had several bugs related to the way streams are handled in the upload resolver. I've spent most of today trying to make sense of it but in the end I've opted to read the whole file into a string and pass that around instead. It's not as efficient but it's simpler and less buggy. Optimisation can come later.

I'll try and write up the issues tomorrow.

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -9,7 +9,7 @@ const request = require('request-promise-native')
 const { promisify } = require('util')
 const xml2js = require('xml2js')
 const Joi = require('joi')
-const { Transform } = require('stream')
+const lodash = require('lodash')
 const mecaExport = require('@elifesciences/xpub-meca-export')
 
 const { ON_UPLOAD_PROGRESS } = asyncIterators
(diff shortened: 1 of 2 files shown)


diversemix:
Be good to chat about this today. Particularly to understand why it's less efficient

diversemix:
Having done a little reading I guess we should consider using `pipeline` when we move to node v10

tamlyn:
I find this a bit tricky to explain but here goes:

The code as it stands (simplified) is:
```js
      const saveFileStream = stream.pipe(progressReport)
      await FileManager.putContent(fileEntity, saveFileStream)
      const convertFileRequest = request.post(/* science beam */)
      stream.pipe(convertFileRequest)

      try {
        const xmlBuffer = await convertFileRequest
        //...
```

This issue here is the `await` on the second line. It waits for the file to be fully uploaded to S3 before piping the same stream to science beam. But by that time the stream has been drained and so science beam fails with a `no content` error. We didn't catch this because in the tests we playback a recording of the response from science beam.

We can fix this by piping the stream to both destinations in the same event loop tick:
```js
      const convertFileRequest = request.post(/* science beam */)
      stream.pipe(convertFileRequest)
      const saveFileStream = stream.pipe(progressReport)
      await FileManager.putContent(fileEntity, saveFileStream)

      try {
        const xmlBuffer = await convertFileRequest
        //...
```

This works until science beam throws an error. If this happens before the S3 upload has completed this leads to an unhandled promise rejection warning since the promise hasn't yet been awaited. The app still works but in future Node (probably 11) this will crash the process. Also it automatically fails the Jest test. Not ideal.

Instead we can use PassThrough streams:
```js
      stream.pipe(progressReport)
      const saveFileStream = stream.pipe(new PassThrough())
      const convertStream = stream.pipe(new PassThrough())
      await FileManager.putContent(fileEntity, saveFileStream)
      const convertFileRequest = request.post(/* science beam */)
      converStream.pipe(convertFileRequest)

      try {
        const xmlBuffer = await convertFileRequest
        //...
```
PassThrough stream just spits out whatever it receives but it has all the usual buffering and back pressure so now draining the `saveFileStream` doesn't affect the `convertStream`. This works for small files but for large files, the upload appears to hang. The problem is that the `convertStream` is not being drained so the backpressure eventually stops the original upload `stream` from flowing but this means the `saveFileStream` also stops flowing, the S3 upload never completes and we're stuck.

I'm sure there is a way to do this properly but there are so many timing related edge cases (problems which have gone unnoticed through code review and the test suite) that I don't feel confident we can be sure that any implementation we come up with is correct and safe.

Reading it all into a string is much simpler. We have a single stream which is converted into a promise then it's promises all the way down. 🐢 But as is the case with turtles, it's slower because it has to wait until the upload from the browser has completed before it can start uploading to S3 and then science beam. It also likely uses more memory which could become significant if simultaneously handling multiple 100MB uploads.

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -164,52 +164,43 @@ ${err}`,
...
       const pubsub = await getPubsub()
-      let uploadedSize = 0
-      let startTime = process.hrtime()
-      const progressReport = new Transform({
-        transform(chunk, encoding, done) {
+      const reportProgress = lodash.throttle(progress => {
+        pubsub.publish(`${ON_UPLOAD_PROGRESS}.${user}`, {
+          uploadProgress: Math.floor(progress * 100),
+        })
+      }, 200)
```
dryhten: Every 200ms now?

tamlyn: Oh, sorry. I was trying different values when it was getting stuck due to the streams being full and forgot to change back. Should I revert to 500?

dryhten: Either that or comment here :)
https://github.com/elifesciences/elife-xpub/issues/650

damnhipster:
@hdrury1 This change means we'll be running the upload and conversion sequentially. Will we want to display this to the user, given that conversion occurs after the 100% upload message?

hdrury1:
@diversemix does this have any effect on the conversation we were having yesterday about multiple file upload?

diversemix:
@hdrury1 ...Yes it may effect what we want to present to the user. On the premise that we don't want to expose the user to the fact we are using sematic extraction - I would suggest we leave things as they are for MVP. The net result is a slight pause after the manuscript upload completes (stuck at 100%) while it goes off and does the extraction. Other alternatives to this could be to show a message like "Processing..." but that then begs a question for the user "what is it doing?"

PR #698, opened 2018-10-02
Simplified configuration within `config/`

damnhipster:
#### Background

Addresses https://github.com/elifesciences/elife-xpub/pull/686#issuecomment-425504373

- [x] link to node-config docs
- [x] local development config
- [x] document other config files


docs/configuration/README.md
```diff
@@ -1,7 +1,27 @@
...
 - `config/default.js` is used as the base configuration
 - `config/${environment}.js` is used for environment specific configuration, which can be stored publically. The environment is specified by setting `NODE_CONFIG_ENV` to the appropriate environment.
 - `config/custom-environment-variables.js` is used for environment specific secrets. This gets populated using configuration within [builder-configuration](https://github.com/elifesciences/builder-configuration), and injected into the app using [elife-xpub-formula](https://github.com/elifesciences/elife-xpub-formula) and [elife-xpub-deployment](https://github.com/elifesciences/elife-xpub-deployment).
+
+
+### Other methods
+
+Other config methods are used for specific functionality, including:
+
+- [pubsweet-components](https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/components) - used for loading server-side components
```
dryhten: Maybe write the filename where you specify the config for these (e.g. `config/components.json`) and explain how to change this file? The link you provided doesn't show me how to change it. Also, wasn't this for all type of components both frontend and backend?

dryhten: The only example on `components.json` I found is in the book here (page 82)
https://coko.foundation/wp-content/uploads/2018/07/Coko_Pubsweet-screen.pdf

damnhipster: Thanks, I couldn't find the book myself.

tamlyn:
I find this a bit confusing. The way I see it `node-config` is the only way to configure the app. Granted it has many ways in which configuration can be provided but all config flows through this package. `components.json`, `authsome.js` etc are all managed via `node-config` and can be overridden in the usual ways as documented in that repo.

docs/configuration/README.md
damnhipster: I'm hesitant to put more code references that what's already there, since it's stuff that would have to be updated/maintained if we were to change anything related to pubsweet configuration. I'm happy to state that all config files live inside `config/`, and linking to the official docs should be enough for people to find the relevant files.

damnhipster:
@tamlyn In that case, we can look at removing the separate config files instead. There's no reason to have them separated and can lead people to believe that the application depends on the existence of those config files.

`node-config` would then be the only documented way to configure the application.

Issue #620, opened 2018-09-18
Configure xPub to use database


damnhipster:
Demo and Production are currently configured to use a postgres instance set up locally on the host, alongside the xPub application container. This is not a production ready solution.

Once we have a persistent, managed database for Demo and Production, we should configure the application to use it.

**Staging**:
- [x] add username, password and hostname for postgres to xpub
- [x] setup database using `npx pubsweet setupdb --clobber` and generate and store admin user secrets
- [x] Test out the environment

**Production**:
- [x] add username, password and hostname for postgres to xpub
- [x] setup database using `npx pubsweet setupdb --clobber` and generate and store admin user secrets
- [x] Test out the environment

**Tasks**:
- [ ] ~run migrations as part of deployment, and test them out in staging before pushing to production~
- [x] remove integration with local DB in demo environment

Considerations
- !!! **MAKE SURE PRODUCTION CONFIG IS NOT POSTED IN SLACK OR ANY UNENCRYPTED CHANNEL OF COMMUNICATION**; use something like keybase instead
- Use the recommended approach for `builder` applications to store the config, or use AWS Secrets Manager and configure xPub to fetch from there
- If there are network errors, they may be related to security groups, and VPC/subnet issues

giorgiosironi:
Reference for [how we magically switch from RDS to a local `postgres` container](https://github.com/elifesciences/digests-formula/blob/master/salt/digests/init.sls#L78-L91) depending on whether RDS is present at all in an environment.

giorgiosironi:
I assign myself (plus someone else on the application side) to provide support for the formula changes.

diversemix:
@damnhipster and @dryhten will start work on this soon in the `prod` environment as demo is being switched to staging.

damnhipster:
- [ ] `db.end` error when running the migration
- [ ] GraphQL error when visiting staging when not logged in

damnhipster:
Tested Staging and Production environments.

These can be found at the following URLS:

|Environment|Link|
|-|-|
|Staging|https://staging--xpub.elifesciences.org/|
|Production|https://xpub.elifesciences.org/|

PR #701, opened 2018-10-02
Mapping database details to environment variables

damnhipster:
#### Background
Relates to #620 

The variables already exist on the host. Currently waiting on https://github.com/elifesciences/elife-xpub-deployment/pull/16 before they're available on the container.

Tested locally

### DIFF
--- a/config/custom-environment-variables.js
+++ b/config/custom-environment-variables.js
@@ -1,5 +1,11 @@
 module.exports = {
   'pubsweet-server': {
+    db: {
+      user: 'PGUSER',
+      host: 'PGHOST',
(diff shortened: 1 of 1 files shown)


config/custom-environment-variables.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
   'pubsweet-server': {
+    db: {
+      user: 'PGUSER',
+      host: 'PGHOST',
+      database: 'PGDATABASE',
+      password: 'PGPASSWORD',
```
dryhten: Might as well add the port.

PR #705, opened 2018-10-03
feat: HTTP endpoint to receive response of MECA import

### LINKED ISSUE #644: MECA - create an endpoint for import results from eJP
Provide eJP with an endpoint that they can call back on once a package has been imported. I'm thinking of something of the form https://xpub.elifesciences.org/meca-result/<manuscript_id> ... with some suitable json as to the result of the import.

Would be good to suggest the json maybe just :
```
{  result: "success" } 
```

tamlyn:
#### Background

- Add HTTP endpoint for eJP to call back on
- Change export signature of MECA package to support hooking into pubsweet-server
- Restructure some files in the MECA package (the diff shows `export.js` as a new file even though it was just renamed)
- Fix a bug in `ManuscriptManager` which silently ignored updates to a non-existent manuscript
- Fix a bug in S3 mock server's random directory name

Closes #644 

Jest integration tests


config/default.js
```diff
@@ -67,6 +67,7 @@ module.exports = {
       remotePath: '',
       disableUpload: false,
     },
+    apiKey: '',
```
diversemix: Can we add this to `custom-environment-variables.js`? @giorgiosironi ... do you have an opinion?

server/meca/routes.js
```diff
@@ -0,0 +1,39 @@
+const logger = require('@pubsweet/logger')
+const config = require('config')
+// TODO avoid this recursive dependency
+const ManuscriptManager = require('@elifesciences/xpub-server/entities/manuscript')
```
damnhipster: We should include this within `server/meca/package.json`

server/meca/routes.js
```diff
@@ -0,0 +1,39 @@
+const logger = require('@pubsweet/logger')
+const config = require('config')
+// TODO avoid this recursive dependency
```
damnhipster: It's worth putting this TODO into an issue.

Incorporating pubsweet's full data model implementation should solve this.

tamlyn: See #708

server/meca/routes.js
```diff
@@ -0,0 +1,39 @@
...
+const ManuscriptManager = require('@elifesciences/xpub-server/entities/manuscript')
+
+module.exports = app => {
+  const apiKey = config.get('meca.apiKey')
+
+  app.post('/meca-result/:id', (req, res) => {
```
damnhipster: It's worth noting that this is the only REST endpoint in the project, and everything else uses the GraphQL API

diversemix: Did you have any particular way you wanted to "note" this?

diversemix:
@tamlyn ... looks like the test is timing out when it successfully completes because of an open handle. See https://gitlab.com/elifesciences/elife-xpub/-/jobs/104279083

server/meca/routes.js
```diff
@@ -0,0 +1,39 @@
...
+
+    ManuscriptManager.save({ id: manuscriptId, status })
+      .then(() => res.sendStatus(204))
+      .catch(err => {
+        logger.error('Failed to process MECA callback', {
+          manuscriptId,
+          error: err.message,
+        })
+        res.status(500).send({ error: err.message })
+      })
```
damnhipster: It may be prudent to declare the catch statement against the `.save` promise. If `res.sendStatus` throws an error, it would get caught and another error may get thrown by `res.status(500)`

damnhipster: Not worth addressing

server/meca/Readme.md
```diff
@@ -0,0 +1,28 @@
...
+        .then(() => console.log('Successfully uploaded archive to SFTP'))
+        .catch(err => console.error(err))
+
+## Export callback endpoint
+
+When the MECA archive has been collected from the SFTP server and processed by the third party,
+they are expected to call a REST endpoint to notify xPub of the result of the import.
+
+This endpoint is authenticated with an API key and accepts a JSON payload of either
+`{"result": "successs"}` or `{"result": "failure", "error": "...", ...}` where `error`
```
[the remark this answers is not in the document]
tamlyn: Good spot. There appears to be a bug in my markdown plugin that it doesn't underline spelling errors inside code blocks.

tamlyn:
Note this was caused by duplicate `pubsweet-server` dependency. The test we set up failed to catch it because it only looks at the root `node_modules` and not other yarn workspaces.

PR #706, opened 2018-10-03
fix: prevent updating an entity which doesn't exist

tamlyn:
#### Background

Fix a bug in the entity access logic that silently ignored saving an entity which had an ID but didn't exist in the database.

Jest integration tests.

### DIFF
--- a/server/xpub/entities/file/index.js
+++ b/server/xpub/entities/file/index.js
@@ -23,12 +23,16 @@ const FileManager = {
   delete: dataAccess.delete,
   new: (props = {}) => lodash.merge({}, empty, props),
   save: async file => {
+    let id = { file }
     if (file.id) {
-      await dataAccess.update(file)
-      return file
+      const updated = await dataAccess.update(file)
+      if (!updated) {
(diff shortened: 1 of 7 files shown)


server/xpub/entities/file/index.js
```diff
@@ -23,12 +23,16 @@ const FileManager = {
   delete: dataAccess.delete,
   new: (props = {}) => lodash.merge({}, empty, props),
   save: async file => {
+    let id = { file }
     if (file.id) {
-      await dataAccess.update(file)
-      return file
+      const updated = await dataAccess.update(file)
+      if (!updated) {
+        throw new Error('File not found')
```
diversemix: Does this throw when the item is identical and so not updated? Is this desired behaviour?

tamlyn: Ah no, sorry bad variable name. `updated` is the number of rows that were updated.

diversemix: ok ... so if there is nothing to update because all the values are the same does this still throw? I assume it would as `updated == 0`

Issue #703, opened 2018-10-03
App crashes when uploading large file


diversemix:
The application crashed (logs below) when uploading a large 50Mb random file (I created using `dd`) 

```
info: ::ffff:172.18.0.1 [03/Oct/2018:08:10:51 +0000] POST /graphql 200 CurrentSubmission 1695 10.697 ms

info: ::ffff:172.18.0.1 [03/Oct/2018:08:10:55 +0000] POST /graphql 200 UpdateSubmission 1694 40.359 ms

info: ::ffff:172.18.0.1 [03/Oct/2018:08:10:55 +0000] GET /favicon.ico 304 - - 0.482 ms


<--- Last few GCs --->

[1:0x3b46e20] 56935220 ms: Mark-sweep 366.9 (379.6) -> 366.8 (380.1) MB, 127.7 / 0.0 ms  (+ 12.8 ms in 1 steps since start of marking, biggest step 12.8 ms, walltime since start of marking 2641 ms) allocation failure GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0xba245da5ee1 <JSObject>
    2: derez(aka derez) [/home/xpub/node_modules/cycle/cycle.js:91] [bytecode=0x299c0592749 offset=266](this=0x35e2d2a82311 <undefined>,value=0x969423e57a9 <Uint8Array map = 0x3b790e440631>,path=0x5f8a74d7b59 <String[29]: $["error"]["options"]["body"]>)
    4: derez(aka derez) [/home/xpub/node_modules/cycle/cycle.js:93] [bytecode=0x299c0592749 offset=360](this=0x35e2d2a82311 <undefined>,value=0x3...

FATAL ERROR: invalid table size Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x121a2cc [node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::OrderedHashTable<v8::internal::OrderedHashSet, 1>::Allocate(v8::internal::Isolate*, int, v8::internal::PretenureFlag) [node]
 6: v8::internal::OrderedHashTable<v8::internal::OrderedHashSet, 1>::Rehash(v8::internal::Handle<v8::internal::OrderedHashSet>, int) [node]
[...]

PR #715, opened 2018-10-04
fix(uploadManuscript): only log error message on science beam failure

dryhten:
re #703

### DIFF
--- a/server/xpub/entities/manuscript/resolvers.js
+++ b/server/xpub/entities/manuscript/resolvers.js
@@ -214,7 +214,7 @@ ${err}`,
         }
       } catch (error) {
         logger.warn('Manuscript conversion failed', {
-          error,
+          error: error.message,
           manuscriptId: id,
           filename,
         })


server/xpub/entities/manuscript/resolvers.js
```diff
@@ -214,7 +214,7 @@ ${err}`,
         }
       } catch (error) {
         logger.warn('Manuscript conversion failed', {
-          error,
+          error: error.message,
```
diversemix: So I am guessing that `error` had the entire file contents in as well?

diversemix: Are there any other members of `error` that might be useful, other than `message`?

dryhten:
Just tested this on staging environment and it solved https://github.com/elifesciences/elife-xpub/issues/703 but there still seems to be a memory leak on file upload. After the upload is done the memory doesn't go down by the size of the file, so if you keep uploading files the node process will use more and more memory. This needs to be tested (maybe there is a garbage collecting mechanism that will eventually take care of it). I will open an issue for it.

diversemix:
I'm not sure you would expect the memory to go down, this is normally considered wasteful. Then again I'm not sure how the garbage collector works in V8

server/xpub/entities/manuscript/resolvers.js
dryhten: I think the message is the most relevant. There is also
name: the name of the error (e.g. `StatusCodeError`)
status code: e.g. 500 (this is included in `error.message` as well)
error: the error generated (this is included in `error.message` along with the code)
options: the options/arguments for `request.post`
constructor: the constructor for the error object (the actual code)
response: the response object from `request.post` (including the file)

diversemix: Let's leave it as-is, happy for it to get merged.

PR #716, opened 2018-10-04
feat: duplicate package checker that works with workspaces

tamlyn:
#### Background

Having duplicates of certain dependencies can cause hard-to-debug problems. The previous solution we had for checking duplicates only looked in `node_modules` and wasn't aware of yarn workspaces. This quick fix looks at the `yarn.lock` file and errors if any of the named packages are present in more than one version.

Manually.


tamlyn:
You could take the view that this is itself a test :)

Issue #699, opened 2018-10-02
Introduce time limit on ScienceBeam extraction


hdrury1:
_As a user, if the title is not extracted from my article file in three seconds I would rather have the option of entering it manually into the submission form, rather than waiting indefinitely for extraction to happen._

After an article has been uploaded, and extraction has been attempted for three seconds, regardless of the outcome we should redirect the user to the 'Submission' page. If the title has not been extracted they can enter it manually into the title field.

We should **not** give feedback that extraction has failed.

PR #722, opened 2018-10-05
feat(uploadManuscript): add timeout for science beam conversion

dryhten:
re #699


hdrury1:
@dryhten can you confirm what the user sees when the session times out?

dryhten:
@hdrury1 the user sees what he would see if the science beam fails, which I guess is "Success..." now since the science beam conversion is the last step being performed, if everything else is successful.

Issue #626, opened 2018-09-20
Search for matches anywhere in the Person Pod


hdrury1:
Currently a user can only search for a person by name.

- The application should match search terms that appear anywhere in the Person Pod (name, institution, keywords)

dryhten:
Right now the Person Pods only contain name and subject areas. We should open a new issue to pull institution as well from the elife api (if there isn't one already).

PR #754, opened 2018-10-08
feat(SearchBox): search for matches anywhere in the person pod

dryhten:
re #626

### DIFF
--- a/app/components/ui/molecules/PeoplePicker.js
+++ b/app/components/ui/molecules/PeoplePicker.js
@@ -153,48 +153,9 @@ class PeoplePicker extends React.Component {
     const inputValue = searchValue.trim().toLowerCase()
     if (!inputValue) return people
 
-    if (inputValue.split(' ').length > 1) {
-      return people.filter(person =>
-        person[field].toLowerCase().includes(inputValue),
-      )
-    }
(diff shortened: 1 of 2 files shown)


app/components/ui/molecules/PeoplePicker.js
```diff
@@ -209,13 +170,14 @@ class PeoplePicker extends React.Component {
     let extendedPeople = [...people].sort((a, b) =>
       a.name.localeCompare(b.name),
     )
-    // TODO add person.aff once we have that info
+
     extendedPeople = extendedPeople.map(person => ({
       ...person,
-      searchValue: `${person.name} ${person.subjectAreas.join(' ')}`,
+      searchValue: `${person.name} ${person.subjectAreas.join(' ')} ${
+        person.aff ? person.aff : ''
```
dryhten: Not entirely sure now that there is only one affiliation for each person? It's called `affiliations` after all.

diversemix: Good point - but we can fix that later

dryhten: Just like we could have added `person.aff` later once we have it (which is what I would've done) but this is fine as well :)

Issue #629, opened 2018-09-20
Remove search box dropdown


hdrury1:
Currently the search box dropdown shows only a list of people that match the current input. This information is a duplicate of the returned person pods.

In the future, we may wish to consider an auto-suggest feature but this requires user testing to flesh out product requirements. For now we should remove the dropdown all together.

dryhten:
I see two main approaches here:

1. We keep the react autosuggest and hide the suggestions list. There are multiple ways of doing this, the best one I can think of would be to remove the `onSuggestionsFetchRequested` prop that we pass to `Autosuggest` (commenting out one line).
2. Completely remove the react autosuggest component and replace it with a styled input field.

It all depends on how likely we are to come back to it later. If it's very likely I would go with approach 1 so we don't have to rewrite a bunch of code in the future (and I would also open an issue to keep track of it). If it's not that likely we can go with option 2 since it's not really an autosuggest element and the code might be confusing.

PR #763, opened 2018-10-09
feat(SearchBox): temporarily hide list of suggestions

dryhten:
re #629


app/components/ui/molecules/SearchBox.test.js
```diff
@@ -54,30 +54,6 @@ describe('SearchBox component tests', () => {
     expect(wrapper.find('[data-test-id="search-icon"]')).toHaveLength(1)
   })
 
-  it.each([[1, 'option'], [2, 'f'], [3, 'xyz'], [4, '']])(
```
dryhten: I would just skip this since we are going to go back at having an autosuggest in the future.

PR #764, opened 2018-10-09
Update development docs

damnhipster:
- Made `docs/development/quick-start.md` easier to find
- Added a section on starting dependencies locally
- Using `yarn` rather than `npm run`, since we're already using yarn to install dependencies


diversemix:
Looks good - thanks @damnhipster

PR #765, opened 2018-10-09
feat: add login interstitial

### LINKED ISSUE #735: Update login screen depending on SSO state
- The user should see a different login screen depending on if they are already logged in via the `profiles` service or not
- The user should be given contextual information about the new submission system in both cases
- If the user is logged in, they should see the 'Continue' button [see here](https://projects.invisionapp.com/share/NEOA2G14KDH#/screens/322120296)
- If the user is not logged in, they should see the 'Login with ORCiD' button [see here](https://projects.invisionapp.com/share/NEOA2G14KDH#/screens/322127272)

tamlyn:
#### Background

Closes #735 

- Read JWT on client and respond to `new-session` claim.
   - If `false` this means the user was already logged in to Journal so show interstitial with 'Continue' button
   - If `true` user has already been shown xPub log in screen so redirect straight to dashboard
- Update the login screen text and layout (took a few liberties with the second paragraph of text).

### DIFF
--- a/app/components/global/AppBar.js
+++ b/app/components/global/AppBar.js
@@ -2,10 +2,9 @@ import React from 'react'
 import { Link, withRouter } from 'react-router-dom'
 import { graphql, compose } from 'react-apollo'
 import { AppBar } from '@pubsweet/ui'
-import { withLoader } from 'pubsweet-client'
 import { CURRENT_USER } from './queries'
 
-const ElifeAppBar = ({ history, currentUser }) => (
+const ElifeAppBar = ({ history, data }) => (
   <AppBar
     brand={<img alt="eLife" src="/assets/elife-logo.png" />}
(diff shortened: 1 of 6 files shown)


tamlyn:
Yes, that part doesn't change. If the user is not logged in to Journal they are first send to the xPub login page where they click the green button and go back to Journal. It's from that point that these changes take effect.

hdrury1:
Hmm. I think a quick demo of what this PR does (tomorrow?) might be really useful. I've confused myself with the word 'interstitial' because it isn't really an interstitial, it's two different flavours of the same login screen.

tamlyn:
Yes, good idea. It is a bit confusing. The way I see it, the two versions of the same page are an implementation detail but you can think of them as two entirely separate pages (login page and interstitial) since the user should never see both in the same flow.

PR #767, opened 2018-10-09
Add the login flow to the documentation

diversemix:


diversemix:
Good idea - all for not confusing people!

PR #768, opened 2018-10-09
docs(PeoplePicker): update PeoplePickerModal.md with new search functionality

dryhten:
`PeoplePickerModal.md` is outdated. The Search Box functionality has changed and should be reflected in the docs.

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Editors/PeoplePickerModal.md
+++ b/app/components/pages/SubmissionWizard/steps/Editors/PeoplePickerModal.md
@@ -4,19 +4,13 @@ A people picker in a modal
 
 Upon opening the People Picker Modal, the search box is empty.
 
-Currently the user can only search by name.
+The user can search by name, affiliation or subject areas.
 
-Input generates a dropdown list of suggested matches by first letter of either first, middle or last name.
-
-The suggestion list is further refined with additional user input.
-
-Selecting a match from the dropdown populates the search box and updates the people pods returned.
-
-Alternatively, the user can elect to search for any full or partial name already entered without choosing from the dropdown list. This returns all matching person pods.
-
-The dropdown list is keyboard-navigable.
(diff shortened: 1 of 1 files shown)


app/components/pages/SubmissionWizard/steps/Editors/PeoplePickerModal.md
```diff
@@ -4,20 +4,14 @@ A people picker in a modal
...
-Selecting a match from the dropdown populates the search box and updates the people pods returned.
-
-Alternatively, the user can elect to search for any full or partial name already entered without choosing from the dropdown list. This returns all matching person pods.
-
-The dropdown list is keyboard-navigable.
+To update the list of peron pods the user has to press enter or click on the search icon after typing something.
 
 Searching for an empty string returns all person pods.
 
+Currently the input won't generate a dropdown list of suggestions (but will at some point in the future).
```
diversemix: @dryhten When we tested this there were suggestions of name matches
@hdrury1 ... not sure there is a plan to do this in the future at the moment... I expect we will do if if someone asks?

dryhten: @diversemix what do you mean? After https://github.com/elifesciences/elife-xpub/pull/763 there is no list of suggestions. Also, check https://github.com/elifesciences/elife-xpub/issues/629#issuecomment-428088027

diversemix: Yes you are right just tested locally... sorry ..  probably confused from when I was testing before !

hdrury1: Original designs included an auto suggest drop down. But until we start getting real world feedback I remain unconvinced that we actually need one (Journal doesn’t currently have one). In any case the product recs are far from finalised so I’d rather remove what we have now (which isn’t an auto suggest anyway).

PR #783, opened 2018-10-11
fix: meca article.xml tweaks

### LINKED ISSUE #771: MECA Changes after testing: 2018-10-11
- [x] Change all attribute values' dashes (-) to underscores (_) in attribute values, a.k.a. 'article-type' => 'article_type', 'potential-senior-editors' => 'potential_senior_editors', etc. This is what our code expects for other attribute values and I'd like to keep it as consistent as possible. This is also why no 'Unresolved Users' sticky note was created for the test import -- our code was expecting underscores and got dashes instead.
- [x] Change the 'subj-group-type="keywords"' attribute to 'subj-group-type="subject_areas"'. This is mostly just for my own sanity, so I don't get keywords and subject areas confused again...
- [x] Change '<subject>cell-biology</subject>' to '<subject>cell biology</subject>' (replace dash with space). This is so the subject area will match the configured subject area on EJP, 'Cell biology'. This is pretty important, as I don't have any way besides direct matching to resolve subject areas in imports. You can ignore upper/lowercase, but everything else should be the same as in your configured 'Major Subject Area(s)' list. I should also note that I'm hard-coding the import script to look in List 2 (as opposed to List 1, 'Research Organism(s)'). So if you ever change which list your subject areas are stored in for some reason, importing subject areas from xPub will stop working. Please let me know if this will be an issue.
- [x] Currently, the code is only using the 'article-type-id' (an integer) for identifying the manuscript type. 'article-type' is therefore unnecessary. You can keep it if you want, but it will result in problems down the line if you ever reorder the tags. I would recommend removing it, and changing 'article-type-id' to 'article_type'.

> With the above changes, for example, the <article-categories> tag would look like this:
```
  <article-categories>
      <subj-group subj-group-type="article_type">
          <subject>13</subject>
      </subj-group>
      <subj-group subj-group-type="subject_areas">
          <subject>cell biology</subject>
      </subj-group>
  </article-categories>
```

tamlyn:
#### Background

Closes #771 

- I made the assumption that the major subject area text eJP matches against is the text label as displayed in the "Major Subject Area(s)" page of the submission form.
- Moved definition of subject areas into config to enable mapping from our internal ID to the external label.
- Swapped the ad-hoc XML testing for a snapshot test as this provides a nicer diff when things change

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -37,6 +37,27 @@ module.exports = {
     api: {
       url: 'https://api.elifesciences.org/',
     },
+    majorSubjectAreas: {
+      'biochemistry-chemical-biology': 'Biochemistry and Chemical Biology',
+      'cancer-biology': 'Cancer Biology',
+      'cell-biology': 'Cell Biology',
+      'chromosomes-gene-expression': 'Chromosomes and Gene Expression',
+      'computational-systems-biology': 'Computational and Systems Biology',
+      'developmental-biology-stem-cells':
+        'Developmental Biology and Stem Cells',
+      ecology: 'Ecology',
+      'epidemiology-global-health': 'Epidemiology and Global Health',
+      'evolutionary-biology': 'Evolutionary Biology',
+      'genetics-genomics': 'Genetics and Genomics',
+      'human-biology-medicine': 'Human Biology and Medicine',
+      'immunology-inflammation': 'Immunology and Inflammation',
+      'microbiology-infectious-disease': 'Microbiology and Infectious Disease',
+      neuroscience: 'Neuroscience',
+      'physics-living-systems': 'Physics of Living Systems',
+      'plant-biology': 'Plant Biology',
+      'structural-biology-molecular-biophysics':
+        'Structural Biology and Molecular Biophysics',
+    },
   },
   login: {
     // TODO swap this mock for the Journal endpoint when available
@@ -95,5 +116,6 @@ module.exports = {
     'validations',
     'fileUpload',
     'login',
+    'elife.majorSubjectAreas',
   ],
 }
--- a/server/meca/file-generators/__snapshots__/article.test.js.snap
+++ b/server/meca/file-generators/__snapshots__/article.test.js.snap
@@ -1,0 +1,193 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Article XML generator generates expected XML 1`] = `
+"<?xml version=\\"1.0\\" encoding=\\"UTF-8\\"?>
+<!DOCTYPE article SYSTEM \\"JATS-journalpublishing1.dtd\\">
+<article>
+  <front>
+    <journal-meta>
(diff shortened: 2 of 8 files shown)


diversemix:
Ok - definitely some weirdness with the tests - they work locally but not in the pipeline :(

tamlyn:
Fixed mock date to always use UTC.

config/default.js
```diff
@@ -37,6 +37,27 @@ module.exports = {
...
+      'genetics-genomics': 'Genetics and Genomics',
+      'human-biology-medicine': 'Human Biology and Medicine',
+      'immunology-inflammation': 'Immunology and Inflammation',
+      'microbiology-infectious-disease': 'Microbiology and Infectious Disease',
+      neuroscience: 'Neuroscience',
+      'physics-living-systems': 'Physics of Living Systems',
+      'plant-biology': 'Plant Biology',
+      'structural-biology-molecular-biophysics':
+        'Structural Biology and Molecular Biophysics',
+    },
```
damnhipster: Where did you fetch this list from?

server/meca/file-generators/article.js
```diff
@@ -20,15 +23,14 @@ function createXmlObject(manuscript, editorsById, affiliations, reviewerMap) {
...
-          {
-            '@subj-group-type': 'article-type-id',
+            '@subj-group-type': 'article_type',
             subject: articleTypeMap[manuscript.meta.articleType],
           },
           manuscript.meta.subjects.map(subject => ({
-            'subj-group': { '@subj-group-type': 'keywords', subject },
+            'subj-group': {
+              '@subj-group-type': 'subject_areas',
+              subject: subjectAreas[subject],
```
damnhipster: It may be worth throwing an error if the subject does not exist within `subjectAreas`. At the moment, it will set the subject to undefined, and we would rely on EJP to return an error.

server/meca/file-generators/__snapshots__/article.test.js.snap
```diff
@@ -0,0 +1,193 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Article XML generator generates expected XML 1`] = `
```
damnhipster: Good use of snapshots, made this much easier to review.

config/default.js
tamlyn: It was previously in `SubjectAreaDropdown`. It should eventually come from the eLife API.

server/meca/file-generators/article.js
tamlyn: True. But it's a very unlikely situation since the same list is used in the drop down which populates the database in the first place.

server/meca/file-generators/__snapshots__/article.test.js.snap
tamlyn: Yeah. Finally found something snapshots are good for!

PR #784, opened 2018-10-11
Create Modal dialog component

### LINKED ISSUE #779: Modal dialog component
- Create the generic modal dialog component used in #511 and #635.
- Content area
- "Accept" and "Cancel" buttons with configurable text
- Browser back is equivalent to clicking "Cancel"
- Use existing 'ModalOverlay' component if possible

jsms90:
#### What does this PR do?
- Add prop to `ModalOverlay` for specifying whether background is "transparent" or not i.e. whether backgroundColour should be white (as in PeoplePicker's full-screen modal) or black with opacity 0.8 (as with small modals)
- Use `ModalOverlay` with transparent background to build new `ModalDialog` component i.e.
  - a white box in the middle of the viewport
  - 24px margin round the outside
  - text in the top half
  - 2 buttons at the bottom - "cancel" on the left & "accept" on the right

#### Any relevant tickets
fixes #779


diversemix:
FYI: I'd like to take Cory through reviewing this - looks like a good example

Issue #736, opened 2018-10-08
Create a 'Thank You' page to finish wizard


hdrury1:
- After confirming their submission via the modal, the user should be redirected to a page which explains that their submission has been received
- This page should set expectations as to when that decision is likely to be made
- This page should replace the need to email the user confirming receipt
- Please expect some tweaks to wording in the future
- Clicking 'Finish' should take the user to their dashboard, where the latest submission will appear at the top of their list

[Invision prototype](https://projects.invisionapp.com/share/NEOA2G14KDH#/screens/322129105)

PR #788, opened 2018-10-12
Create a 'Thank You' page to finish submission

Cooryd:
#### Background
Thank You page required after completing wizard form with button returning user to dashboard and text requesting they bookmark dashboard page.

#### What does this PR do?
Adds a new route of /thankyou/:id to the app and a component for displaying the required information.

#### Any relevant tickets
 #736 
Locally

### DIFF
--- a/app/components/pages/ThankYou/ThankYou.js
+++ b/app/components/pages/ThankYou/ThankYou.js
@@ -1,0 +1,34 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+import ExternalLink from '../../ui/atoms/ExternalLink'
+import ButtonLink from '../../ui/atoms/ButtonLink'
+
+const CenteredContent = styled(Box)`
+  text-align: center;
+`
+const SubText = styled.p`
+  color: ${th('colorTextSecondary')};
+  font-size: ${th('fontSizeBaseSmall')};
+`
+const BookmarkLink = styled(ExternalLink)`
+  display: inline-block;
+`
+
+const ThankYou = ({ title }) => (
+  <CenteredContent mx="auto" width={[1, 600]}>
+    <h1>Thank you</h1>
+    <p>{`Your submission, "${title}" has been received.`}</p>
+    <p>You will be informed of a decision soon.</p>
+    <SubText>
+      You may want to bookmark the link below to check the progress of your
+      submission.
+      <BookmarkLink href="/">elifesciences.org/submissions</BookmarkLink>
+    </SubText>
+    <ButtonLink primary to="/">
+      Finish
+    </ButtonLink>
+  </CenteredContent>
+)
+export default ThankYou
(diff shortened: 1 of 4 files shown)


app/components/pages/ThankYou/index.js
```diff
@@ -0,0 +1,31 @@
...
+`
+const ThankYouPage = ({ match }) => (
+  <Query query={MANUSCRIPT_TITLE} variables={{ id: match.params.id }}>
+    {({ data, loading, error }) => {
+      if (loading) {
+        return <div>Loading...</div>
+      }
+
+      if (error) {
+        return <div>{error.message}</div>
```
jsms90: Perhaps we could use `ErrorPage` in here? @tamlyn @diversemix what do you think?

app/components/pages/ThankYou/ThankYou.js
```diff
@@ -0,0 +1,47 @@
...
+
+const CenteredContent = styled(Box)`
+  text-align: center;
+`
+const SubText = styled(SmallParagraph)`
+  color: ${th('colorTextSecondary')};
+`
+
+const BookmarkLink = styled(ExternalLink)`
+  display: inline-block;
```
jsms90: Just an FYI, it looks like this on firefox
image

app/components/pages/ThankYou/ThankYou.js
```diff
@@ -0,0 +1,47 @@
...
+      <BookmarkLink href="/">elifesciences.org/submissions</BookmarkLink>
+    </SubText>
+    <ButtonLink primary to="/">
+      Finish
+    </ButtonLink>
+  </CenteredContent>
+)
+
+ThankYou.propTypes = {
+  title: PropTypes.string,
```
jsms90: Might it be more accurate to make this required, rather than adding "Untitled manuscript" as a default?

If I start submitting a manuscript & go to http://localhost:3000/thankyou/3b9b39a0-08c9-430d-8c4c-cd94aa722f38, then I see this:
image

(i.e. an empty string, rather than saying 'untitled...')

'Untitled manuscript' makes sense on the dashboard, if they've only partially completed a submission. But a user shouldn't be able to get to this page without having entered a title, right?

Cooryd: This should be the very last page they see after completing the wizards steps, and as the title fields mandatory you're correct they should never get to this page without a title. I'll make it required.

app/components/pages/ThankYou/index.js
Cooryd: This looks good with the `ErrorPage` component, I'll make this gave and re-commit.

app/components/pages/ThankYou/ThankYou.js
```diff
@@ -0,0 +1,47 @@
...
+`
+
+const BookmarkLink = styled(ExternalLink)`
+  display: inline-block;
+`
+
+const ThankYou = ({ title }) => (
+  <CenteredContent mx="auto" width={[1, 600]}>
+    <H1>Thank you</H1>
+    <Paragraph>{`Your submission, "${title}" has been received.`}</Paragraph>
```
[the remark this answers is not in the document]
Cooryd: While it works in the browser, that was causing the linter to throw an no-unescaped-entities error, the template string was my fix to get around this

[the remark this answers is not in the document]
diversemix: Maybe? 
`<Paragraph>Your submission, &quot;{title}&quot; has been received.</Paragraph>`

PR #789, opened 2018-10-12
feat: App bar (desktop view)

### LINKED ISSUE #777: App bar component (desktop view)
Product requirements

- The links for the reviewer guide, author guide and contact us pages should appear even when the user is not logged in
- The active link is colorText
- Inactive links are colorBorder
- If not logged in, the user should have a 'Login' link on the right-hand side (colorPrimary)

app bar

- If logged in, the user should have a user icon (24px x 24px) and dropdown (4px to the right of the icon)

app bar

tamlyn:
#### Background

- Closes #777 
- Create a LogoutPage component to encapsulate the steps necessary to log out and clear the apollo cache
- Remove explicit webpack node_modules resolution as it was causing the wrong version of a package to be included in the build and breaking React Router's NavLink


diversemix:
Why the removal of the paths in `webpack/common-resolve.js` ? @tamlyn

tamlyn:
> Remove explicit webpack node_modules resolution as it was causing the wrong version of a package to be included in the build and breaking React Router's NavLink

It has been like that in PubSweet since forever. As far as I can tell, the reason it was set in the first place was to support the very suspect practice of having PubSweet components import files from the containing app. That was removed over a year ago making that bit of the config redundant but it hadn't caused any problems until now so no one removed it.

client/elife-theme/src/icons/material/Profile.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+
+const Profile = props => (
+  <svg height="24px" viewBox="0 0 24 24" width="24px" {...props}>
+    <path d="M12 5.9a2.1 2.1 0 1 1 0 4.2 2.1 2.1 0 0 1 0-4.2m0 9c2.97 0 6.1 1.46 6.1 2.1v1.1H5.9V17c0-.64 3.13-2.1 6.1-2.1M12 4C9.79 4 8 5.79 8 8s1.79 4 4 4 4-1.79 4-4-1.79-4-4-4zm0 9c-2.67 0-8 1.34-8 4v3h16v-3c0-2.66-5.33-4-8-4z" />
+    <path d="M0 0h24v24H0z" fill="none" />
```
jsms90: I'm guessing we want fill to be white?

Designs:
image


Actual:
image

app/components/global/AppBar/ProfileMenu.js
```diff
@@ -0,0 +1,49 @@
...
+  iconName: 'User',
+  overrideName: '@pubsweet-pending.AppBar.Profile',
+})`
+  display: block;
+  fill: ${th('colorTextReverse')};
+  background-color: ${th('colorText')};
+  border-radius: ${th('borderRadius')};
+`
+
+const IconArrow = styled.div`
```
jsms90: Might it be worth moving this into `atoms`?

Since this icon isn't created in the same way as other icons and therefore doesn't live with the others, popping it into `atoms` might make it more visible for future devs. And they'll prob want to use it if/when they make the other menus in our application conform to designs

client/elife-theme/src/icons/material/Profile.js
tamlyn: Yeah, my (edited) [question on the ticket](https://github.com/elifesciences/elife-xpub/issues/777#issuecomment-429310826) got missed. Have created new issue #795.

Issue #774, opened 2018-10-11
DashboardListItem component


tamlyn:
![image](https://user-images.githubusercontent.com/115310/46796470-b69cce00-cd44-11e8-9fd2-fed3ca0d510e.png)

- Show relative date based on last status change date
- Show status text and colour based on client side status code (to be defined)

PR #794, opened 2018-10-15
774 dashboard list item

jyenes:
#### Background
this Creates a New Component that will be used on the Dashboard, to show the items submitted.

#### Any relevant tickets
fixes: #774


app/components/ui/molecules/DashboardListItem.js
```diff
@@ -0,0 +1,89 @@
...
+  ({
+    1: 'Waiting for decision',
+    2: 'Rejected',
+    3: 'Approved',
+  }[status])
+
+const dashboardDateText = date => {
+  const diffDays = differenceInCalendarDays(new Date(), date)
+  if (diffDays < 0) {
+    return 'Invalid date'
```
tamlyn: Would it make more sense to say "Today" for cases like this which could occur if the user's computer time is slightly out?

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+  if (diffDays === 0) {
+    return 'Today'
+  }
+  if (diffDays === 1) {
+    return 'Yesterday'
+  }
+  return `${diffDays} days ago`
```
tamlyn: I believe this is meant to switch to "1 month ago" etc. at some point. Not sure where I saw that.

app/components/ui/molecules/DashboardListItem.md
```diff
@@ -0,0 +1,11 @@
+TODO
```
diversemix: What is ToDo?

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -0,0 +1,89 @@
...
+import { differenceInCalendarDays, format } from 'date-fns'
+import { th } from '@pubsweet/ui-toolkit'
+import PropTypes from 'prop-types'
+
+const dashboardStatusColor = (status, theme) =>
+  ({
+    1: theme.colorTextSecondary,
+    2: theme.colorError,
+    3: theme.colorSuccess,
+  }[status])
```
tamlyn: Rather than returning the actual colour from this function, you can return the colour name (e.g. `colorError`). Then you can use that in `StatusBox` to index into the theme. That way you don't have to wrap the component in `withTheme`.

dryhten:
Just leaving this here: we will also need to wrap this component in a Link one that points to the submission. We can do it at this level by accepting a manuscript id/link prop or at the component where we integrate everything together.

app/components/ui/molecules/DashboardListItem.js
dryhten: This is what I remember as well. It would be nice if we find that invision page to link it in the github issue. Was is just months and then years ago?

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -0,0 +1,89 @@
...
+  return `${diffDays} days ago`
+}
+
+const Root = styled(Flex)`
+  height: ${th('space.5')}
+  align-items: center;
+`
+const StatusBox = styled(Box)`
+  color: ${props => props.color};
+  font-size: small;
```
jsms90: Right now, this seems to compute the font-size as `13px` so everything is slightly smaller than the designs.

You might be better off separating the layout (`Box`) and leave font size, colour, etc to another element (e.g. `Paragraph`)?

How about:
```js
<Box>
  <Paragraph />
</Box>

<Box>
  <Paragraph />
</Box>

<Box>
  <Paragraph />
</Box>
```

Then it should have the right font sizing

app/components/ui/molecules/DashboardListItem.js
dryhten: As far as I know, `new Date()` should be in UTC and as long as we store the date in UTC as well this should not happen unless there is a problem somewhere else in the app. I was thinking that if we show `Invalid date` to the user he might report it so we know there is a problem. If we show `Today` we might never know there is a problem, but I understand that it's nicer to show this to the user.

app/components/ui/molecules/DashboardListItem.js
chugginselifesciences: Ideally it would be one of:
"Today"
"Yesterday"
"x days ago"
"x months ago" 

I know this could get complex calculating no. of days per month. I think rounding down to nearest multiple of 30 days would suffice for now for "x months ago", unless there's a library or similar that can do it more accurately without lots of work. 

We have the actual date below so it doesn't have to be 100% accurate.

app/components/ui/molecules/DashboardListItem.js
diversemix: I think Tam was considering more of a problem with clock drift  (see http://www.faqs.org/rfcs/rfc958.html )  rather than timezones.

app/components/ui/molecules/DashboardListItem.js
jsms90: Yeah we can always change it later if @chugginselifesciences wants :+1: But I think we can safely say that it's not going to be 13px, right? I think it's a good approach to go by eye and use our best guess. But we can still take guidance from the styleguide and make this component conform to one of the things that exist there, right? i.e. 16px (fontSizeBase) or 14px (fontSizeBaseSmall), rather than introducing a new size now.

chugginselifesciences: I'll be updating the style guide in InVision a.s.a.p. so hopefully you'll have more info today. But the specific font-size of the status text should be 16px, and should wrap to a 2nd line if needed. However we have changed the status wording following the meeting last week.

In progress submissions will be "In progress" and once submitted will simply stay as "submitted" in an inactive state.

app/components/ui/molecules/DashboardListItem.js
dryhten: Sorry, how is theme added to props if I don't wrap it in `withTheme`?

jyenes:
@jsms90 @tamlyn @dryhten @diversemix  can you review it again, please? changes for design are now in place

app/components/ui/atoms/ManuscriptStatus.js
```diff
@@ -6,12 +6,14 @@ import PropTypes from 'prop-types'
 const StatusBox = styled(Box)`
   color: ${props => props.theme[props.color]};
   font-size: 16px;
+  padding-left: 24px;
+  flex: 0 0 120px;
```
tamlyn: FYI we try to use values from the theme whenever possible. In this case `fontSizeBase`, `space.3` and `space.7`.

PR #797, opened 2018-10-15
Wizard final step confirmation modal

### LINKED ISSUE #511: Create modal to end wizard
**User Story:**

_As a submitter I want the option to edit my submission data before it is sent for assessment._

Functional:

- Modal should appear after the user hits 'Submit' on the final screen of the submission wizard.

- 'Back' should close the modal and return to last screen of the wizard. The user should then be able to navigate forward and back through the wizard to make any changes as before.

- 'Continue' should redirect to the Thank You page.

- Using the browser's back button should have the same effect as hitting 'Back' on the modal.

**Designs*

image

- After confirming their submission via the modal, the user should be redirected to a page which explains that their submission has been received. see #736

tamlyn:
#### Background

Apologies for the big PR. One this led to another and here we are. Happy to break it into smaller pieces but they will need to be fed in sequentially as it's all a bit interlinked.

- Add confirmation dialog at end of the wizard, closes #511 
  - Create a dummy submit button on final step of wizard
  - Move real submit button into a ModalDialog
- Add size prop to ModalDialog as it is not obvious how to size automatically based on the content
- Create helper component to save modal state in browser history
  - Update Editors step to use this component
- Update Button min-width width from new style guide
- Fix bug where auto save on final step of wizard caused a submission
- Re-enable flaky SFTP content end to end test with more logging
- ~~Add check to prevent updating a submitted manuscript~~

Tests updated, some new tests.

### DIFF
--- a/app/components/pages/SubmissionWizard/WizardStep.js
+++ b/app/components/pages/SubmissionWizard/WizardStep.js
@@ -7,6 +7,7 @@ import ButtonLink from '../../ui/atoms/ButtonLink'
 import { FormH2 } from '../../ui/atoms/FormHeadings'
 import AutoSave from './AutoSave'
 import ProgressBar from './ProgressBar'
+import WizardSubmit from './WizardSubmit'
 
 const BoxNoMinWidth = styled(Box)`
   min-width: 0;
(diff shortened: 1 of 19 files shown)


tamlyn:
Note E2E tests caught a bug here to do with the timing of the final AutoSave. Will investigate.

Issue #776, opened 2018-10-11
Map manuscript statuses


tamlyn:
- Users should not see all server manuscript statuses
- Create a map of server manuscript status codes to a (smaller set of) client visible status codes
- Return client status code in GraphQL response
- Create a (client side) map of client visible status codes to display text and display colour

Map contents to be defined.

PR #799, opened 2018-10-16
feat(Dashboard): creating an atom component to handle the manuscript …

jyenes:
creating an atom component to handle the manuscript status, besides handle and mapping those statuses in the server side, to display the correct info.

fixes: #776


server/xpub/entities/manuscript/resolvers.js
```diff
@@ -237,6 +237,22 @@ ${err}`,
       return team.teamMembers[0].alias
     },
 
+    async clientStatus(manuscript) {
+      let clientStatus = 'REJECTED'
```
tamlyn: Probably shouldn't default to `REJECTED`. Maybe throw an error instead?

server/xpub/entities/manuscript/resolvers.js
```diff
@@ -237,6 +237,22 @@ ${err}`,
...
+      } else if (
+        manuscript.status === 'MECA_EXPORT_PENDING' ||
+        manuscript.status === 'MECA_EXPORT_FAILED' ||
+        manuscript.status === 'MECA_EXPORT_SUCCEEDED' ||
+        manuscript.status === 'MECA_IMPORT_FAILED' ||
+        manuscript.status === 'MECA_IMPORT_SUCCEEDED'
+      ) {
+        clientStatus = 'WAITING_FOR_DECISION'
+      }
+      return clientStatus
```
tamlyn: Maybe a `switch` would be a neater way to do this?

app/components/ui/atoms/ManuscriptStatus.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box } from 'grid-styled'
+import PropTypes from 'prop-types'
+
+const StatusBox = styled(Box)`
+  color: ${props => props.theme[props.color]};
+  font-size: 13px;
```
tamlyn: Is `13px` right? The designs seem to use `16px` aka `fontSizeBase`.

app/components/pages/Dashboard/index.js
```diff
@@ -12,6 +12,7 @@ export const MANUSCRIPTS = gql`
         title
       }
       status
+      clientStatus
```
tamlyn: Probably don't need `status` any more.

app/components/pages/SubmissionWizard/operations.js
```diff
@@ -19,6 +19,7 @@ const manuscriptFragment = gql`
   fragment WholeManuscript on Manuscript {
     id
     status
+    clientStatus
```
tamlyn: Don't think we need `status` here either.

tamlyn:
Tests are failing due to the `clientStatus` being sent back to the server and not recognised. Need to add it to the `omitList` here: https://github.com/elifesciences/elife-xpub/blob/develop/app/components/pages/SubmissionWizard/SubmissionOperations.js#L21

PR #802, opened 2018-10-16
Disable test coverage

damnhipster:
We're not using the test coverage as a statistic, and it only serves to make the test output more verbose.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,11 +57,6 @@
     ]
   },
   "jest": {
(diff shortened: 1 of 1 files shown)


jyenes:
so as discussed, create an issue to re-enable it later on and we can disable it right now

diversemix:
@jyenes ... can you link to the issue to re-enable in here?

damnhipster:
#806

Issue #782, opened 2018-10-11
Sticky footer component


jsms90:
We have a repeated pattern of a white "footer" that sticks to the bottom of the page, which has a button / multiple buttons inside.

This has already been implemented once in the People Picker modal (with buttons on the left):
![image](https://user-images.githubusercontent.com/15656538/46808507-6df50d00-cd64-11e8-82a9-20b374055a8c.png)

We now have another one in issue #781 (with button on the right):
![screen shot 2018-10-11 at 13 10 03](https://user-images.githubusercontent.com/17023702/46803022-fec4ec00-cd56-11e8-827d-af4ebbe0b2c0.png)

We also have designs for a future feature (multi-selection in the dashboard) with a delete button in the footer (right hand side):
![image](https://user-images.githubusercontent.com/15656538/46808783-055a6000-cd65-11e8-9d48-4cb4ede62390.png)

@chugginselifesciences Should the height of the dashboard footer be the same as whatever we chose for the People Picker modal last time?

chugginselifesciences:
Yes the intention is that the same footer would be used for multi-select, and anywhere else we may use it in future, which should be 132px tall including 1px border

"On mobile" (or before the breakpoint where we go from 1 to 2 columns) the plan would be for the fixed footer to be 73px tall including the 1px border.

I'll test the people picker modal when staging is up and running again.

chugginselifesciences:
Good point as long as it's referred to. I'll speak to Hannah about it in the morning. There's probably a sensible way to keep adding to it and linking to the tickets or adding more detail if needed etc.

PR #805, opened 2018-10-16
Create StickyFooter component & use in PeoplePicker & Dashboard (submission button in mobile view)

### LINKED ISSUE #781: Add fixed new submission button to bottom of mobile dashboard
screen shot 2018-10-11 at 13 10 03

- Bar should be fixed to bottom of screen
- Dashboard scrolls behind it

[Invision prototype](https://projects.invisionapp.com/share/FDOIB61EKV5#/screens/322899361)

jsms90:
See [Slack thread](https://elifesciences.slack.com/conversation/C6LKMFK5Z/p1539615822000100) for latest discussion on specific CSS values

#### What does this PR do?
- Update `Button` component's CSS to conform to styleguide
- Add sticky footer with 'New Submission' button (and update text of desktop's submission button)
- Hide/show correct 'New Submission' button based on screen size

#### Any relevant tickets
fixes #781 #782 

Visually


app/components/global/layout/DisplayBasedOnScreenSize.js
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components'
+
+export const HideOnMobile = styled.div`
+  @media (max-width: 480px) {
+    display: none;
+  }
+`
+
+export const ShowOnMobile = styled.div`
+  @media (min-width: 480px) {
```
tamlyn: I think the way matching works, at exactly `480px` neither of these will be displayed.

app/components/global/layout/DisplayBasedOnScreenSize.js
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components'
+
+export const HideOnMobile = styled.div`
+  @media (max-width: 480px) {
```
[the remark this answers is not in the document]
tamlyn: Did you copy and paste `${th('breakpoints.0)}` because if so, you're missing a `'`.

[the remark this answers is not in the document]
Cooryd: It's odd as the media query is correctly showing in the inspector. The browser is just ignoring it 🤔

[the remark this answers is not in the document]
jsms90: @tamlyn haha, no I didn't

@Cooryd Do you mean if we put in 'breakpoints.0', it's showing in the inspector? How weird

app/components/global/layout/DisplayBasedOnScreenSize.js
jsms90: Oh! Thanks :+1: 
Will fix tomorrow

app/components/global/layout/DisplayBasedOnScreenSize.js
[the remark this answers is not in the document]
Cooryd: Yes, I can see the 480px within the browser inspector so it seems to be generating the class correctly but the browser isn’t rendering the query like you said.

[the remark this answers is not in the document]
tamlyn: Oh. The breakpoint is `480` without units. Need to add `px`.
```suggestion
  @media (max-width: ${th('breakpoints.0')}px) {
```

(Trying out new GitHub code suggestion thingie ^^^)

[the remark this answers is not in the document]
jsms90: Ooo I like the suggestion thingy. But I'd like to make a commit that changes both '480px' to 'breakpoints' at once :thinking:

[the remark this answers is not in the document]
Cooryd: https://www.styled-components.com/docs/advanced#media-templates details using a a helper function for media queries which might be useful. There example fixes it as a max-width though, theres also [styled-components-breakpoint](https://github.com/wearehumblebee/styled-components-breakpoint) which takes this an adds more functionality so you can set breakpoint ranges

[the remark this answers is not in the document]
jsms90: Ah that's a great shout. Thanks @Cooryd!

I originally had the media queries directly on the `Box` & `StickyFooter` components that I want to hide. But then extracted them into generic wrapper components, so that we could reuse these queries elsewhere. But adding an extra, unnecessary bit of HTML to the page is definitely not ideal. Creating a helper function is much better. Cheers! :+1:

app/components/global/layout/DisplayBasedOnScreenSize.js
jsms90: @tamlyn What would you suggest?

I'm thinking that maybe it would be best to only use `max-width` inside a helper & then taking advantage of precedence like this:

the top submission button (desktop only) would have:
```js
${media.mobile`display: none;`}
```
but then the submission button in the footer (mobile only) would have:
```js
${media.desktop`display: none;`}
${media.desktop`display:block;`}
```

Do you guys reckon that's a good approach?

jsms90: To keep things in line with [how `grid-styled` uses breakpoints](https://github.com/rebassjs/grid#responsive-styles), I'm using `min-width` instead of `max-width`:
> Rebass Grid props accept arrays as values for mobile-first responsive styles, where the first value is for all breakpoints, then each value after is for a min-width media query from that breakpoint and up.

jsms90:
@tamlyn @Cooryd ready for a re-review whenever you have time :smile:

app/components/global/layout/media.js
```diff
@@ -0,0 +1,20 @@
+import { css } from 'styled-components'
+import theme from '@elifesciences/elife-theme'
```
tamlyn: Note that this breaks the assumption that the theme is configured via the themeprovider only. The way this helper is being used so far is always within a styled component, so it could return a function which then reads the theme off the `props`.

app/components/pages/AuthorGuide/TheEditorialProcess.js
```diff
@@ -0,0 +1,21 @@
...
+      selective journal that aims to publish work of the highest scientific
+      standards and importance. Leading academic researchers evaluate new
+      submissions and approximately two-thirds are returned to the authors
+      without further peer review. Approximately half of the articles that are
+      selected for peer review go on to be published.
+    </Paragraph>
+  </div>
+)
+
+export default AuthorGuide
```
tamlyn: Did you mean to include this in this PR?

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -4,12 +4,23 @@ import { Link } from 'react-router-dom'
...
 import Paragraph from '../../ui/atoms/Paragraph'
 import ManuscriptStatus from '../../ui/atoms/ManuscriptStatus'
+import StickyFooter from '../../ui/atoms/StickyFooter'
 
 /* Temporary dashboard view pending receipt of designs */
 
+const StyledBox = styled(Box)`
+  display: none;
+  ${media.mobile`display:block;`};
+`
```
tamlyn: The name `StyledBox` is quite meaningless. Maybe `HideFromMobile`?

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -4,12 +4,23 @@ import { Link } from 'react-router-dom'
...
+
+const StyledStickyFooter = styled(StickyFooter)`
+  ${media.mobile`display: none;`};
+`
```
tamlyn: Again, `MobileOnlyStickyFooter` would communicate more.

app/components/global/layout/media.js
```diff
@@ -0,0 +1,20 @@
...
+const media = Object.keys(sizes).reduce((acc, label) => {
+  acc[label] = (...args) => css`
+    @media (min-width: ${sizes[label]}px) {
+      ${css(...args)};
+    }
+  `
+  return acc
+}, {})
+
+export default media
```
tamlyn: It's a bit unexpected that `media.mobile<style>` actually applies that style to everything *but* mobile.

app/components/global/layout/media.js
```diff
@@ -0,0 +1,20 @@
...
+}
+
+const media = Object.keys(sizes).reduce((acc, label) => {
+  acc[label] = (...args) => css`
+    @media (min-width: ${sizes[label]}px) {
+      ${css(...args)};
+    }
+  `
+  return acc
+}, {})
```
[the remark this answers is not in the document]
jsms90: I'm unsure.

I realise that adding in this mapping will mean changing the code in 2 places, if/when someone adds another breakpoint. That's never ideal.

At the same time, it dramatically improves readability (in my opinion) to say `media.mobile` vs `media[0]`. One of my issues with `grid-styled` (and I have many :sweat_smile:) is that I'm forever going back to the theme and looking up which pixel value "`space.3`" refers to :laughing: 

Happy to change it you feel it's worth it though

app/components/pages/AuthorGuide/TheEditorialProcess.js
jsms90: Oops! Removed :+1:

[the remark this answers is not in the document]
jsms90: Oops! Removed :+1:

app/components/pages/Dashboard/Dashboard.js
jsms90: Done :+1:

app/components/pages/Dashboard/Dashboard.js
jsms90: Done :+1:

PR #809, opened 2018-10-16
Exposing public config at runtime

### LINKED ISSUE #787: Login link fails on staging
The login URL on staging and production is incorrect. This is because the URL is configured at build time, so it's not picking up what environment it's on.

This should be changed to a solution where the URL can be configured at run time by the server.

- [x] Load a config file onto the page using `<script src="/config.js" />`, which sets config on the `window`
- [ ] Make a change in pubsweet to allow server-side templating within html responses, and embed the config on the page within `<script></script>` tags.

**Steps to reproduce**

Go to https://staging--xpub.elifesciences.org/login
Click login
Doesn't log you in

damnhipster:
#### Background

- Creates a route on the server to fetch JS to load a `window.config` object on the browser. It enables us to use environment specific config, rather than compiling separate bundles per environment.
- Suppressing log messages on a resolver test

#### Any relevant tickets

Closes #787 

- Locally by setting `NODE_CONFIG_ENV` to staging
- Unit and e2e tests

### DIFF
--- a/app/components/pages/Login/index.js
+++ b/app/components/pages/Login/index.js
@@ -3,14 +3,13 @@ import { Flex, Box } from 'grid-styled'
 import { Redirect } from 'react-router-dom'
 import { withApollo } from 'react-apollo'
 import gql from 'graphql-tag'
-import config from 'config'
 import { H2 } from '@pubsweet/ui'
 import OrcidButton from '../../ui/atoms/OrcidButton'
 import ButtonLink from '../../ui/atoms/ButtonLink'
 import Paragraph from '../../ui/atoms/Paragraph'
 import ExternalLink from '../../ui/atoms/ExternalLink'
 
-const { url: loginUrl, signupUrl, legacySubmissionUrl } = config.login
+const { url: loginUrl, signupUrl, legacySubmissionUrl } = window.config.login
 
 const EXCHANGE_TOKEN_MUTATION = gql`
   mutation($token: String) {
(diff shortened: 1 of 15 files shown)


damnhipster:
I'm having issues changing the `publicKeys` array to `isPublic` properties on relevant objects due to pubsweet, so that will be done in a separate PR.

app/components/pages/Login/index.js
```diff
@@ -3,14 +3,13 @@ import { Flex, Box } from 'grid-styled'
...
 import gql from 'graphql-tag'
-import config from 'config'
 import { H2 } from '@pubsweet/ui'
 import OrcidButton from '../../ui/atoms/OrcidButton'
 import ButtonLink from '../../ui/atoms/ButtonLink'
 import Paragraph from '../../ui/atoms/Paragraph'
 import ExternalLink from '../../ui/atoms/ExternalLink'
 
-const { url: loginUrl, signupUrl, legacySubmissionUrl } = config.login
+const { url: loginUrl, signupUrl, legacySubmissionUrl } = window.config.login
```
tamlyn: How about using [Webpack externals](https://webpack.js.org/configuration/externals/) to avoid this change? If not, you'll need to update the other places where `config` is used in the front-end (and it would break theoretical compatibility with PubSweet components).

damnhipster: Should we use Webpack externals to replace `common-resolve.js`?

tamlyn: Delete! Delete!

PR #811, opened 2018-10-17
734 - Put together Dashboard Page

### LINKED ISSUE #734: Dashboard component
- Assemble from Tabs and DashboardListItem components
  - List items ordered by status update time descending
- The 'Archive' tab will be clickable but show only a static message (to be defined)
- The active tab should be indicated by an underline (colorPrimary) and colorText
- Before the user has begun any new submissions, the dashboard should be presented in an empty state
- The user should be able to access eJP from their xPub dashboard in case they cannot find their submission(s)
  - This message is pushed down by a container with min-height
- Each entry in the dashboard should show its current status, title and date status was last updated
- Move new submission button to the right
- List items may be clickable (depending on status) 
  - "Continue submission" links to the wizard
  - "Invited to submission" links to eJP
  - Other statuses not clickable
- A submission should appear on this dashboard with this status from the moment a user has clicked 'New Submission'

**Designs**

[Invision Prototype](https://invis.io/NEOA2G14KDH)

image

image

Cooryd:
#### Background
Dashboard page needs compiling from the new Tabs and DashboardListItem component.

What does this PR do?

- Replaces the temporary dashboard page with a styled dashboard using the Tabs and DashboardListItem component.
- Adds a flex-grow styling to the DashboardListItem component to take up full list width when manuscript title is short.
- Adds a sortBy to the query ran fetching all manuscripts 

#### Any relevant tickets
Closes #734 

locally


app/components/pages/Dashboard/Dashboard.js
```diff
@@ -1,90 +1,78 @@
...
 import StickyFooter from '../../ui/atoms/StickyFooter'
+import Tabs from '../../ui/molecules/Tabs'
+import DashboardList from '../../ui/molecules/DashboardList'
+import { ALL_MANUSCRIPTS } from './oporations'
+import ManuscriptsQuery from './ManuscriptsQuery'
 
-/* Temporary dashboard view pending receipt of designs */
-
-const BoxHiddenFromMobile = styled(Box)`
+const RightAlignedButton = styled(Box)`
```
jsms90: This naming kinda makes it sound like styling has been applied to an actual `<Button>`, whereas what you are renaming is a `Box` (a `div` under the hood).

In this specific situation, your div's content happens to be a button :+1: But in another situation, you might put something completely different inside. This Box doesn't care what it's contents are. So long as you want to right-align the contents & hide it from mobile, you want to use this same Box. So I reckon the naming needs to reflect those 2 aspects of what this Box does, rather than what goes inside it. Does that make sense?

So what do you choose that's going to encapsulate those 2 pieces of functionality about this Box, i.e. those 2 additional bits of CSS. Well...`RightAlignedBoxHiddenFromMobile` is really long. And the minute you add another bit of CSS that makes it blue, it becomes `BlueRightAlignedBoxHiddenFromMobile`. And so on & so on & so on.

So, whilst I understand [the desire for more declarative naming](https://github.com/elifesciences/elife-xpub/pull/805#discussion_r225962686), I think that, in general, this is why you often see `StyledBox` or `StyledHeader`, `StyledParagraph`, etc - it's more extensible.

`StyledXYZ` usually seems the most appropriate to me, until it's such a common pattern throughout your application that it needs to be its own `atom`.

Feel free to disagree, of course.

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -1,90 +1,78 @@
...
+
 const MobileOnlyStickyFooter = styled(StickyFooter)`
   display: block;
   ${media.mobileUp`display: none;`};
 `
 
-const Header = styled.th`
-  text-align: left;
-  padding: 3px 6px;
+const EmptyListMessage = styled.div`
```
jsms90: Could we make this a styled `p` (or `Paragraph`) just for semantics? Off the top of my head, I can't see a downside, but perhaps there is one?

jsms90: Ah, just seen `CentredSmallParagraph` below. Could we use this again & perhaps this `div` around it could be responsible only for adding the `margin-top` value?

Better yet, use a grid-styled element? :smile:

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -1,90 +1,78 @@
...
+          New Submission
+        </Button>
+      </RightAlignedButton>
+      <Tabs>
+        <Tabs.List>
+          <Tabs.Tab>Submissions</Tabs.Tab>
+          <Tabs.Tab>Archive</Tabs.Tab>
+        </Tabs.List>
+        <DashboardPanel data-test-id="manuscripts">
+          <ManuscriptsQuery query={ALL_MANUSCRIPTS}>
```
jsms90: I like this change. Having ManuscriptsQuery & operations in separate files makes it much more readable, imo.

But this is a small departure from our prior decision to use the top-level page component (`index.js`) to handle all graphql queries and mutations. Without fully understanding why that decision was made, I would probably highlight this with @tamlyn when he gets back.

It might be worth discussing appropriate patterns for organising graphql on the frontend? LGTM

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -1,90 +1,78 @@
...
 import media from '../../global/layout/media'
-import { FormH2, FormH3 } from '../../ui/atoms/FormHeadings'
-import Paragraph from '../../ui/atoms/Paragraph'
-import ManuscriptStatus from '../../ui/atoms/ManuscriptStatus'
+import SmallParagraph from '../../ui/atoms/SmallParagraph'
+import ExternalLink from '../../ui/atoms/ExternalLink'
 import StickyFooter from '../../ui/atoms/StickyFooter'
+import Tabs from '../../ui/molecules/Tabs'
+import DashboardList from '../../ui/molecules/DashboardList'
+import { ALL_MANUSCRIPTS } from './oporations'
```
jsms90: typo in the file name: 'op**e**rations'

app/components/pages/Dashboard/ManuscriptsQuery.js
```diff
@@ -0,0 +1,22 @@
...
+
+const ManuscriptsQuery = ({ query, children }) => (
+  <Query fetchPolicy="cache-and-network" query={query}>
+    {({ data, loading, error }) => {
+      if (loading && !data.manuscripts) {
+        return <div>Loading...</div>
+      }
+
+      if (error) {
+        return <div>{error.message}</div>
```
jsms90: At some point we might want to design a view for this, @chugginselifesciences ?

app/components/pages/SubmissionWizard/SubmissionOperations.js
```diff
@@ -3,7 +3,7 @@ import { graphql, compose } from 'react-apollo'
 import React from 'react'
 import omitDeep from 'omit-deep-lodash'
 import ErrorPage from '../Error'
-import { MANUSCRIPTS } from '../Dashboard'
+import { ALL_MANUSCRIPTS } from '../Dashboard/oporations'
```
jsms90: same typo in filename as above: 'op**e**rations'

server/client-config/routes.test.js
```diff
@@ -14,31 +14,26 @@ describe('Client config handler', () => {
     jest.mock('config', () => ({
       publicConfig: {
         isPublic: true,
-        key: 'value'
+        key: 'value',
```
jsms90: Is everyone's linting not set up in the same way?

app/components/ui/molecules/DashboardList.js
```diff
@@ -0,0 +1,46 @@
...
+import { Link } from 'react-router-dom'
+import { th } from '@pubsweet/ui-toolkit'
+import DashboardListItem from '../../ui/molecules/DashboardListItem'
+import SmallParagraph from '../../ui/atoms/SmallParagraph'
+
+const DashboardLink = styled(Link)`
+  text-decoration: none;
+  color: ${th('colorText')};
+`
+const EmptyListMessage = styled.div`
```
jsms90: Same comments about `div` vs `p` as above

app/components/ui/molecules/DashboardList.js
```diff
@@ -0,0 +1,46 @@
...
+  margin-top: ${th('space.7')};
+  text-align: center;
+  color: ${th('colorTextSecondary')};
+`
+
+const DashboardList = ({ manuscripts }) => {
+  if (!manuscripts.length) {
+    return (
+      <EmptyListMessage>
+        You currently have no active submissions
```
jsms90: Any reason why this isn't inside a `p` tag but the next line is?

app/components/ui/molecules/DashboardList.md
```diff
@@ -0,0 +1,23 @@
+Dashboard list of submissions
+
+```js
+;<DashboardList
+  manuscripts={[
```
jsms90: Could you also add the other view of this component to this markdown file - the `if (!manuscripts.length)` version?

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -1,90 +1,78 @@
...
-                <Button onClick={() => deleteManuscript(manuscript.id)} small>
-                  Delete
-                </Button>
-              </Cell>
-            </tr>
-          ))}
-        </tbody>
-      </table>
-    )}
+const Dashboard = ({ createManuscript }) => (
```
jsms90: Could we add a markdown file for this component so that the dashboard appears in styleguidist? :pray:

app/components/pages/Dashboard/Dashboard.js
Cooryd: I'm a derp 🙌

app/components/pages/Dashboard/Dashboard.js
Cooryd: This makes sense, I'll change this back to `BoxHiddenFromMobile` for the time being then.

app/components/pages/Dashboard/Dashboard.js
```diff
@@ -1,90 +1,78 @@
...
-const BoxHiddenFromMobile = styled(Box)`
+const RightAlignedButton = styled(Box)`
+  margin: ${th('space.3')} 0 ${th('space.1')};
+  text-align: right;
   display: none;
   ${media.mobileUp`display: block`};
 `
 
+const DashboardPanel = styled(Tabs.Panel)`
+  min-height: 300px;
```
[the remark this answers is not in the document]
Cooryd: I put this here as the `min-height` is used to solve a specific dashboard requirement of pushing the `< CenterdSmallParagraph>` component down the page when the dashboard list is empty. If the `Tabs` were to be used elsewhere, the min-height might not be required or might be different.

app/components/pages/Dashboard/ManuscriptsQuery.js
Cooryd: I agree, I think for the current batch of work this should be fine as the loading message shouldn't be noticeably visible as long as you're on a half decent connection but It would be good to have a styled message or 'spinner' in the future.

app/components/pages/Dashboard/Dashboard.js
Cooryd: I wasn't able to add this to a markdown as it needs the Apollo wrapper for the `ManuscriptsQuery` components. I did some refactoring of the code though and decided this could all be shifted into the main index.js as it is the core 'page' file now.  Please review the structure changes I'm about to push up and let me know if it makes sense or if it was better to have this `Dashboard` component

app/components/pages/Dashboard/Dashboard.js
Cooryd: I've extracted the content of these dashboard panels into their own files as each one should be it's own 'top' level file of sorts, which would kind of fit with what had been agreed previously. I have a feeling these panels may be mapped to routes in the future too so doing this would make it easier to alter later on.

jsms90:
Hmm, just noticing your last commit. We probably need tests for the functionality on this page. The lack of tests for the 'New Submission' buttons is totally my bad! Should have been done in PR #805 really

But perhaps we should also have tests for the `DashboardList` component? e.g. clicking on a row in the dashboard list - https://github.com/elifesciences/elife-xpub/pull/811/files#diff-99564bfef39de41bd6fafb9b3f039286R37

We could raise a separate issue for tests though?

app/components/pages/Dashboard/ManuscriptsQuery.js
jsms90: Yeah I think the loading screen would be ideal, but the lack of error screen could conceivably be more important. Not only for slow internet connections, which is a very real possibility, but what if there is an error on our side?

PR #836, opened 2018-10-18
fix: clean up styleguide errors

tamlyn:
#### Background

- Fix the styleguide examples to clean up the console errors.
- Add a button to show/hide the sticky footer example otherwise it's always there.

### DIFF
--- a/app/components/ui/atoms/StickyFooter.md
+++ b/app/components/ui/atoms/StickyFooter.md
@@ -3,18 +3,24 @@ Footer which is stuck to the bottom of the viewport.
 ...look to the bottom of the viewport to see this component :)
 
 ```js
+initialState = { show: false }
 const Button = require('@pubsweet/ui').Button
 const { Flex, Box } = require('grid-styled')
-;<StickyFooter>
-  <Flex>
-    <Box mr={3}>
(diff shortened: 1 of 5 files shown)


app/components/ui/atoms/StickyFooter.md
```diff
@@ -3,18 +3,24 @@ Footer which is stuck to the bottom of the viewport.
...
-    </Box>
-    <Box>
-      <Button primary onClick={() => console.log('accepted')}>
-        Click on me!
-      </Button>
-    </Box>
-  </Flex>
-</StickyFooter>
+;<div>
+  <button onClick={() => setState({ show: !state.show })}>Toggle</button>
```
jsms90: Might be worth changing the text that describes this component. Right now it says '....look to the bottom of the viewport to see this component', which is no longer the case.

Could also be good to make it really clear (without needing to first look at the code) that this button has nothing to do with the `StickyFooter` itself.

diversemix:
@jsms90 .. could you make this change so we can push this through?

app/components/ui/atoms/StickyFooter.md
diversemix: @jsms90 .. it works for me. It toggles the `StickyFooter` at the bottom of the viewport :thinking:

PR #837, opened 2018-10-18
feat: callout text box

### LINKED ISSUE #823: Create blue calloutbox
Blue box to call visual attention to the contents

- Takes _any_ content (so far only have designs for paragraphs of text inside), insets from the regular paragraphs with border around & background colour

For use in multiple places within the static pages of the site:
Component in [Zeplin](https://app.zeplin.io/project/5bacf042615a8b6e20e84a5f/styleguide/components):
image

In context - [Author Guide in eJP](https://submit.elifesciences.org/html/elife_author_instructions.html#process):
image

tamlyn:
#### Background

Close #823

### DIFF
--- a/app/components/ui/atoms/CalloutTextBox.js
+++ b/app/components/ui/atoms/CalloutTextBox.js
@@ -1,0 +1,16 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const Frame = styled(Box)`
+  border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorPrimary')};
+  background-color: ${th('colorPrimary')}20;
+`
+
+const CalloutTextBox = ({ onClose, children, ...props }) => (
+  <Frame mb={2} p={3} {...props}>
+    {children}{' '}
+  </Frame>
+)
+export default CalloutTextBox
--- a/app/components/ui/atoms/CalloutTextBox.js
+++ b/app/components/ui/atoms/CalloutTextBox.js
@@ -5,6 +5,7 @@ import { th } from '@pubsweet/ui-toolkit'
 
 const Frame = styled(Box)`
   border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorPrimary')};
+  // 8-character hex code - applying opacity value on top of 6-character hex code from theme
   background-color: ${th('colorPrimary')}20;
 `
 
--- a/app/components/ui/atoms/CalloutTextBox.js
+++ b/app/components/ui/atoms/CalloutTextBox.js
@@ -8,7 +8,7 @@ const Frame = styled(Box)`
   background-color: ${th('colorPrimary')}20;
 `
 
-const CalloutTextBox = ({ onClose, children, ...props }) => (
+const CalloutTextBox = ({ children, ...props }) => (
   <Frame mb={2} p={3} {...props}>
     {children}{' '}
   </Frame>
--- a/app/components/ui/atoms/CalloutTextBox.md
+++ b/app/components/ui/atoms/CalloutTextBox.md
@@ -1,0 +1,8 @@
+```js
+<CalloutTextBox>
+  On January 1, 2017, eLife introduced a fee for publication. A fee of $2,500*
+  is collected for all published papers submitted on or after this date;
+  however, authors with insufficient funding to pay the fee are eligible for a
+  fee waiver (read more).
+</CalloutTextBox>
+```


app/components/ui/atoms/CalloutTextBox.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const Frame = styled(Box)`
+  border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorPrimary')};
+  background-color: ${th('colorPrimary')}20;
```
tamlyn: Am I allowed to do this? 😁

diversemix: Looks like you have... It does assume the format of 'colorPrimary' ... Maybe we should put a comment near its definition?

app/components/ui/atoms/CalloutTextBox.js
```diff
@@ -0,0 +1,16 @@
...
+`
+
+const CalloutTextBox = ({ onClose, children, ...props }) => (
+  <Frame mb={2} p={3} {...props}>
```
jsms90: It's worth noting that `p={3}` means that there is padding top, left, bottom & right of `24px`. So, unlike the designs:
image

We don't have `18px` top and bottom. In fact we have `24px`:
image

@chugginselifesciences are you fine with this? I imagine that @tamlyn made this call based on the fact that we don't have a `18px` measurement in our grid units (see [`space` in our theme](https://github.com/elifesciences/elife-xpub/blob/develop/client/elife-theme/src/index.js#L49-L59))

app/components/ui/atoms/CalloutTextBox.js
```diff
@@ -0,0 +1,16 @@
...
+import styled from 'styled-components'
+import { Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const Frame = styled(Box)`
+  border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorPrimary')};
+  background-color: ${th('colorPrimary')}20;
+`
+
+const CalloutTextBox = ({ onClose, children, ...props }) => (
```
jsms90: I think `onClose` might have been added accidentally here. Perhaps @tamlyn was originally copy/pasting from our `CalloutBox` file, but there is no open/close functionality associated with this component.

app/components/ui/atoms/CalloutTextBox.js
jsms90: I'm also still not 100% convinced about putting `mb={2}` onto components themselves, rather than letting the parent handle the spacing between each component on the page :thinking:

app/components/ui/atoms/CalloutTextBox.js
jsms90: I think it's safe to assume that all of our colours, coming from our theme are in hex format. But I agree that maybe we should put a comment above this line, clarifying that the added 20 relates to the opacity. I imagine that this won't happen in many places within our code, but if it does, perhaps we should think about creating a helper function so that this is more obvious to read

chugginselifesciences:
The intention was to get a consistent visible space between the text and container. So 18px space with line height gives an apparent 24px gap ignoring the descenders. 

image

The screenshot you added gives a larger 'visible' space

image


I realise that it might not be consistent with how we're doing vertical spacing in most places, but we should be doing something similar with h1s.

image

jsms90:
I think the issue here comes from the fact that we have not implemented the ["typography spacing" part of the styleguide](https://app.zeplin.io/project/5bacf042615a8b6e20e84a5f/screen/5bc0cddca55ed15fd3fb3973) at all.
image

We hadn't seen this concept until after creating the wizard steps & it didn't seem worth adding this extra margin/padding around the lettering, as all of this is much more important when the headings & paragraphs are next to each other in a page flow. Since we had only been creating headings at the top of a small form & were only using "paragraphs" for a single small line of text (rather than an actual paragraph), we didn't go back and add them in.

This seems like the right time to start changing our `Paragraph` & `Headings` components (or adding new ones that exist for this different use case), so that we have this extra spacing around them? In which case, I'll create some new tickets.

app/components/ui/atoms/CalloutTextBox.js
jsms90: Do you reckon this helps, @diversemix ?

jsms90:
@chugginselifesciences yeah I agree that this isn't important for this component. I imagine that it will be more important for the static pages. Either way, happy to leave the spacing-related comments & merge as is :+1:

app/components/ui/atoms/CalloutTextBox.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const Frame = styled(Box)`
+  border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorPrimary')};
+  // 8-character hex code - applying opacity value on top of 6-character hex code from theme
```
diversemix: Looks good!

PR #844, opened 2018-10-19
NewRelic Browser

### LINKED ISSUE #619: Configure New Relic to monitor xPub
- [x] add NewRelic library to xPub
- [ ] create NewRelic config, either in the application or on the NewRelic dashboard
- [ ] add NewRelic application key to xPub
- [ ] consider setting up NewRelic Infra, if eLife requires it

damnhipster:
#### Background

Enables NewRelic Browser on [staging](https://rpm.newrelic.com/accounts/1707959/browser/162979288) and [production](https://rpm.newrelic.com/accounts/1451451/browser/162983119)

Only production has NewRelic Pro enabled, which enables features such as JS errors and network request monitoring.

#### Any relevant tickets

Closes #619 

Ran application in production and staging mode, and verified the following:
- No console errors
- Correct license keys on staging and production
- NewRelic stats sent over network requests on the browser
- Stats appear in NewRelic Browser
- NewRelic Browser not enabled when running locally


config/default.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     logger,
     uploads: 'uploads',
     enableExperimentalGraphql: true,
+    secret: 'some-secret',
```
tamlyn: I think it's best not to put a default secret in the config in order to avoid accidentally deploying to production with a non-secret secret.

damnhipster: Agreed, it's a better idea to keep it only in `development.js`.

config/prod.js
```diff
@@ -25,4 +25,8 @@ module.exports = {
       },
     },
   },
+  newrelic: {
+    licenseKey: 'c53c018d69',
```
giorgiosironi: for the record, this (which ends up in the HTML page) is different from the full license key that APM and other products use. Not sensitive I'd say

PR #851, opened 2018-10-22
813 - sidebar layout

### LINKED ISSUE #813: Create layout component for the sidebar-content pattern

2 columns:
- thin component on left (sidebar)
- rest on the right (content)

Layout component:
image

Used in context (one of many):
image

@chugginselifesciences what do you reckon the min-width of the sidebar should be?

Cooryd:
#### Background

Creates a Side Nav layout component for use on static content pages. Takes config to generate a navigation list on the left side of the page.

#### Any relevant tickets

closes #813

Visually in browsers.
Created basic test script to make sure nav links are generating correctly.


app/components/global/layout/SideNavLayout.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Link } from 'react-router-dom'
+import { Flex, Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const SideNavLink = styled(Link)`
```
tamlyn: I'd recommend using [NavLink](https://reacttraining.com/react-router/web/api/NavLink) since this automatically sets the `active` class based on the URL.

app/components/global/layout/SideNavLayout.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { NavLink } from 'react-router-dom'
+import { Flex, Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const SideNavLink = styled(NavLink)`
```
jsms90: Can I suggest that we pull a generic `NavLink` styled in the eLife way out into `atoms`?

Then we can use it here and [in `AppBar`](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/global/AppBar/AppBar.js#L17-L28) (albeit with ever so slightly different styles applied in each situation).

Partly just to reduce duplication. But also because, having it in `atoms` then makes it more obvious to anyone wanting to use a `NavLink` again, that we already have a pattern designed for this.

Issue #780, opened 2018-10-11
Update DashboardListItem Component with responsive layout


hdrury1:
![screen shot 2018-10-11 at 13 05 34](https://user-images.githubusercontent.com/17023702/46802814-64fd3f00-cd56-11e8-96f8-9f96c665c938.png)

tamlyn:
At what break point should it switch to this view?

diversemix:
@chugginselifesciences says "480" .. but he's just sticking it in the designs as we speak

PR #852, opened 2018-10-22
style(Dashboard): dashboard Items responsive design

jyenes:
#### Any relevant tickets
closes: #780


app/components/ui/molecules/DashboardListItem.js
```diff
@@ -29,32 +30,44 @@ const dashboardDateText = date => {
...
-const RelativeDate = styled.p`
+const RelativeDate = styled(Box)`
   margin-bottom: 0;
-  font-size: 16px;
+  font-size: ${th('size.7')};
 `
-const AbsoluteDate = styled.p`
+const AbsoluteDate = styled(Box)`
   font-size: 12px;
   margin-top: 0;
```
jsms90: As above - to remove default margins on `<p>` element?

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -29,32 +30,44 @@ const dashboardDateText = date => {
...
+
+  ${media.mobileUp`
+    flex-direction: column;
+    text-align: right;
+    flex: 0 0 120px;
+  `};
 `
-const RelativeDate = styled.p`
+const RelativeDate = styled(Box)`
   margin-bottom: 0;
```
jsms90: If you're changing this from a `p` to a `Box`, I don't think this line is necessary any more? I imagine that you only had this before, in order to remove the default margins that exist on `<p>` tags?

app/components/ui/atoms/ManuscriptStatus.js
```diff
@@ -2,12 +2,16 @@ import React from 'react'
 import styled from 'styled-components'
 import { Box } from 'grid-styled'
 import PropTypes from 'prop-types'
+import { th } from '@pubsweet/ui-toolkit'
+import media from '../../global/layout/media'
 
 const StatusBox = styled(Box)`
```
jsms90: Given that this styling is very specific to the use of `ManuscriptStatus` within `DashboardListItem` (as opposed to it being re-usable in multiple contexts), I'm not sure that having `ManuscriptStatus` as its own separate atom is particularly necessary? If anything, it just means that a developer now needs to look in 2 files to know how `DashboardListItem` is built.

What do you think of putting this inside `DashboardListItem.js`?

Cooryd: I've got no reason other than personal choice to argue against this. I find working on smaller files easier as it means theres less helper functions to scroll through to find what you're after and with cmd - click it makes it simple enough to navigate to imported files. Thats my opinion though, so I wouldn't disagree with making this change if the consensus was there.

diversemix: My vote is the path of least resistance -- so it looks like I'm siding with "if it aint broke don't fix it"

jsms90: Yeah I'm happy to chalk this up to personal preference. If anyone actually finds it an unnecessary extra piece in future, they're welcome to amalgamate the two. There's little point in changing it now, if it's not a widely held opinion :+1:

diversemix:
Looks good to me 
image

Issue #814, opened 2018-10-18
Create grid component for Contact Us page - 3 columns on desktop


jsms90:
**Product requirements/Acceptance criteria:**

Functional:
- On desktop: 3 columns
- On mobile: ?

**Relevant links to design(s):**
![image](https://user-images.githubusercontent.com/15656538/47146938-09d2cb80-d2c5-11e8-922a-0aa8ad3b7345.png)
![image](https://user-images.githubusercontent.com/15656538/47146953-135c3380-d2c5-11e8-9507-951af63be394.png)

PR #864, opened 2018-10-24
StaffGrid: generic Grid to display staff cards

jyenes:
closes: #814

### DIFF
--- a/app/components/ui/molecules/StaffGrid.js
+++ b/app/components/ui/molecules/StaffGrid.js
@@ -1,18 +1,12 @@
 import React from 'react'
-import styled from 'styled-components'
 import { Flex, Box } from 'grid-styled'
-import { th } from '@pubsweet/ui-toolkit'
-
-const Card = styled(Box)`
-  padding-top: ${th('space.3')};
-`
 
 const StaffGrid = ({ children, ...props }) => (
-  <Flex flexWrap="wrap" mx={-2} {...props}>
+  <Flex flexWrap="wrap" mb={-3} mx={-2} {...props}>
     {children.map((item, index) => (
-      <Card key={item.key || index} px={2} width={[1 / 2, 1 / 2, 1 / 3]}>
+      <Box key={item.key || index} mb={3} px={2} width={[1 / 2, 1 / 2, 1 / 3]}>
         {item}
-      </Card>
+      </Box>
     ))}
   </Flex>
 )
--- a/app/components/ui/molecules/StaffGrid.js
+++ b/app/components/ui/molecules/StaffGrid.js
@@ -1,30 +1,20 @@
 import React from 'react'
 import styled from 'styled-components'
-import StaffCard from './StaffCard'
+import { Flex, Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
 
-const { Flex, Box } = require('grid-styled')
-
-const Grid = styled(Flex)`
-  flex-flow: row wrap;
-  justify-content: space-between;
-`
 const Card = styled(Box)`
-  padding-top: 20px;
+  padding-top: ${th('space.3')};
 `
 
-const StaffGrid = ({ staff }) => (
-  <Grid>
-    {staff.map(person => (
-      <Card px={2} width={[1 / 2, 1 / 2, 1 / 3]}>
-        <StaffCard
-          jobTitle={person.jobTitle}
-          name={person.name}
-          photoURL={person.photoURL}
-          telephone={person.telephone}
-        />
+const StaffGrid = ({ children, ...props }) => (
+  <Flex flexWrap="wrap" mx={-2} {...props}>
+    {children.map((item, index) => (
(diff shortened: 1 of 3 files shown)


app/components/ui/molecules/StaffGrid.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import styled from 'styled-components'
+import StaffCard from './StaffCard'
+
+const { Flex, Box } = require('grid-styled')
```
tamlyn: Use `import` in client side components.

app/components/ui/molecules/StaffGrid.js
```diff
@@ -0,0 +1,29 @@
...
+
+const Grid = styled(Flex)`
+  flex-flow: row wrap;
+`
+const Card = styled(Box)`
+  padding-top: 20px;
```
tamlyn: Unless we have specific designs for this, I'd suggest using `24px` here (i.e. `space.3`) for consistency with the grid.

app/components/ui/molecules/StaffGrid.js
```diff
@@ -0,0 +1,29 @@
...
+`
+
+const StaffGrid = ({ staff }) => (
+    <Grid>
+      {staff.map(person => (
+        <Card Box px={2} width={1 / 3}>
```
jsms90: Is the word `Box` here a typo? I'm not sure that it's doing anything.

I reckon you might want media queries on `Card`? Right now, on mobile, it looks like this:
image

I imagine that we'll want some kind of breakpoint, after which this becomes a 2X2 grid. What do you think?

Also, is there any particular reason why we have `px={2}` (i.e. 12px) here, but the vertical spacing between cards is 20px?

app/components/ui/molecules/StaffGrid.js
jsms90: If we go with 24px instead of 20px, then we can use a theme variable `${th('space.3')}`.

Is there any particular reason that it's 20px right now?

app/components/ui/molecules/StaffGrid.js
tamlyn: Having padding on both sides causes the cards to be indented from the text. Have a look at how `TwoColumnLayout` uses negative margins to prevent this.

Also when resized, the images get squeezed. Fixing this probably requires changes to `StaffCard`.

app/components/ui/molecules/StaffGrid.md
```diff
@@ -0,0 +1,31 @@
...
+    jobTitle: 'Senior Editorial Assistant',
+    telephone: '+44 [0]1223 750665',
+  },
+  {
+    photoURL: 'https://submit.elifesciences.org/images/Susanna.png',
+    name: 'Susanna Richmond',
+    jobTitle: 'Editorial Assistant',
+    telephone: '+44 [0]1223 855377',
+  },
+]
```
tamlyn: There's nothing *wrong* with passing an array of JavaScript objects to this component but it's not very *Reacty*. Again, have a look at `TwoColumnLayout` to see how you can pass the data as simple React elements. This also has the advantage that you decouple the content from the layout, allowing the layout to potentially be used with other components.

tamlyn: Actually I just noticed `TwoColumnLayout` [can be improved](https://github.com/elifesciences/elife-xpub/pull/865).

app/components/ui/molecules/StaffGrid.js
```diff
@@ -0,0 +1,30 @@
...
+  justify-content: space-between;
+`
+const Card = styled(Box)`
+  padding-top: 20px;
+`
+
+const StaffGrid = ({ staff }) => (
+  <Grid>
+    {staff.map(person => (
+      <Card px={2} width={[1 / 2, 1 / 2, 1 / 3]}>
```
jsms90: I had a comment here but must not have added it to the review properly.

Perhaps we could do `px={3}` so that the horizontal spacing in between cards (i.e. gutters between columns) is `24px` & therefore the same as the vertical spacing between cards (i.e. the gutters in between rows)

diversemix:
@jsms90 ... Are we good to merge this now?

app/components/ui/molecules/StaffGrid.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Flex, Box } from 'grid-styled'
+import { th } from '@pubsweet/ui-toolkit'
+
+const Card = styled(Box)`
+  padding-top: ${th('space.3')};
```
jsms90: I'm really sorry for leaving these comments in so many bits @jyenes but there's one final thing I've just noticed. This is creating spacing above the grid, as well as between the cards.

How about doing something similar to the x-axis padding?

I'm thinking something like:
```js
<Flex mb={-3} mx={-2}>
    {children.map((item, index) => (
      <Box mb={3} mx={2}
```

jyenes: don't worry @jsms90 good catch

PR #865, opened 2018-10-24
fix: allow TwoColumnLayout to handle single child

tamlyn:
#### Background

Use `React.Children.map` to allow layout to handle single, non-array child.


tamlyn:
Previously this worked:

    <TwoColumnLayout><p>one thing</p><p>another thing</p></TwoColumnLayout>

but this didn't

    <TwoColumnLayout><p>only thing</p></TwoColumnLayout>

Now both work.

Issue #834, opened 2018-10-18
Create layout component for DropdownNavigation-Content view


jsms90:
This encapsulates the pattern for "mobile"

Layout component:
![image](https://user-images.githubusercontent.com/15656538/47165666-65677e00-d2f2-11e8-9da6-6d661f336a74.png)

Used in context:
![image](https://user-images.githubusercontent.com/15656538/47166060-3a315e80-d2f3-11e8-9b36-538f485be5fe.png)

jsms90:
From what I can tell, this seems to have been done as part of PR #855 

In styleguidist, the `SectionalLayout` component has these views:
- desktop  
![image](https://user-images.githubusercontent.com/15656538/47492591-482c3580-d845-11e8-88b8-29f9453eb02a.png)

- mobile
![image](https://user-images.githubusercontent.com/15656538/47492566-39458300-d845-11e8-84ba-8cde7af474e8.png)

So the mobile version of `SectionalLayout` already satisfies this task.

(I have a feeling that this was just a poor breakdown of issues, or separation of the designs into components, on my part.)

Issue #820, opened 2018-10-18
Create dropdown component for mobile version of static pages


jsms90:
**Product requirements/Acceptance criteria:**

Depends on #821 

Functional:
- The words in the top button of the dropdown should show the words associated with the current page
- The list below should contain all of the sub-pages that a user can navigate to within the Author Guide/Reviewer Guide/Contact Us page that they are on
- When a user clicks on .... (which part of this component? just the words? just the down chevron? both? the whole bar?)
  - the dropdown should appear
  - the down chevron should turn into an up chevron
- When a user clicks on one of the items in the dropdown menu
  - the user should be navigated to the chosen sub-page  

**Relevant links to design(s):**
![image](https://user-images.githubusercontent.com/15656538/47155987-f97b1a80-d2dd-11e8-9e7a-111a87d7556b.png)
![image](https://user-images.githubusercontent.com/15656538/47155953-e405f080-d2dd-11e8-84ad-89d64ebc13b6.png)

damnhipster:
Depends on #821

Issue #786, opened 2018-10-11
Create submission/archive dropdown menu ("tabs" for mobile version)


jsms90:
**Product requirements/Acceptance criteria:**

Functional:
- "Submissions" selected by default
- When a user clicks on the down arrow a menu appears, containing "Submissions" & "Archive"
- When a user selects an alternative from the menu, the list of manuscripts below updates (in the same way that they would be clicking a tab on the desktop view)

**Relevant links to design(s):**
[Mobile invision prototype](https://projects.invisionapp.com/share/FDOIB61EKV5#/screens/322899359)
![image](https://user-images.githubusercontent.com/41907215/46821528-d520ba80-cd80-11e8-96f9-0570c65926f5.png)

jsms90:
Sounds like, in this case, the switch between desktop and mobile actually changes which component we render altogether, this might be a candidate for the library that @tamlyn was talking about - https://github.com/elifesciences/elife-xpub/issues/781#issuecomment-428960111

tamlyn:
@hdrury1 Would it make sense to deprioritise this mobile view since we only have two tabs and they fit OK side by side on a mobile screen?

hdrury1:
@tamlyn Can we take a look at what the mobile rendering is like once the desktop version has been completed? Definitely worth getting that right first. I'll rearrange the tickets in To Do accordingly.

Cooryd:
The side tabs on #505 will be styled in a similar way on mobile. Would be useful for this component to be abstract enough to be used there as well.

PR #867, opened 2018-10-24
create NavigationDropdown component

### LINKED ISSUE #821: Create generic dropdown menu component
Needs a better name than "dropdown" to differentiate between:
- this component
- the user menu component (which is also a dropdown)
- the Article Type dropdwown component on step 3 of the wizard (currently using pubsweet's "Menu" component, which needs to be replaced - see #326)

Generic component for use in mobile version of static pages #820 & mobile version of dashboard #786 

Used in at least 2 places in the app:
image

image

damnhipster:
~Addresses #834~

#### Background
We have multiple different dropdowns in our codebase now:
- SubjectAreaDropdown - multi-select based on `react-select`
- Menu component from pubsweet (used for Article Types menu - needs to be replaced)
- ~Search in PeoplePicker~ - dropdown has been removed (for now)
- Hamburger menu
- UserMenu

This is a different type again

#### What does this PR do?
Creates dropdown component for use in:
- mobile version of dashboard (instead of tabs)
- mobile version of sub-routing menu on static pages

#### Any relevant tickets
fixes #821 

Relying on react-select tests


Cooryd:
This bit of work seems to link to #821. Maybe when you come to styling this Select it could be cast out into a stand alone component for reuse on the Dashboard tabs mobile view?

jsms90:
I would say that this is actually a combination of #821 (the generic dropdown component, for use in multiple situations) & issue #820 (using this dropdown component in 1 specific context)

Since work has already started on these, I'm going to update the kanban board, just so someone else doesn't start this work accidentally :+1:

app/components/global/layout/DropdownNavLayout.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+import NavLink from '../../ui/atoms/NavLink'
+import Select, { createFilter, components } from 'react-select'
+
+const Root = styled.div``
+
+const DropdownNavLink = styled(NavLink)`
```
jsms90: I know this PR is still a WIP but I would suggest that we leave this part (i.e. _which_ content is going to get inserted into the dropdown menu) to another PR, which addresses #820 ?

app/components/global/layout/DropdownNavLayout.js
```diff
@@ -0,0 +1,64 @@
...
+  display: block;
+  padding: ${th('space.1')};
+`
+
+class DropdownNavLayout extends React.Component {
```
jsms90: If you find it easier to create a Nav-specific dropdown component first (issue #820), that's fair enough.

But just be aware that we'll need to refactor this later and pull out the code which is generic to all dropdowns, regardless of the content in the list. Then use that generic dropdown (issue #821) in order to build the version in the dashboard (issue #786). So you might be better off just making the generic one first? Just a suggestion.

Either way, it might be useful to have this component in the `ui/atoms` folder rather than `global/layout`?

jsms90:
I think it's worth picking up where you left off on this branch, rather than starting a new branch @damnhipster 

On it now :+1:

app/components/ui/atoms/NavigationDropdown.js
```diff
@@ -0,0 +1,87 @@
...
+        components={{
+          IndicatorSeparator: null,
+          DropdownIndicator: CustomDropdownIndicator,
+        }}
+        getOptionLabel={({ label }) => label}
+        getOptionValue={({ value }) => value}
+        isSearchable={false}
+        maxMenuHeight={1200} // longer than any phone screen
+        onChange={({ value }) => this.setState({ value })}
+        options={options.filter(({ value }) => value !== this.state.value)}
```
Cooryd: I'm not sure we should be filtering the selected value from the designs it looks like the current selected should be top of the select list and `textColor`

app/components/ui/atoms/NavigationDropdown.js
```diff
@@ -0,0 +1,87 @@
...
+        ...base,
+        // remove default space between control & menu
+        marginBottom: 0,
+        // leave room for the bottom border of the Control component to be visible on focus
+        marginTop: this.props.theme.borderWidth,
+      }),
+      option: (base, { isSelected, isFocused }) => ({
+        ...base,
+        backgroundColor: this.props.theme.colorBackground,
+        color: this.props.theme.colorText,
```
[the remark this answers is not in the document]
jsms90: Sorry, should have made this more explicit in the top comment of the PR. Specs have changed - see [this comment from @chugginselifesciences](https://github.com/elifesciences/elife-xpub/issues/821#issuecomment-433848150)

app/components/ui/atoms/NavigationDropdown.js
jsms90: Good point. I didn't clock that we were supposed to leave the current page in the dropdown.

Asking now, especially based on the lack of colour distinction: https://github.com/elifesciences/elife-xpub/issues/821#issuecomment-434657224

app/components/ui/atoms/NavigationDropdown.js
```diff
@@ -0,0 +1,87 @@
...
+      <Select
+        components={{
+          IndicatorSeparator: null,
+          DropdownIndicator: CustomDropdownIndicator,
+        }}
+        getOptionLabel={({ label }) => label}
+        getOptionValue={({ value }) => value}
+        isSearchable={false}
+        maxMenuHeight={1200} // longer than any phone screen
+        onChange={({ value }) => this.setState({ value })}
```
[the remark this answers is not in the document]
jsms90: Oh yes. I've changed this in PR #929 but should be here really :+1:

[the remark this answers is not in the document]
jsms90: Ah ok, no.

In order to know which link should be black, this component will need it's own state: https://github.com/elifesciences/elife-xpub/issues/821#issuecomment-434665623

app/components/ui/atoms/NavigationDropdown.js
jsms90: Ok, removed the filter based on https://github.com/elifesciences/elife-xpub/issues/821#issuecomment-434665623 :+1:

app/components/ui/atoms/NavigationDropdown.js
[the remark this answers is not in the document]
jsms90: Spec has changed back - https://github.com/elifesciences/elife-xpub/issues/821#issuecomment-434665623

Doing this now :+1: 

Unfortunately, react-select doesn't easily allow you to pass in something like "initially selected item", so this isn't going to be quite as simple as it would be, if this component's state weren't controlled by the parent.

Hopefully won't take long

app/components/ui/atoms/NavigationDropdown.js
```diff
@@ -0,0 +1,97 @@
...
+        getOptionLabel={({ label }) => label}
+        getOptionValue={({ value }) => value}
+        isSearchable={false}
+        maxMenuHeight={1200} // longer than any phone screen
+        onChange={this.handleSelection}
+        options={options}
+        placeholder=""
+        styles={this.customReactSelectStyles}
+        value={options.filter(
+          ({ value }) => value === this.state.selectedValue,
```
Cooryd: Should this be `this.props.selectedValue`? the state `selectedValue` will only be set on the constructor then looks to be set to `{}` in the `this.handleSelection` function. Am I missing some build in `Select` functionality?

jsms90: Woops! No, that's totally my bad. Line 51 shouldn't set state to be nothing :laughing: 

Thanks :+1:

jsms90: Also, styleguidist is now working correctly, so should make it a bit easier!

jsms90:
@hdrury1 

### Colours

> Component looks like it has a grey background colour before it's opened... it should be white

I noticed this too, but I think this might be something to do with styleguidist, because everywhere in `react-select`'s code, it explicitly sets the background colour to be white & when I put this component into our app, it looks white.

...Not sure why it looks slightly grey on styleguidist :flushed:

I'm pretty sure we're good here. :heavy_check_mark: 

> Looks like a blue background is applied when choosing an option from the dropdown (there shouldn't be)

Interesting. There must be some `react-select` default styling that I'm not overriding correctly. Will address that now.

Assuming it's a light blue, rather than eLife's `colorPrimary` blue? (otherwise I'm really confused)

> Focus ring is covering the grey border (this may be something to do with the broswer)

Yeah, as we discussed on slack, I'll remove all visual indication of focus for now, since we've already removed it everywhere else from xpub anyway. I've opened a new issue for us to address the issue of focus styling at some point #940 :grin:

jsms90:
### Dropdown indicator (arrow)
> Arrow should be inverted when the menu is open (pointing upwards)

Good point - found this issue on `react-select`: https://github.com/JedWatson/react-select/issues/2926. So I could start there.

But, this might not be the right approach, depending on how we resolve the next problem:

> Arrow should be right aligned to the text, rather than the component (see design)

This one may be tricky! The way that `react-select` is built, is for the dropdown indicator to be on the right hand side of the Control (the top/selected value).

I'm genuinely not sure how to achieve the positioning in the designs...

Also, looking ahead slightly, if we do intend to put the multi-select button inside this dropdown component, then that may affect the choices I make here:
image

@chugginselifesciences how likely is this to be the final position of that button?

jsms90:
All done except for the arrow positioning @hdrury1

jsms90:
## :tada: Ok, done with the positioning :tada:

FYI for anyone trying to do this with react-select in future, I've removed the `DropdownIndicator` component altogether
```js
<Select
  components={{
    DropdownIndicator: null,
  }}
  options={options}
/>
```
and re-created the arrow icon with CSS (as we already do in the `ProfileMenu` component), rather than using an SVG from Material icons.

By doing that, it means I can position this CSS on the `SingleValue` sub-component of react-select using the `::after` pseudo class.

So far, I haven't found any issues with the functionality. So I imagine that react-select doesn't entirely rely on a `DropdownIndicator` existing, but if we ever find an unusual problem, this may be the first place to look.

hdrury1:
Seen this locally and LGTM

Issue #871, opened 2018-10-24
Create hamburger menu component


jsms90:
**Product requirements/Acceptance criteria:**
- Clicking on the burger should open the full menu from the left-hand side of the screen
- Clicking on an option should close the menu and load the selected page
- Clicking on the cross should close the menu with no action
- Clicking off the menu (i.e. to the right) should close the menu with no further action

Note: The links for the reviewer guide, author guide and contact us pages should appear even when the user is not logged in

**Relevant links to design(s):**

Before click:
![image](https://user-images.githubusercontent.com/15656538/47441229-97238d80-d7a7-11e8-956b-0edd8ca08e50.png)

After click:
![image](https://user-images.githubusercontent.com/41907215/46817671-6b4fe300-cd77-11e8-8135-5af6218a4f72.png)

- The active link is colorText
- Inactive links are colorTextSecondary

[See Zeplin](https://app.zeplin.io/project/5bacf042615a8b6e20e84a5f/styleguide/components)

PR #876, opened 2018-10-25
871 - Burger Menu Component

Cooryd:
#### Background
A Burger menu is need for the mobile view of the App bar menu. This should be a 'Burger' Icon that shows a white side panel overlay when clicked with a cross to hide the menu and navigation links below.

What does this PR do?

Adds the BurgerMenu molecule component for implementation within the AppBar

#### Any relevant tickets
#871 

Visually 
Unit test for show/hide toggle and correct NavItem creation


app/components/ui/molecules/BurgerMenu.js
```diff
@@ -0,0 +1,75 @@
...
+
+const MenuPanel = styled.div`
+    max-width: 240px;
+    background-color: ${th('colorBackground')}
+    height: 100%;
+`
+
+const MenuItem = styled(Box)`
+  border-top: ${th('borderWidth')} ${th('borderStyle')} ${th('colorBorder')};
+  height: ${th('space.6')};
```
jsms90: I'm not sure whether this is necessary? The height seems to be generated well by the combination of line height on the navlink & padding that you've added to each `MenuItem`? But maybe I've missed something, and it's adding something more?

app/components/ui/molecules/BurgerMenu.js
```diff
@@ -0,0 +1,75 @@
...
+      <React.Fragment>
+        <ButtonAsIconWrapper
+          data-test-id="burger-menu-expand"
+          onClick={() => this.setState({ menuOpen: true })}
+        >
+          <Icon iconName="Menu" overrideName="@pubsweet-pending.AppBar.Menu" />
+        </ButtonAsIconWrapper>
+        <ModalOverlay open={this.state.menuOpen} transparentBackground>
+          <MenuPanel>
+            <MenuItem p={3}>
```
jsms90: Since you've got `p={3}` in a couple of places (here and line 50), perhaps it's worth popping this in via styled components.

i.e. line 18:
```js
const MenuItem = styled(Box).attrs({p:3})`
```

Not super important, but could be nice in the unlikely event that someone might want to add any extra `MenuItem`s?

app/components/ui/molecules/BurgerMenu.js
```diff
@@ -0,0 +1,75 @@
...
+  menuItems: PropTypes.arrayOf(
+    PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      link: PropTypes.string.isRequired,
+    }),
+  ),
+}
+
+BurgerMenu.defaultProps = {
+  menuItems: [],
```
jsms90: Could we make this a required field instead of defaulting to nothing? I assume we wouldn't want to allow for a situation where the burger menu had no items inside?

app/components/ui/atoms/ButtonAsIconWrapper.js
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components'
+
+const ButtonAsIconWrapper = styled.button.attrs({
```
jsms90: :+1:

app/components/ui/atoms/CrossIconButton.js
```diff
@@ -1,19 +1,9 @@
...
-const ButtonAsIconWrapper = styled.button.attrs({
-  type: 'button',
-})`
-  background-color: transparent;
-  border: none;
-  line-height: 0;
-  padding: 0;
-`
-
 const CrossIconButton = props => (
```
jsms90: When I created this component, I was debating whether this was worth having (vs just having `ButtonAsIconWrapper`.

Do you think it's still worth having as its own component? I'm tempted to say refactor by removing this component altogether & explicitly wrapping `CrossIcon` in `ButtonAsIconWrapper` in any context where it's necessary.

What do you think @Cooryd ?

app/components/ui/molecules/BurgerMenu.js
Cooryd: This is a fix for the Cross MenuItem, It was setting its inner height to be 31px rather than 24px (standard line-height) of the other link items, Couldn't figure out why the Icon was doing this as its dimensions were 24 x 24 and line-height was the same as all others

PR #883, opened 2018-10-25
refactor: split data access code from resolvers and routes

### LINKED ISSUE #708: Separate DB access code from resolvers
Having the data access logic and the GraphQL resolvers in the same package can lead to circular dependencies. For example the `finishSubmission` resolver needs to call the meca export package but the meca callback route handler needs to access the Manuscript entity in order to update its status. At this point having the meca functionality as a separate package no longer makes sense.

One solution to this is to separate out the data access code from the higher level resolvers and routes:

```
Resolvers ➡️ DB
  ↘️ MECA ↗️
```

tamlyn:
#### Background

Closes #708 

Lots of changed files but mostly just moving things around and fixing up import paths. Shouldn't be any change in functionality.


diversemix:
I merged this sooner rather than later - as its a refactor :+1:

PR #884, opened 2018-10-25
Add guides to s3. Also get all images from the journal site

### LINKED ISSUE #872: Add files to S3 that are currently hosted on eJP
Many of the static pages have links to various PDFs, docx & xlsx files. See PR https://github.com/elifesciences/elife-xpub/pull/849

- [x] give access to the `Xpub` IAM group to a prefix in `elife-cdn`
- [x] populate the `elife-cdn` S3 Bucket with the files into a prefix like `/xpub/...`
- [x] change the static links referenced in the above PR

diversemix:
#### Background

Hosts the files that were in eJP on S3
Also fix staff photos by getting them from the Journal site

#### Any relevant tickets

Closes #872 

Locally with the styleguide

### DIFF
--- a/app/components/pages/Staff/EditorialPage.js
+++ b/app/components/pages/Staff/EditorialPage.js
@@ -13,19 +13,19 @@ const EditorialPage = () => (
       <StaffCard
         jobTitle="Editorial Manager"
         name="Wei Mun Chan"
-        photoURL="https://submit.elifesciences.org/images/WeiMun.png"
+        photoURL="https://prod--journal-cms.elifesciences.org/sites/default/files/iiif/person/2017-07/weimun.jpg"
         telephone="+44 [0]1223 855379"
       />
       <StaffCard
         jobTitle="Journal Development Editor"
         name="Maria Guerreiro"
-        photoURL="https://submit.elifesciences.org/images/Maria.png"
+        photoURL="https://prod--journal-cms.elifesciences.org/sites/default/files/iiif/person/2017-07/maria.jpg"
         telephone="+44 [0]1223 855376"
       />
       <StaffCard
         jobTitle="Senior Editorial Assistant"
         name="Susanna Richmond"
-        photoURL="https://submit.elifesciences.org/images/Susanna.png"
+        photoURL="https://prod--journal-cms.elifesciences.org/sites/default/files/iiif/person/2017-07/susanna.jpg"
         telephone="+44 [0]1223 855377"
       />
       <StaffCard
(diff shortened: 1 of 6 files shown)


app/components/pages/Staff/EditorialPage.js
```diff
@@ -13,19 +13,19 @@ const EditorialPage = () => (
...
         name="Maria Guerreiro"
-        photoURL="https://submit.elifesciences.org/images/Maria.png"
+        photoURL="https://prod--journal-cms.elifesciences.org/sites/default/files/iiif/person/2017-07/maria.jpg"
         telephone="+44 [0]1223 855376"
       />
       <StaffCard
         jobTitle="Senior Editorial Assistant"
         name="Susanna Richmond"
-        photoURL="https://submit.elifesciences.org/images/Susanna.png"
+        photoURL="https://prod--journal-cms.elifesciences.org/sites/default/files/iiif/person/2017-07/susanna.jpg"
```
tamlyn: How can we record that these files are being used in xPub as well as Journal? I can imagine someone renaming, updating or moving these in Journal without realising that it breaks the xPub pages.

diversemix: Very good point! ... @giorgiosironi  @thewilkybarkid ???

jsms90:
Just wondering why some links seem to start with `https://elife-cdn.s3.amazonaws.com` & these new links start with `https://cdn.elifesciences.org`?

e.g. [in Editorial Process](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/pages/AuthorGuide/EditorialProcess.js#L74) & [Journal Metrics](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/pages/AuthorGuide/JournalMetrics.js#L23)

diversemix:
Ooo... We might need to change these to use the CDN? Good spot.

app/components/pages/Staff/EditorialPage.js
thewilkybarkid: Hard-coding any of these details looks like a bad idea, but you can use the `/people` API to fetch the IIIF path (so you can make it whatever size you need etc).

giorgiosironi:
> Just wondering why some links seem to start with https://elife-cdn.s3.amazonaws.com & these new links start with https://cdn.elifesciences.org?

Yes:
- `elife-cdn.s3.amazonaws.com` is the bucket
- `cdn.elifesciences.org` is the Fastly CDN on top of the bucket, to be preferred unless it's an internal application that needs to skip caches

app/components/pages/Staff/EditorialPage.js
giorgiosironi: And, `iiif.elifesciences.org` is not just a dynamic resizing service but also a CDN with cache, edge nodes located across the world, etc so it's the best way to access an image

PR #891, opened 2018-10-26
docs: remove 'Other' section from styleguide

tamlyn:
#### Background

- `ignore` glob isn't working (don't know why) which was causing components to be included twice
- just need to remember to add a section to the config if we ever have components in folders other than `app/{global,pages,ui}`.


damnhipster:
Addressed by #893

PR #892, opened 2018-10-26
888 - Consolidate AppBar menu config to be passed in one place

Cooryd:
#### Background

Previously AppBar's Desktop menu was hard coded NavLinks and a seperate set of config was passed to burger menu for the mobile view. This left the possibility of adding to one in the future and forgetting about the other.

What does this PR do?

Allows config for the AppBar menu to be passed in through the Layout Component as `userMenuItems` for signed in users and `defaultMenuITems` for all users. This one entry point for config is then used to generate both mobile and desktop view menus

#### Any relevant tickets

closes#888

locally

### DIFF
--- a/app/components/global/AppBar/AppBar.js
+++ b/app/components/global/AppBar/AppBar.js
@@ -46,28 +46,25 @@ const DesktopView = styled(Flex)`
   `};
 `
 
-const AppBar = ({ user, userMenuItems, defaultMenuItems }) => {
-  const menuItems = [...(user ? userMenuItems : []), ...defaultMenuItems]
-  return (
-    <AppBarContainer>
-      <BurgerMenu menuItems={menuItems} />
-      <LogoLink href="https://elifesciences.org">
-        <img alt="eLife" src="/assets/elife-logo.png" />
-      </LogoLink>
-      <DesktopView>
-        <Separator />
-        <Box flex="1 1 auto">
-          {menuItems &&
-            menuItems.map(item => (
-              <AppBarLink key={item.link} to={item.link}>
-                {item.label}
-              </AppBarLink>
-            ))}
-        </Box>
-      </DesktopView>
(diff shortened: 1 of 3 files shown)


app/components/global/AppBar/AppBar.js
```diff
@@ -46,24 +46,28 @@ const DesktopView = styled(Flex)`
...
-            Dashboard
-          </AppBarLink>
-        )}
-      </Box>
-    </DesktopView>
-    <ProfileMenu user={user} />
-  </AppBarContainer>
-)
+const AppBar = ({ user, userMenuItems, defaultMenuItems }) => {
+  const menuItems = [...(user ? userMenuItems : []), ...defaultMenuItems]
```
damnhipster: Makes more sense to default `userMenuItems` in the params, i.e.
`const AppBar = ({ user, userMenuItems = [], defaultMenuItems }) => { ... }`

damnhipster: It's confusing to check for the existence of `user` before we use `userMenuItems` given the parameters we're passing in. Instead we can pass in both within a single object as the first parameter, i.e.
```
const AppBar = ({ user, defaultMenuItems }) => {
  const { currentUser, userMenuItems = [] } = user
  const menuItems = [...userMenuItems, ...defaultMenuItems]
  ...
}
```

Cooryd: This wouldn't work with how the logic is currently as that would require me changing 
```suggestion
  const menuItems = [...(user && userMenuItems), ...defaultMenuItems]
```
which would break when the user isn't signed in.

damnhipster: Right, I would suggest changing it to 
```suggestion
  const menuItems = [...(loggedIn ? userMenuItems : []), ...defaultMenuItems]
```

Issue #877, opened 2018-10-25
Implement REST endpoint /ping


diversemix:
Need to implement this endpoint on GET `/ping` for the health check to `return 200 "pong"`  ... when everything is healthy.

Also need to remove its implementation here: `elife-xpub-formula/salt/elife-xpub/config/etc-nginx-sites-enabled-xpub.conf`

diversemix:
@giorgiosironi  FYI ... @hdrury1 This really needs to go in ASAP

giorgiosironi:
The formula implementation is a stop gap, as it allows a server to be used as long as nginx is running. An implementation in the app will be more precise as it will make sure the container is running, and probably try to connect to the database. By the way this won't block the multiple servers expansion.

PR #894, opened 2018-10-26
feat: create a new rest endpoint ping

jyenes:
closes: #877

### DIFF
--- a/server/client-config/routes.js
+++ b/server/client-config/routes.js
@@ -1,3 +1,4 @@
+const logger = require('@pubsweet/logger')
 const config = require('config')
 const { pickBy } = require('lodash')
 const DBExists = require('@pubsweet/db-manager/src/helpers/db-exists.js')
@@ -19,6 +20,13 @@ const s3 = new AWS.S3({
   apiVersion: '2006-03-01',
 })
 
+const sendResult = (res, statusCode, body) => {
+  if (statusCode !== 200) {
+    logger.error(`HealthCheckError: ${body}`)
+  }
+  res.send(statusCode, body)
+}
+
 module.exports = app => {
   app.get('/config.js', (req, res) => {
     const clientConfig = pickBy(config, value => value.isPublic)
@@ -31,15 +39,15 @@ module.exports = app => {
     try {
       const exists = await DBExists()
       if (!exists) {
-        res.send(404, 'database error')
+        sendResult(404, 'database error')
       }
       const { Body } = await s3.getObject({ Key: 'test-file' }).promise()
       if (!Body) {
-        res.send(404, 's3 connection error')
+        sendResult(404, 's3 connection error')
       }
-      res.send('pong')
+      sendResult(200, 'pong')
     } catch (error) {
-      res.send(500, 'ko')
+      sendResult(500, error)
     }
   })
 }
--- a/server/client-config/routes.js
+++ b/server/client-config/routes.js
@@ -12,4 +12,8 @@ module.exports = app => {
     res.type('js')
     res.send(response)
   })
+
+  app.get('/ping', (req, res) => {
+    res.send('pong')
+  })
 }
--- a/server/client-config/routes.js
+++ b/server/client-config/routes.js
@@ -40,11 +40,14 @@ module.exports = app => {
       const exists = await DBExists()
       if (!exists) {
         sendResult(404, 'database error')
+        return
       }
       const { Body } = await s3.getObject({ Key: 'test-file' }).promise()
       if (!Body) {
         sendResult(404, 's3 connection error')
+        return
       }
+
       sendResult(200, 'pong')
     } catch (error) {
       sendResult(500, error)
--- a/server/client-config/routes.js
+++ b/server/client-config/routes.js
@@ -14,8 +14,8 @@ const nocache = (req, res, next) => {
 }
 
 const s3 = new AWS.S3({
-  ...config.get('aws.credentials'),
-  ...config.get('aws.s3'),
+  ...config.aws.credentials,
+  ...config.aws.s3,
   apiVersion: '2006-03-01',
 })
 
--- a/server/client-config/routes.js
+++ b/server/client-config/routes.js
@@ -4,6 +4,13 @@ const { pickBy } = require('lodash')
 const DBExists = require('@pubsweet/db-manager/src/helpers/db-exists.js')
 const AWS = require('aws-sdk')
 
+const FILE_HEALTH_CHECK = 'healtcheck-file'
+const DATABASE_ERROR = 'Database Error'
+const S3_CONNECTION_ERROR = 'S3 Connection Error'
+const DEFAULT_ERROR = 'Internal Server Error'
+const SUCCESFULL_RESPONSE = 'pong'
+const S3_API_VERSION = '2006-03-01'
+
 const template = clientConfig => `
   window.config = ${JSON.stringify(clientConfig)}
 `
@@ -17,7 +24,7 @@ const nocache = (req, res, next) => {
 const s3 = new AWS.S3({
   ...config.aws.credentials,
   ...config.aws.s3,
-  apiVersion: '2006-03-01',
+  apiVersion: S3_API_VERSION,
 })
 
 const sendResult = (res, statusCode, body) => {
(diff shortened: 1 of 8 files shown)


server/client-config/routes.js
```diff
@@ -12,4 +12,8 @@ module.exports = app => {
     res.type('js')
     res.send(response)
   })
+
+  app.get('/ping', (req, res) => {
+    res.send('pong')
```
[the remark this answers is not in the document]
diversemix: There is an argument that it should try and contact `journal` and `sciencebeam` too ... the latter being not critical and the former would be ok if you are still logged.

This is just a record of thoughts - not a request

server/client-config/routes.js
```diff
@@ -12,4 +26,20 @@ module.exports = app => {
...
+      if (!exists) {
+        res.send(404, 'database error')
+      }
+      const { Body } = await s3.getObject({ Key: 'test-file' }).promise()
+      if (!Body) {
+        res.send(404, 's3 connection error')
+      }
+      res.send('pong')
+    } catch (error) {
+      res.send(500, 'ko')
```
damnhipster: It's worth logging at this point using `logger.error`

diversemix:
@jyenes ... I made a change to do the logging. I was in the middle of something else - hence the silly mistakes. Let me know what you think.

server/client-config/routes.js
```diff
@@ -12,4 +34,23 @@ module.exports = app => {
...
   })
+
+  app.get('/ping', nocache, async (req, res) => {
+    try {
+      const exists = await DBExists()
+      if (!exists) {
+        sendResult(res, 404, 'database error')
+        return
+      }
+      const { Body } = await s3.getObject({ Key: 'test-file' }).promise()
```
diversemix: @jyenes Should we call the `test-file` something more descriptive to avoid it getting deleted? For example `healthcheck-file` ?

jyenes: agree

server/client-config/routes.js
```diff
@@ -12,4 +34,23 @@ module.exports = app => {
...
+      }
+      const { Body } = await s3.getObject({ Key: 'healtcheck-file' }).promise()
+      if (!Body) {
+        sendResult(res, 404, 's3 connection error')
+        return
+      }
+
+      sendResult(res, 200, 'pong')
+    } catch (error) {
+      sendResult(res, 500, error)
```
damnhipster: This would send the stack trace back as a response, which we should avoid as it can be a security risk.

We should return back `Internal Server Error` and log the error.

server/client-config/routes.js
```diff
@@ -1,9 +1,31 @@
+const logger = require('@pubsweet/logger')
 const config = require('config')
 const { pickBy } = require('lodash')
+const DBExists = require('@pubsweet/db-manager/src/helpers/db-exists.js')
```
tamlyn: Note that `dbExists` is exported as a function from the package root so you can avoid a fragile deep require.

server/client-config/routes.js
```diff
@@ -12,4 +34,23 @@ module.exports = app => {
...
+      const exists = await DBExists()
+      if (!exists) {
+        sendResult(res, 404, 'database error')
+        return
+      }
+      const { Body } = await s3.getObject({ Key: 'healtcheck-file' }).promise()
+      if (!Body) {
+        sendResult(res, 404, 's3 connection error')
+        return
+      }
```
tamlyn: With this setup, if the database it down, you don't get to know anything about S3. As and when more checks are added, this could be a problem. Another way to structure this could be to check all the things and then report the status of each in the response (with an HTTP error status if any one of them has an error).

server/client-config/routes.js
```diff
@@ -12,4 +34,23 @@ module.exports = app => {
     res.type('js')
     res.send(response)
   })
+
+  app.get('/ping', nocache, async (req, res) => {
```
tamlyn: I'm not sure that the `client-config` package is the right place for this. I appreciate that having the server code split into packages like this makes it tricky to know where to put little things like this. I would probably put it in `xpub-server`.

server/client-config/routes.test.js
```diff
@@ -36,4 +62,10 @@ describe('Client config handler', () => {
     const response = await request.get('/config.js').expect(200)
     expect(response.text).toMatchSnapshot()
   })
+
+  it('returns something pointing ping', async () => {
+    const request = makeApp()
+    const response = await request.get('/ping').expect(200)
+    expect(response.text).toMatchSnapshot()
```
tamlyn: Snapshots are useful for matching against complex values which are expected to change and need updating. (For instance the client config object in the test above is likely to change in future as more values are added.) Here it would be better to use `.toEqual` as that puts the expected value right in the test.

server/client-config/routes.test.js
```diff
@@ -36,4 +62,10 @@ describe('Client config handler', () => {
     const response = await request.get('/config.js').expect(200)
     expect(response.text).toMatchSnapshot()
   })
+
+  it('returns something pointing ping', async () => {
```
tamlyn: Could we also have tests for the failure cases?

server/client-config/routes.js
damnhipster: Agreed, given that it's `xpub-server` which is dependant on the database and S3.

jyenes:
@tamlyn can you recheck the changes?

diversemix:
@damnhipster could you review this too? (As I worked on it with Javier)

server/xpub-server/routes.js
```diff
@@ -0,0 +1,67 @@
...
+module.exports = app => {
+  app.get('/ping', nocache, async (req, res) => {
+    const errors = []
+    let statusCode = 200
+    let body = SUCCESFULL_RESPONSE
+    let response
+    try {
+      response = await Promise.all([
+        dbExists(),
+        s3.getObject({ Key: FILE_HEALTH_CHECK }).promise(),
```
damnhipster: We may as well `putObject` so we don't have to manually add the file to S3. And then run `getObject` afterwards, just as a sanity check.

server/xpub-server/routes.js
```diff
@@ -0,0 +1,67 @@
...
+      // only log changes in state to not flood the logs
+      logger.error(`HealthStats: ${thisHealthStatus}, Errors: ${errors}`)
+      lastHealthStatus = thisHealthStatus
+    }
+
+    if (errors.length > 0) {
+      body = errors
+      if (statusCode === 200) {
+        // this is http gone
+        statusCode = 410
```
damnhipster: Not sure if `410` is the correct response. If there are errors where the server is unable to correctly communicate with it's dependencies or the dependencies have changed their responses, that's surely still an `Internal Server Error`.

server/xpub-server/routes.js
jyenes: not sure if this is a good idea, the ping endpoint is being hit every 5 min, so putting an object will require to delete the object as well, maybe we are overloading this endpoint once actually what want is just check if we have connectivity

giorgiosironi: 5 _seconds_

giorgiosironi: I would favor the database check as it's an absolute dependency, but I'd skip the S3 one in favor of `/ping` working reliably across environments.

server/xpub-server/routes.js
```diff
@@ -0,0 +1,67 @@
...
+  apiVersion: S3_API_VERSION,
+})
+
+let lastHealthStatus = null
+
+module.exports = app => {
+  app.get('/ping', nocache, async (req, res) => {
+    const errors = []
+    let statusCode = 200
+    let body = SUCCESFULL_RESPONSE
```
giorgiosironi: `SUCCESSFUL`?

damnhipster: Full of success

server/xpub-server/routes.js
damnhipster: The S3 client will let you overwrite silently by default, but I agree with your point about keeping the response times low on this endpoint.

diversemix: Yes - let's keep this as-is

server/xpub-server/routes.js
diversemix: As long as its not `20x` and is `4xx` I don't care for the time being :)

server/xpub-server/routes.js
jyenes: :) misspelling fixed

Issue #854, opened 2018-10-22
fix spacing between dashboard items & text below


jsms90:
Linux OS

Firefox:
![image](https://user-images.githubusercontent.com/15656538/47306004-1fc4f100-d623-11e8-95a8-341d9681ab97.png)

Chrome:
![image](https://user-images.githubusercontent.com/15656538/47306050-3e2aec80-d623-11e8-80a6-8fd899979d3b.png)

jyenes:
not able to reproduce the issue @jsms90 in both browsers look good to me

Cooryd:
<img width="1043" alt="screenshot 2018-10-26 at 14 43 24" src="https://user-images.githubusercontent.com/22455561/47570265-a59bb180-d92d-11e8-89fa-f4d0d3e36c35.png">

This is on osx in chrome. I think it has something to do with having multi-lined title dashboard item.

## To recreate
Create a submission with title 'Shearing in flow environment promotes evolution behaviour in microbial populations'. Then create 3 new submissions with no title

PR #895, opened 2018-10-26
style(Dashboard): fix text in the bottom of the dashboard

jyenes:
closes: #854

### DIFF
--- a/app/components/pages/Dashboard/index.js
+++ b/app/components/pages/Dashboard/index.js
@@ -21,6 +21,7 @@ const BoxHiddenFromMobile = styled(Box)`
 
 const DashboardPanel = styled(Tabs.Panel)`
   min-height: 300px;
+  margin-bottom: 12px;
 `
 
 const MobileOnlyStickyFooter = styled(StickyFooter)`
@@ -29,7 +30,6 @@ const MobileOnlyStickyFooter = styled(StickyFooter)`
 `
 
 const CenterdSmallParagraph = styled(SmallParagraph)`
-  margin-top: 12px;
   color: ${th('colorTextSecondary')};
   text-align: center;
 `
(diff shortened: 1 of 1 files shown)


app/components/pages/Dashboard/index.js
```diff
@@ -29,6 +29,7 @@ const MobileOnlyStickyFooter = styled(StickyFooter)`
 `
 
 const CenterdSmallParagraph = styled(SmallParagraph)`
+  margin-top: 12px;
```
jsms90: As a general rule, we put `margin-bottom` on components wherever possible. We should record this somewhere. You weren't to know.

This is because combining `margin-bottom` on a higher-up component & `margin-bottom` on a lower-down component would result in margin collapse. So the unwritten rule of using only margin-bottom for vertical spacing avoids this scenario.

Could we pop `margin-bottom on the `DashboardPanel` instead?

jyenes: yes, sure!

app/components/pages/Dashboard/index.js
```diff
@@ -21,6 +21,7 @@ const BoxHiddenFromMobile = styled(Box)`
 
 const DashboardPanel = styled(Tabs.Panel)`
   min-height: 300px;
+  margin-bottom: 12px;
```
jsms90: Oh. Sorry. This isn't actually the right component to add it to. This is adding an extra bottom margin to each dashboard list item.

I wasn't paying attention to the code. I just assumed from the name that `DashboardPanel` meant the entire list.

It should be on this component: https://github.com/elifesciences/elife-xpub/pull/895/files#diff-ddd4166fe5ead236eb34c57ce9d7b2c2R67

Do you see what I mean?

PR #896, opened 2018-10-26
886 manuscript status updates

### LINKED ISSUE #886: Copy change: 'Waiting for Decision' to 'Submitted'
Mentioned in our last in-person meeting and/or standup

**Designs**:
From Zeplin:
image

---

While we're at it, we should probably also put each of the different views of this component into styleguidist

jsms90:
#### What does this PR do?
- Change all instances of 'waiting for decision' to 'submitted'
- Add all different states of `DashboardListItem` to styleguidist:
  - different titles
  - each of 3 manuscript statuses
  - different dates, to show the various outputs (today, yesterday, x days ago)
- Changed `title` to required prop in DashboardListItem, since it's parent (DashboardList) is actually the one responsible for deciding what title gets sent to it (i.e. decides whether it says 'untiltled' or not)

#### Any relevant tickets
fixes #886 

Existing test altered

### DIFF
--- a/app/components/ui/molecules/DashboardListItem.js
+++ b/app/components/ui/molecules/DashboardListItem.js
@@ -85,12 +85,8 @@ const DashboardListItem = ({ statusCode, title, date }) => (
 
 DashboardListItem.propTypes = {
   statusCode: PropTypes.string.isRequired,
-  title: PropTypes.string,
+  title: PropTypes.string.isRequired,
   date: PropTypes.instanceOf(Date).isRequired,
-}
-
-DashboardListItem.defaultProps = {
-  title: 'Untitled manuscript',
 }
 
 export default DashboardListItem
--- a/app/components/ui/molecules/DashboardListItem.md
+++ b/app/components/ui/molecules/DashboardListItem.md
@@ -1,10 +1,43 @@
-List item in the dashboard that links to manuscript. It shows status (e.g. accepted, submitted), manuscript title and date since status last changed.
+List item in the dashboard. It shows the manuscript title, submission status & the date of the last status change.
 
 ```js
+const today = new Date()
 ;<DashboardListItem
   statusCode={'CONTINUE_SUBMISION'}
   title={
     'Shearing in flow environment promotes evolution of social behaviour in microbial populations'
+  }
+  date={today}
+/>
+```
+
+```js
+const today = new Date()
+;<DashboardListItem
+  statusCode={'SUBMITTED'}
+  title={
+    'Cross-talk between PRMT1-mediated methylation and ubiquitylation on RBM15 controls RNA splicing'
+  }
+  date={new Date(today.setDate(today.getDate() - 1))}
+/>
+```
+
+```js
+const today = new Date()
+;<DashboardListItem
+  statusCode={'SUBMITTED'}
+  title={
+    'INAVA-ARNO complexes bridge mucosal barrier function with inflammatory signaling'
+  }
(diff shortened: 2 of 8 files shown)


app/components/ui/molecules/DashboardListItem.md
```diff
@@ -1,11 +1,44 @@
-List item in the dashboard that links to manuscript. It shows status (e.g. accepted, waiting approval), manuscript title and date since status last changed.
```
jsms90: FYI, I've changed the description slightly because this component doesn't know anything about linking to a manuscript - this logic is handled in `DashboardList`

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -85,12 +85,8 @@ const DashboardListItem = ({ statusCode, title, date }) => (
 
 DashboardListItem.propTypes = {
   statusCode: PropTypes.string.isRequired,
-  title: PropTypes.string,
+  title: PropTypes.string.isRequired,
   date: PropTypes.instanceOf(Date).isRequired,
 }
 
-DashboardListItem.defaultProps = {
-  title: 'Untitled manuscript',
```
jsms90: To be honest, this feels to me like something that the `DashboardListItem` should be handling, rather than the `DashboardList`. What do you think @Cooryd?

How about if we passed the whole manuscript object down from `DashboardList` into `DashboardListItem` & allowed this component to be responsible for rendering appropriately?

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -85,12 +85,8 @@ const DashboardListItem = ({ statusCode, title, date }) => (
 
 DashboardListItem.propTypes = {
   statusCode: PropTypes.string.isRequired,
-  title: PropTypes.string,
+  title: PropTypes.string.isRequired,
```
Cooryd: Does this cause any problems for submissions before they've reached the upload manuscript stage? Is there no title data present or does gql return an empty string on the index?

app/components/ui/molecules/DashboardListItem.js
Cooryd: This makes sense to me, would certainly make more sense to convert the iso date to a Date object within the `DashboardListItem`

jsms90:
@Cooryd happy for this to go in? I've raised a new issue for the slight refactor #898

app/components/ui/molecules/DashboardListItem.md
```diff
@@ -1,11 +1,44 @@
...
+const today = new Date()
 ;<DashboardListItem
   statusCode={'CONTINUE_SUBMISION'}
   title={
     'Shearing in flow environment promotes evolution of social behaviour in microbial populations'
   }
+  date={today}
+/>
+```
+
```
damnhipster: The list items don't appear on the styleguide other than the first one. 

Take a look at the [FileUpload.md](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/ui/molecules/FileUpload.md) for the correct usage.

jsms90: I'm surprised that not having text in between code blocks would make a difference.

For the file upload, they are distinctly different states & therefore easily labelled as something different. I'm not sure what text I would put at the top of these, other than just re-iterating what you see below ('Different title, status = Submitted, today's date ?) :woman_shrugging:

jsms90: Now that the styleguide PRs have been merged, seems to be fine to me:
image

PR #902, opened 2018-10-26
refactor: split manuscript resolvers into separate files

tamlyn:
#### Background

- Split large resolvers out into separate files to make them more manageable

~~WIP because of an interesting issue that this refactor has exposed:~~
- we explicitly close the pubsub database connection so that the test run exits
- we don't await the return value from the upload progress subscription
- because the messages go via the database, they sometimes arrive after the test has completed
- when this happens, the database connection has already been closed so it causes the *next* Jest test suite to error in weird ways
- this wasn't an issue before because the upload tests were not last in the test file so the messages always arrived before the suite ended and the connection was closed


tamlyn:
Waiting on https://gitlab.coko.foundation/pubsweet/pubsweet/merge_requests/441

giorgiosironi:
The `test` job is stuck on Jenkins saying:
```
Test Suites: 43 passed, 43 total
Tests:       4 skipped, 186 passed, 190 total
Snapshots:   3 passed, 3 total
Time:        80.211s
Ran all test suites in 2 projects.
Jest did not exit one second after the test run has completed.
This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```
(from https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Felife-xpub/detail/PR-902/2/pipeline)

How does that compare with the output on Gitlab, which is green?

damnhipster:
/i'.ve managed to reproduce the problem on Gitlab [here[(https://gitlab.com/elifesciences/elife-xpub/-/jobs/116547120). It seems to be intermittent.

tamlyn:
Hmmm. Interesting. I've also reproduced it locally using the Docker image. Will investigate.

tamlyn:
Mocked out the offending module. Pipelines now completing locally, on Gitlab and on Jenkins.

server/xpub-server/entities/manuscript/resolvers/index.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+describe('Manuscripts', () => {
+  const profileId = userData.identities[0].identifier
+  const badProfileId = badUserData.identities[0].identifier
+  let userId
+
+  beforeEach(async () => {
+    replaySetup('success')
+    await Promise.all([
+      fs.remove(config.get('pubsweet-server.uploads')),
```
damnhipster: Do we still need this?

PR #927, opened 2018-10-30
897 - AppBar fixes

Cooryd:
#### Background

General tidy-up and fix PR for the `AppBar` component to address the following issues:

- Wrapping issue with nav menu at < 663px before it gets to mobile display
- Vertical alignment issue with `BurgerMenu` icon 
- Too much space between eLife logo and nave menu on desktop screen size.
- No PropTypes on `AppBar` component
- No unit tests for `AppBar` component

What does this PR do?

- Change the breakpoint at which the `BurgerMenu` displays\
- Vertically center the `BurgerMenu` icon within the `AppBar`
- Change `AppBar` to `justify-content: left;` for desktop display and `margin-left:auto;` the `ProfileMenu` to right align it.
- Add PropTypes to `AppBar`
- Add tests for `AppBar`

#### Any relevant tickets

closes#897

Visually
Unit test


hdrury1:
I would be tempted to remove the `margin` on the logo altogether... what do you think @chugginselifesciences?

With margin:

screen shot 2018-11-01 at 13 45 49

Without margin:

screen shot 2018-11-01 at 13 45 36

app/components/global/AppBar/AppBar.test.js
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+describe('AppBar', () => {
+  it('Only receives default menu links when no user object passed', async () => {
+    const wrapper = makeWrapper(defaultProps, currentUserMock({}))
+    await new Promise(resolve => {
+      setTimeout(resolve, 0)
+    })
+    wrapper.update()
+    expect(wrapper.find(AppBar).props().menuItems).toHaveLength(3)
```
jsms90: Correct me if I've misunderstood. But this seems like it's testing `index.js` rather than `AppBar.js`?

Specifically this code:
https://github.com/elifesciences/elife-xpub/blob/4e1ca9b68028091362a95d59bb255e30a3aeed7d/app/components/global/AppBar/index.js#L11-L13

because these tests are checking the length of the `menuItems` prop, based on the `user` prop.

If I'm understanding correctly, I imagine that this file should be `index.test.js` rather than `AppBar.test.js`? You can also see this from the fact that [your `makeWrapper` function is building `<AppBarWrapper>`](https://github.com/elifesciences/elife-xpub/pull/927/commits/89cf0b19cf5e35546212306de58a05cde9387e2c#diff-6372abc70d14d6e8c893cb0f7b83fc6fR57)

---

One test that you could write for `AppBar` itself, would be to check whether the appropriate amount of `NavLink`s are rendered, based on props?

But essentially [that mapping](https://github.com/elifesciences/elife-xpub/pull/927/files#diff-db536dc0676835888e436d994badfc32R62) is the only logic built into AppBar itself, so I'm not sure there are actually many tests we could write for this component?

app/components/global/AppBar/AppBar.md
```diff
@@ -3,11 +3,24 @@ Overall navigation bar for the app
...
 
 ```js
-<AppBar user="anyone" />
+<AppBar
+  menuItems={[
+    { label: 'Author guide', link: '/author-guide' },
+    { label: 'Reviewer guide', link: '/reviewer-guide' },
+    { label: 'Contact us', link: '/contact-us' },
+  ]}
+  user="anyone"
```
jsms90: Is it possible for us to pass in a dummy user object, so that the ProfileMenu within the `AppBar` component in styleguidist is functional?

Right now, this doesn't conform to your new PropTypes definition, so we're getting console errors:
image

and we're also not seeing the 'Dashboard' link appear.

app/components/global/AppBar/AppBar.js
```diff
@@ -71,4 +70,18 @@ const AppBar = ({ user, menuItems }) => (
...
 )
 
+AppBar.propTypes = {
+  menuItems: PropTypes.arrayOf(
+    PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      link: PropTypes.string.isRequired,
+    }),
+  ).isRequired,
+  user: PropTypes.shape({}),
```
jsms90: `PropTypes.object`?

Perhaps it would also be an idea to add PropTypes to the `ProfileMenu` component within this PR, for greater specificity on what shape this object needs to take?

app/components/global/AppBar/AppBar.js
```diff
@@ -42,16 +42,15 @@ const Separator = styled.div`
...
-  width:80%
   align-items: center;
   display: none;
-  ${media.mobileUp`
+  ${media.tabletPortraitUp`
     display: flex;
   `};
 `
 
 const AppBar = ({ user, menuItems }) => (
```
jsms90: Might be good to add a destructuring assignment on `user` that matches your defaultProps below?

chugginselifesciences:
@hdrury1 I'd intended to have equal spacing between the logo, the divider and the "Author guide" text, so would prefer to keep the margin. Should also be the same as the space between the left edge and the logo.

app/components/global/AppBar/AppBar.js
```diff
@@ -71,4 +70,18 @@ const AppBar = ({ user, menuItems }) => (
...
+    PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      link: PropTypes.string.isRequired,
+    }),
+  ).isRequired,
+  user: PropTypes.shape({}),
+}
+
+AppBar.defaultProps = {
+  user: null,
```
jsms90: What would you think of `user = {}` rather than `user = null`?

hdrury1:
At the moment it looks as though it's a lot further from the left edge than from the divider.

app/components/global/AppBar/AppBar.js
Cooryd: We're using the `react/forbid-prop-types` linting rule which prevents `any` `array` and `object` as propTypes. I suppose I should be looking at the user values being applied within the `ProfileMenu` component and defining the minimum required attributes of the object if it exists here.

app/components/global/AppBar/AppBar.js
Cooryd: The `ProfileMenu` component checks if user exists before displaying either a sign in link or the profile menu. Defaulting it to an object would require these checks to be changed to something like `Object.keys(user).length` which if quite a bit more verbose. Would there be any benefit of casting an empty object to an undefined prop?

app/components/global/AppBar/index.test.js
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+describe('AppBarWrapper', () => {
+  it('Only receives default menu links when no user object passed', async () => {
+    const wrapper = makeWrapper(defaultProps, currentUserMock({}))
+    await new Promise(resolve => {
+      setTimeout(resolve, 0)
+    })
+    wrapper.update()
+    expect(wrapper.find(AppBar).props().menuItems).toHaveLength(3)
```
jsms90: How about `const numDefaultLinks = defaultProps.defaultMenuItems.length` above
and then here: `.toHaveLength(numDefaultProps)`?

Avoids hard-coding, so you could change the props you want to pass in above & this test will still pass. It also then make it super explicit what's being checked here i.e. why "3".

app/components/global/AppBar/index.test.js
```diff
@@ -0,0 +1,82 @@
...
+  })
+
+  it('Receives both default and user menu items when user object passed', async () => {
+    const wrapper = makeWrapper(defaultProps, currentUserMock(mockUserData))
+
+    await new Promise(resolve => {
+      setTimeout(resolve, 0)
+    })
+    wrapper.update()
+    expect(wrapper.find(AppBar).props().menuItems).toHaveLength(4)
```
jsms90: Similar to above `const numOverallLinks` (or a better variable name :sweat_smile: )

app/components/global/AppBar/AppBar.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const menuConfig = [
+      { label: 'a', link: '/a' },
+      { label: 'b', link: '/b' },
+      { label: 'c', link: '/c' },
+    ]
+    const wrapper = makeWrapper({ user: {}, menuItems: menuConfig })
+
+    menuConfig.forEach((config, index) => {
+      const link = wrapper.find(AppBarLink).at(index)
+      expect(link.props().to).toEqual(config.link)
```
jsms90: could we maybe call it `menuItems` instead of `menuConfig`? And then each one `item` instead of `config`?

At first, I thought this was referring to the actual config files. Partly because I know that we have our `Logout` link in the ProfileMenu component, getting its link from this config file.

What do you reckon?

app/components/global/AppBar/AppBar.js
```diff
@@ -77,11 +77,6 @@ AppBar.propTypes = {
       link: PropTypes.string.isRequired,
     }),
   ).isRequired,
-  user: PropTypes.shape({}),
```
jsms90: `user` is still a prop that's getting passed into this component, so I wouldn't suggest removing PropTypes altogether. Just clarifying the type?

jsms90: We could always raise a new issue for this, rather than fixing in this PR?

Cooryd: I'll leave this out of this PR for now. I'm unsure what the correct format for this should be as the linter blocks the generic `object` type and this component doesn't really need to know anything about the `user` as it just forwards it on to the `ProfileMenu`. I was reluctant to remove this as I've always thought that if using propTypes in a component then all props should be assigned but it also felt redundant adding the code for the user shape to a component that doesn't care about its format.

jsms90: Yeah I totally get what you mean. But I reckon, if we're going to use PropTypes, let's use it consistently everywhere. I think it's really good that you've been adding it to files that are missing them. As for this, specifically, I reckon it's worth duplicating the specific shape of the user object, both here in this file & in the ProfileMenu file.

When we start seeing enough areas of duplication, that might be time to refactor out into a generic types file, to be imported as and when needed. I'm not sure that we're there yet. But ensuring that we have PropTypes on all files will be a good way to see when we've got that level of duplication.

And yeah, sorry, I didn't realise the linter blocks you from just saying `object`. But I think that's an attempt to get you to specify the _shape_ of the object, perhaps?

So just like [line 74](https://github.com/elifesciences/elife-xpub/pull/927/files/81a6776b624e4eb17cb2927d4012ad6187323bdd..ff9a8a42a972891b810d0c5398b5a161403f8b90#diff-db536dc0676835888e436d994badfc32R74), you would do
```js
PropTypes.objectOf(
  ProTypes.shape({
    identities
    .... etc (whatever else needs to be on here - haven't looked)
  })
})
```

Issue #921, opened 2018-10-30
Contact Us static page tweaks


hdrury1:
Ideally, the grid of people should drop down to one column at the smallest breakpoint and the photos should have a consistent height. See https://elifesciences.org/about/people/staff for reference.

Also see https://github.com/elifesciences/elife-xpub/issues/889 for how these images should be used.

Issue #889, opened 2018-10-26
Sort out image URLs


diversemix:
see comment on #884 
@thewilkybarkid comment:

> So curl -v https://api.elifesciences.org/people/15fb82ec | jq .image.uri gives https://iiif.elifesciences.org/journal-cms/person%2F2017-07%2Fsusanna.jpg (the IIIF base URI). So https://iiif.elifesciences.org/journal-cms/person%2F2017-07%2Fsusanna.jpg/full/full/0/default.jpg is the image at full size.
>
> There's focal point information there too; Journal has to calculate the region, take a look at https://github.com/elifesciences/journal/blob/develop/src/Helper/IiifRegionCalculator.php (and the images on https://elifesciences.org/about/people/staff too, this is where the parameters for https://iiif.elifesciences.org/journal-cms/person%2F2017-07%2Fsusanna.jpg/0,53,1000,566/502,284/0/default.webp come from).

hdrury1:
Done in #937

PR #937, opened 2018-10-31
refactor(StaffPages): modify responsive to one column on mobile view

jyenes:
#### Background
modify responsive to one column on mobile view, and changed pictures to have the same aspect ratio.

#### Any relevant tickets
closes: #921

### DIFF
--- a/app/components/ui/molecules/StaffGrid.js
+++ b/app/components/ui/molecules/StaffGrid.js
@@ -4,7 +4,7 @@ import { Flex, Box } from 'grid-styled'
 const StaffGrid = ({ children, ...props }) => (
   <Flex flexWrap="wrap" mb={-3} mx={-2} {...props}>
     {children.map((item, index) => (
-      <Box key={item.key || index} mb={3} px={2} width={[1 / 2, 1 / 2, 1 / 3]}>
+      <Box key={item.key || index} mb={3} px={2} width={[1 / 1, 1 / 2, 1 / 3]}>
         {item}
       </Box>
     ))}
--- a/app/components/ui/molecules/StaffGrid.js
+++ b/app/components/ui/molecules/StaffGrid.js
@@ -4,7 +4,7 @@ import { Flex, Box } from 'grid-styled'
 const StaffGrid = ({ children, ...props }) => (
   <Flex flexWrap="wrap" mb={-3} mx={-2} {...props}>
     {children.map((item, index) => (
-      <Box key={item.key || index} mb={3} px={2} width={[1 / 1, 1 / 2, 1 / 3]}>
+      <Box key={item.key || index} mb={3} px={2} width={[1, 1 / 2, 1 / 3]}>
         {item}
       </Box>
     ))}
--- a/app/components/pages/StaticPages/ContactUs/EditorialStaff.js
+++ b/app/components/pages/StaticPages/ContactUs/EditorialStaff.js
@@ -15,31 +15,31 @@ const EditorialStaff = () => (
       <StaffCard
         jobTitle="Editorial Manager"
         name="Wei Mun Chan"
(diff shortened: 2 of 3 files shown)


jsms90:
@jyenes closes #889 as well?

For some reason I can't see the pictures on firefox, but can on chrome. Does this happen to anyone else, or is it just my machine?

app/components/ui/molecules/StaffGrid.js
```diff
@@ -4,7 +4,7 @@ import { Flex, Box } from 'grid-styled'
 const StaffGrid = ({ children, ...props }) => (
   <Flex flexWrap="wrap" mb={-3} mx={-2} {...props}>
     {children.map((item, index) => (
-      <Box key={item.key || index} mb={3} px={2} width={[1 / 2, 1 / 2, 1 / 3]}>
+      <Box key={item.key || index} mb={3} px={2} width={[1 / 1, 1 / 2, 1 / 3]}>
```
jsms90: This can just be `1` rather than `1 / 1`? Ever so slightly easier to read.

@hdrury1 @chugginselifesciences, just to clarify, what is meant by "at the smallest breakpoint" in issue #921?

This is what we have right now:
- one column < 480
- 480 <= two columns < 768

Is that right?

hdrury1: @jsms90 Yes.

hdrury1:
Yup, images not loading in Firefox or Safari.

jyenes:
@jsms90 @hdrury1 there was an issue with the image format, now has been sorted out.

hdrury1:
LGTM

jyenes:
@hdrury1 I was checking #884 and it seems I'm using correctly the iiif CDN, @thewilkybarkid can you confirm the method I have used is correct?

thewilkybarkid:
It's now using IIIF images rather than the source image, which is better... but is still hard-coded data.

jyenes:
@thewilkybarkid with hardcoded data you mean the cdn parameters? I'm using those to get the same width and height in all pictures if we use `full/full/0/default.jpg`
then we are having issues becasue the images have different sizes, here is an example

screenshot 2018-10-31 at 17 39 17


that's why I'm using `0,17,1000,566/502,284/0/default.jpg` 

`0,17,1000,566` keeps the aspect ratio 
`502,284` this is the width and height I'm requiring

using those parameters the page looks like this:
screenshot 2018-10-31 at 17 50 02


using `full/full/0/default.jpg`
screenshot 2018-10-31 at 17 51 50

diversemix:
@jyenes I think the idea is to use the endpoint `https://api.elifesciences.org/people?type=executive` (which we already do in the people picker) - this returns a list of people you can query / display and their associated image urls

thewilkybarkid:
Yep. `person%2F2017-07%2Fweimun.jpg`, for example, is an image identifier. If the picture is ever replaced, it will change. The CDN will hold on to it for a while (we have a `max-age` of 1 year), but at some point it will disappear.

(I'm assuming that the CMS cleans up old images. But if it doesn't that shouldn't be relied on.)

thewilkybarkid:
Same applies to the people themselves of course (it's correct now, but is guaranteed to change).

hdrury1:
I suggest we get this in as is and raise a new ticket to implement dynamic updates. Ideally the page should update whenever we add a new staff member to journal cms but this needs investigating as (I believe) we don’t currently categorise staff by department anyway.

jyenes:
I see the point of using the people API, but I have some doubts is the people API queryable by staff type? I mean something like this` https://api.elifesciences.org/people?type=editorial-staff` which will return the people we need to fill the content in that page, otherwise, we can not make it dynamic, so we need to change that page, in any case, the person change.

diversemix:
Agreed. @hdrury1 / @jyenes ... but this can be a tidy up after launch.

Issue #908, opened 2018-10-29
Introduce min-height to Dashboard


hdrury1:
Latest changes to the dashboard have removed the min-height from above the "Can't find a submission..." text which makes it difficult to distinguish from the text above when the tab is empty:

![screen shot 2018-10-29 at 14 32 55](https://user-images.githubusercontent.com/17023702/47657753-8d27d300-db89-11e8-9d65-fe85d96cfdc7.png)

PR #941, opened 2018-10-31
fix(Dashboard): fix space between footer and empty message

jyenes:
####What does this PR do?
fix space between footer and empty message
closes: #908

### DIFF
--- a/app/components/pages/Dashboard/index.js
+++ b/app/components/pages/Dashboard/index.js
@@ -29,6 +29,7 @@ const MobileOnlyStickyFooter = styled(StickyFooter)`
 `
 
 const CenterdSmallParagraph = styled(SmallParagraph)`
+  margin-top: ${th('space.7')};
   color: ${th('colorTextSecondary')};
   text-align: center;
 `
--- a/app/components/pages/Dashboard/index.js
+++ b/app/components/pages/Dashboard/index.js
@@ -20,6 +20,7 @@ const BoxHiddenFromMobile = styled(Box)`
 `
 
 const TabbedSubmissions = styled(Tabs)`
+  min-height: 300px;
   margin-bottom: ${th('space.2')};
 `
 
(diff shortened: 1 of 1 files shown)


app/components/pages/Dashboard/index.js
```diff
@@ -29,6 +29,7 @@ const MobileOnlyStickyFooter = styled(StickyFooter)`
 `
 
 const CenterdSmallParagraph = styled(SmallParagraph)`
+  margin-top: ${th('space.7')};
```
jsms90: This spacing (space.7) is much larger than it used to be (space.2). It has this affect on the submissions tab:
image

Was this intentional?

Also, please see the last PR comment about `margin-top`. This issue is probably best resolved by applying a `min-height` to the `TabbedSubmissions` component & leaving the existing `margin-bottom` (unless it needs to be increased to `space.7`?)

jyenes: @jsms90 sorry, my bad, I will fix it now

jsms90: no problem at all :+1:

chugginselifesciences:
I feel like it needs more space when there are enough items in the list so it's top edge isn't so close. But looks ok when empty/in archive. No problem if it needs to go a bit lower to account for extra space when the list is full enough that the min-width isn't a factor.

jsms90:
@chugginselifesciences how much space would you like?

chugginselifesciences:
@jsms90 it looks like it's around 8px from the bottom of the list at the moment. Without mocking it up I would guess that the 'standard' vertical space of 24px would be enough. Thanks.

app/components/pages/Dashboard/index.js
```diff
@@ -20,6 +20,7 @@ const BoxHiddenFromMobile = styled(Box)`
 `
 
 const TabbedSubmissions = styled(Tabs)`
+  min-height: 300px;
   margin-bottom: ${th('space.2')};
```
jsms90: Which means that this just needs to be `space.3` @jyenes

jyenes: done

diversemix:
New screenshots .... @chugginselifesciences 

image
-----
and....
image

PR #942, opened 2018-10-31
style(StaticPages): fix responsive grid as applied to static page layout

### LINKED ISSUE #920: Static pages menu items wrapping
The panel itself is too narrow, so items are wrapping onto a second line between 1000px and 1137px viewport. This should be avoided:

screen shot 2018-10-30 at 10 57 25

We could either set a min-width or change the padding to a margin. Currently the panel doesn't seem as wide as the two grid columns it takes up in Sketch, so the Sketch design is not being replicated:

screen shot 2018-10-30 at 11 00 01

Note also that the font size should be 14px; currently it is 16px.

jsms90:
#### Background
- Design uses a 12 column grid on sketch
- Developers use fractions/percentages of viewport width (see [`grid-styled` docs](https://github.com/rebassjs/grid#responsive-styles))

To recap on our use of grid-styled...

We have an array of 4 breakpoints [in our theme](https://github.com/elifesciences/elife-xpub/blob/develop/client/elife-theme/src/index.js#L61): `[480, 768, 1000, 1272]`.

So whenever we add `width={[1stNum, 2ndNum, 3rdNum, 4thNum, 5thNum]}` to one of our `Flex` or `Box` components, it means that those numbers are applied to the component's width as follows:
- `width=1stNum` until viewport width reaches 480px
- `2ndNum` from 480px (inclusive) until 768px
- `3rdNum` from 768px (inclusive) until 1000px
- `4thNum` from 1000px (inclusive) until 1272px
- `5thNum` from 1272px (inclusive)

Similarly, we can use arrays for `p` & `m` on grid-styled components, so that these adhere to the breakpoints.

In the case of our `StaticPage` component, we are using `ml` & `mr` to achieve column-like spacing to the left and right of our content. So we use `${myNumber}vw` because vw = "viewport width"

#### What does this PR do?
- Change `padding-right` on sidebar to `margin-right`, so prevent wrapping of text
- Add missing numbers to the following arrays, so that together they add up to 12 / 12 for each breakpoint:
  - `margin-left` on sidebar (SideNavContainer)
  - `width` on sidebar
  - `width` on MainContainer
  - `margin-right` on MainContainer

This was done using https://invis.io/FDOIB61EKV5 as a guide

#### Any relevant tickets
fixes #920 

Visually

### DIFF
--- a/app/components/pages/StaticPages/SideNav.js
+++ b/app/components/pages/StaticPages/SideNav.js
@@ -12,6 +12,8 @@ const SideNavLink = styled(NavLink)`
   &:first-child {
     padding-top: 0;
   }
+  font-size: ${th('fontSizeBaseSmall')};
+  line-height: ${th('lineHeightBaseSmall')};
 `
 
 const SideNav = ({ navList }) => (
--- a/app/components/pages/StaticPages/SideNav.js
+++ b/app/components/pages/StaticPages/SideNav.js
@@ -1,0 +1,42 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+import { Box } from 'grid-styled'
+
+import NavLink from '../../ui/atoms/NavLink'
+
+const SideNavLink = styled(NavLink)`
+  display: block;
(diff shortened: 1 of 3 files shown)


jsms90:
@hdrury1 as in `fontSizeBaseSmall`? So the line height is 18px?

Issue #919, opened 2018-10-30
Titles of inactive submissions should be grey


hdrury1:
Dashboard list items with the status 'Submitted' should be in an inactive state, with the title colorSecondary. I'm not sure if this specific state exists in the codebase as yet, and whether it's worth creating it if not. Either way the title color should change.

xPub:

![screen shot 2018-10-30 at 10 43 49](https://user-images.githubusercontent.com/17023702/47712923-e5fe7680-dc30-11e8-8c51-b14f7eed9d23.png)

Design:

<img width="870" alt="screen shot 2018-10-30 at 10 45 34" src="https://user-images.githubusercontent.com/17023702/47712944-f3b3fc00-dc30-11e8-94a9-4c48d982cadd.png">

PR #949, opened 2018-11-01
style(DashboarListItem): make title grey for submitted documents

jyenes:
#### Background
make title grey for submitted documents

What does this PR do?
closes: #919

### DIFF
--- a/app/components/ui/molecules/DashboardListItem.js
+++ b/app/components/ui/molecules/DashboardListItem.js
@@ -29,6 +29,13 @@ const dashboardDateText = date => {
   return `${diffDays} ${diffDays === 1 ? 'day' : 'days'} ago`
 }
 
+const mapColor = statusCode =>
+  ({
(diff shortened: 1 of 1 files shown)


app/components/ui/molecules/DashboardListItem.js
```diff
@@ -29,6 +29,13 @@ const dashboardDateText = date => {
   return `${diffDays} ${diffDays === 1 ? 'day' : 'days'} ago`
 }
 
+const mapColor = statusCode =>
```
jsms90: Perhaps this is a good example of why `ManuscriptStatus` should be part of `DashboardListItem`? We now have the colour mapping function being repeated inside 2 files.

I'm all for breaking components into small pieces, but only when the smallest component makes sense in isolation. In this situation, we don't use `ManuscriptStatus` anywhere else & it actually adds complexity having these in separate files, rather than reducing complexity.

@jyenes @Cooryd Any thoughts?

Issue #687, opened 2018-09-28
Heading font-family errors (probably noto sans semibold)


chugginselifesciences:
**Describe the bug**
Some headings appear to be displayed in Arial instead of Noto Sans. In chrome there are "Failed to decode downloaded font" errors. 

**To Reproduce**
1. go to https://demo--xpub.elifesciences.org/submit/editors
2. Look at h3 "Suggest Senior Editors" 
3. in chrome developer tools check errors

**Expected behavior**
Noto Sans semibold should be displayed. 

**Screenshots**
Below it looks like the 2 screenshots in the middle look more like the Arial Bold version above (probably with added font-weight), than the Noto Sans Semibold version below.
![xpub-font-issue](https://user-images.githubusercontent.com/41907215/46220355-c3332680-c341-11e8-80ee-b60cb022fbab.png)

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser Chrome & Firefox
 - Version: 69.0.3497, 61.0.1

**Additional context**
Google fonts does not have Noto Sans Semibold listed as an available 'style'. Maybe this has something to do with it?

chugginselifesciences:
It looks like we're using Noto Sans and font-weight: 700 on the journal site instead of noto sans semibold. Is that something we can try first?

chugginselifesciences:
I'm getting these errors in XPub (in chrome)

![image](https://user-images.githubusercontent.com/41907215/46279492-218a2000-c561-11e8-9181-f7e162c98be4.png)

But these in the journal

![image](https://user-images.githubusercontent.com/41907215/46279519-31a1ff80-c561-11e8-88b8-443932e1ee3d.png)

chugginselifesciences:
Maybe this sheds some light on the problem? https://stackoverflow.com/questions/22001035/font-weight-ignored-in-chrome 

This seems to be specific to XPub, and seems to only really be a problem in Chrome. Note how at font-weight: 700 it looks significantly "lighter" than in Firefox and Safari

![image](https://user-images.githubusercontent.com/41907215/46615291-c1bae880-cb0f-11e8-8d5a-4e2490c0801a.png)

diversemix:
Nice summary :+1:  ... the SO article suggests adding `* {-webkit-font-smoothing: antialiased;}` - have we tried this @dryhten ?

chugginselifesciences:
[...]

Issue #913, opened 2018-10-30
Styleguide not visually representative


diversemix:
Issues with the fonts particularly. The elife logo doesn't load either. Have a chat with @chugginselifesciences if you are picking this up.

diversemix:
This is a simple fix ... in styleguide file assets needs to changed from a path `/assets/...` to  `/static/...` This loads the images and the fonts correctly - as verified by @chugginselifesciences 

@hdrury1 @jsms90

tamlyn:
As far as I know, there is no config for this in Styleguidist. The problem is that the folder is called `static` but it is copied into the webpack build as `assets`.

What we can do is update the weback config so that it doesn't rename the folder and either update all the links to start with `static` or rename the folder in the repo to `assets`.

diversemix:
Sounds like renaming the folder might also see off any other bugs that are hiding by not using the default. But happy either way.

PR #959, opened 2018-11-02
Fix semibold issue

### LINKED ISSUE #915: Fonts often too heavy or too light
In various places fonts are either too heavy or too light, depending on browser/device etc.

Usually buttons are too light (at least in chrome on Mac - Mojave renders even lighter but not the only cause)
image

And headings can often be too heavy, especially in Safari and Firefox
image

Have we decided/confirmed how we are intending to achieve the required Noto Sans Semibold? 

https://github.com/elifesciences/elife-xpub/issues/687 doesn't appear to have solved the issue.

It looks like we're using combinations of noto sans, noto sans semibold, and font-weight of 700. In places we have semi bold at 700 which gives extra heavy and blurry headings. And then we aren't using semibold in places we should be. And in some places we use "noto sans" (not semi bold) but use font-weight 700 to "fake" the semi bold weight, but it's blurry.

Where possible we should be using "Noto Sans Semibold" at 400 if this renders correctly. 

Apparently on the journal site it looks like we're achieving semibold by using noto sans and font-weight 700, but apparently where this exists it's actually being switched out with noto sans semibold at 400. I'm not sure how, but gives the results we want. Should we be doing the same in XPub?

I'm happy to investigate when and where it's not displayed correctly, but think we need to decide "how" we're going to apply it properly and then just use it everywhere we want to use semibold.

jsms90:
#### Background
Noto Sans SemiBold + non-Chrome browsers do not play well :cry: 

#### What does this PR do?
- override default font family of `Button` component in pubsweet, to now use `Noto Sans SemiBold`
- over-override font family `Action` component (that sometimes uses Button under the hood), so this is back to just `Noto Sans`
- add `font-weight: 400` everywhere that we use our `fontHeading` theme variable (aka `Noto Sans SemiBold` font family)

#### Any relevant tickets
fixes #915 
relates #687 

N/A - style change

### DIFF
--- a/client/elife-theme/src/elements/ui/Action.js
+++ b/client/elife-theme/src/elements/ui/Action.js
@@ -1,0 +1,9 @@
+import { css } from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+export default css`
+  // Action is sometimes a Link underneath & sometimes a Button
+  // eLife styling makes a button's font: Noto Sans SemiBold
+  // This reverses that override
+  font-family: ${th('fontInterface')};
+`
--- a/client/elife-theme/src/elements/ui/Action.js
+++ b/client/elife-theme/src/elements/ui/Action.js
@@ -1,9 +1,0 @@
-import { css } from 'styled-components'
-import { th } from '@pubsweet/ui-toolkit'
-
-export default css`
-  // Action is sometimes a Link underneath & sometimes a Button
-  // eLife styling makes a button's font: Noto Sans SemiBold
-  // This reverses that override
-  font-family: ${th('fontInterface')};
-`
--- a/app/components/global/AppBar/ProfileMenu.js
+++ b/app/components/global/AppBar/ProfileMenu.js
@@ -59,7 +59,8 @@ const MenuItem = styled.li`
 `
 
 const MenuHeading = styled.div`
-  font-weight: bold;
+  font-family: ${th('fontHeading')};
+  font-weight: 400;
   border-bottom: ${th('borderWidth')} ${th('borderStyle')} ${th('colorBorder')};
   padding: ${th('space.2')};
   padding-right: ${th('space.3')};
--- a/app/components/ui/molecules/DashboardListItem.js
+++ b/app/components/ui/molecules/DashboardListItem.js
@@ -50,7 +50,8 @@ const Root = styled(Flex)`
 
 const TitleBox = styled(Box)`
   color: ${props => props.theme[props.color]};
-  font-weight: bold;
+  font-family: ${th('fontHeading')};
+  font-weight: 400;
   text-align: left;
   flex-grow: 1;
   ${media.tabletPortraitUp`
--- a/app/components/ui/molecules/StaffCard.js
+++ b/app/components/ui/molecules/StaffCard.js
@@ -11,7 +11,8 @@ const StaffImg = styled.img`
 `
 
 const StaffName = styled.div`
-  font-weight: bold;
+  font-family: ${th('fontHeading')};
+  font-weight: 400;
 `
 const StaffCard = ({ name, telephone, jobTitle, photoURL }) => (
   <React.Fragment>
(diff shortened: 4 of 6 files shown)


hdrury1:
For @chugginselifesciences reference (I'm using Firefox):

screen shot 2018-11-02 at 12 38 05

screen shot 2018-11-02 at 12 37 50

Annoyingly, headers now look good but font weight on buttons looks too heavy.

hdrury1:
Chrome:

screen shot 2018-11-02 at 12 40 07

Safari:

screen shot 2018-11-02 at 12 40 42

client/elife-theme/src/elements/ui/Action.js
```diff
@@ -0,0 +1,9 @@
+import { css } from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
+
+export default css`
+  // Action is sometimes a Link underneath & sometimes a Button
+  // eLife styling makes a button's font: Noto Sans SemiBold
+  // This reverses that override
+  font-family: ${th('fontInterface')};
```
jsms90: Hmm, also this overriding the override isn't working. I guess because the Button one is more specific.

So now we have `Action` text (Pre-fill my details) which is more bold than it should be (because it's actually a `Button` with different styling):
image

:thinking:

jsms90:
I guess this needs @chugginselifesciences' input. I've put the label back on the issue, for the purposes of the project board.

client/elife-theme/src/elements/ui/Action.js
jsms90: Fixed in the latest commit

chugginselifesciences:
> I guess this needs @chugginselifesciences' input. I've put the label back on the issue, for the purposes of the project board.

Hi @jsms90. I've followed Hem's instructions to take a look on the style guide, but it's not loading fonts correctly (it's showing Arial). If @hdrury1 is ok with it can I suggest you get to a point where you think it looks close to what we're aiming for, and then merge and put it in the design review column and then I can take a look on staging?

The screenshots look ok to me.

@diversemix does this make sense to you given your investigations into the problem I'm having?

jsms90:
@chugginselifesciences Yeah that's fair enough. I was forgetting that you wouldn't be able to see these components correctly yet.

So essentially, this is what we have atm:

1. Chrome
image
image


2. Firefox
image
image

Same code. Different browsers :cry:

diversemix:
@jsms90 ... what we are finding is that what the browser says is displayed in the `Element` tab is certainly not what is displayed. @Cooryd and I did some playing around with code and within the browser and we cannot force the styleguide to be show what is shown within the app. 

For visuals appearances like fonts the styleguide is currently useless. FYI: @hdrury1

jsms90:
Yeah I completely understand @diversemix. But what we're discussing now is a separate issue - #913. One that is affecting Chris' ability to review this PR before we merge.

I completely agree that #913 needs sorting out, if we're going to enable Chris to review components using styleguidist. We definitely need to do that, otherwise @chugginselifesciences needs to wait until we've finished developing an entire epic at times, before being able to offer a design review, which obviously slows down our cycles.

But in the meantime, while styleguidist is not representative, I'm hoping that the copy/pasted pictures above while do for now?

client/elife-theme/src/elements/ui/Action.js
chugginselifesciences: I think we need option 3. I might now know enough about the implications for pubsweet, but my understanding is if we need something, we do it, and if it's helpful to pubsweet we add it.

I don't think pubsweet should prevent us doing what we need to do in this instance, and it seems like it might helpful to others to have the option to have different fonts/weights on "buttons" and "links". So unless there's a problem or reason I'm not aware of I think we should we should go with:

> adding a prop to pubsweet's Action component, to provide us with some way of differentiating between Actions which are actually Links & Actions which are actually Buttons

One thing that might be worth noting is that we have `Link`s that use both Noto Sans and Noto Serif fonts - not sure if its related but surely using different `font-family`s

Issue #926, opened 2018-10-30
Heading on login screen overlaps appbar on mobile


hdrury1:
1. Go to https://staging--xpub.elifesciences.org/login while on mobile
2. See header overlapping the app bar

![screen shot 2018-10-30 at 14 16 28](https://user-images.githubusercontent.com/17023702/47724527-c5451980-dc4e-11e8-9774-77e4f603196b.png)

jsms90:
FYI this is happening because of the somewhat hacky way we tried to get this into the center of the screen (vertically)

https://github.com/elifesciences/elife-xpub/blob/3411b2ae625ba0faf87bb0d139f42902b15dbfca/app/components/pages/Login/index.js#L78-L79

PR #960, opened 2018-11-02
fix(Login): fix issue overlapping header

jyenes:
#### Background
there is an issue in login page for mobile, that overlaps the header

What does this PR do?
closes: #926

### DIFF
--- a/app/components/pages/Login/index.js
+++ b/app/components/pages/Login/index.js
@@ -5,10 +5,13 @@ import { withApollo } from 'react-apollo'
 import gql from 'graphql-tag'
 import config from 'config'
 import { H2 } from '@pubsweet/ui'
+import styled from 'styled-components'
+import { th } from '@pubsweet/ui-toolkit'
 import OrcidButton from '../../ui/atoms/OrcidButton'
 import ButtonLink from '../../ui/atoms/ButtonLink'
 import Paragraph from '../../ui/atoms/Paragraph'
 import NativeLink from '../../ui/atoms/NativeLink'
+import media from '../../global/layout/media'
 
 const { url: loginUrl, signupUrl, legacySubmissionUrl } = config.login
 
@@ -16,6 +19,12 @@ const EXCHANGE_TOKEN_MUTATION = gql`
   mutation($token: String) {
     exchangeJournalToken(token: $token)
   }
+`
+const Container = styled(Box)`
+  padding-top: ${th('space.3')};
+  ${media.tabletPortraitUp`
+    padding-top:${'calc(50vh - 300px)'};
+  `};
 `
 
 class LoginPage extends React.Component {
@@ -75,7 +84,7 @@ class LoginPage extends React.Component {
     }
 
     return (
-      <Box mx="auto" width={[1, 410]}>
+      <Container mx="auto" width={[1, 410]}>
         {/* roughly centre the box vertically */}
         <Box mb={4}>
           <H2>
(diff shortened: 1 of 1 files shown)


diversemix:
Bummer! ...
> Calc of max, or max of calc in CSS
> No you cannot. max() and min() have been dropped from CSS3 Values and Units. They may be re- introduced in CSS4 Values and Units however. There is currently no spec for them, and the calc() spec does not mention that either are valid inside a calc() function.
> 
> --David Storey

Cooryd:
Maybe have a `@media(max-height: 650px)` and set the `margin-top: ${th('space.3')}` to prevent the overlap on small screen sizes?

app/components/pages/Login/index.js
```diff
@@ -75,7 +75,7 @@ class LoginPage extends React.Component {
     }
 
     return (
-      <Box mt="calc(50vh - 300px)" mx="auto" width={[1, 410]}>
+      <Box mt="calc(100px + 50vh - 300px)" mx="auto" width={[1, 410]}>
```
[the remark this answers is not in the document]
jyenes: I tried on the desktop but looked fine, in my 13 inches screen, I have decreased the screen zoom and I managed to reproduce the error, I was doing some test and removing just this margin-top is being displayed correctly either in desktop or mobile

screenshot 2018-11-02 at 14 43 39
screenshot 2018-11-02 at 14 44 05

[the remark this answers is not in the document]
jsms90: @jyenes The reason this margin-top was there in the first place is that the spec for the desktop version asks for the content to be _vertically_ aligned in the centre of the screen, as well as horizontally central.

jsms90:
@diversemix @Cooryd did you wanna have another look?

app/components/pages/Login/index.js
```diff
@@ -17,6 +20,12 @@ const EXCHANGE_TOKEN_MUTATION = gql`
     exchangeJournalToken(token: $token)
   }
 `
+const Container = styled(Box)`
+  padding-top: ${th('space.3')};
+  ${media.tabletPortraitUp`
+    padding-top:${'calc(50vh - 300px)'};
```
Cooryd: The AppBar takes up a total of 144px of space on the screen at desktop (71px tall, 1px border, 72px bottom-margin). This isn't being included in this calc which is causing the content to be pushed too far down.

jsms90: @Cooryd I don't think the bottom margin on the AppBar comes into play here because of margin collapsing? The `margin-top` on this `Box` is larger than the `margin-bottom` on the AppBar, so the margin-top on the Box will win out. Right?

Our margin-top comes from this calculation:
(`height of viewport` - `height of app bar` - `height of Box`) / `2`
(`100vh` - `72px` - `548px`) / 2
which can be simplified to: `50vh - 310px`

I think 300px rather than 310 may just have been a rough estimate, or rounding?

jsms90:
@diversemix @Cooryd, sorry. I wrote that comment on the wrong PR :sweat_smile: 

We finished the scrolling PR first & that's the one I meant to put the comment on. My bad :laughing:

app/components/pages/Login/index.js
jyenes: screenshot 2018-11-05 at 13 04 36
screenshot 2018-11-05 at 13 04 15

Cooryd:
Have approved as this works for 99% of use cases now. There is still the obscure scenario where the user is on desktop but reduces the height of the browser causing the text to overflow the `AppBar` still. Will raise as another ticket.

PR #961, opened 2018-11-02
Improve dashboarditem dates and add tests

diversemix:
#### Background

Improves the display of dates on the dashboard item.

#### Any relevant tickets

Closes 957

Locally

### DIFF
--- a/app/components/ui/molecules/DashboardListItem.js
+++ b/app/components/ui/molecules/DashboardListItem.js
@@ -1,32 +1,41 @@
 import React from 'react'
 import styled from 'styled-components'
 import { Flex, Box } from 'grid-styled'
-import {
-  differenceInCalendarDays,
-  differenceInCalendarMonths,
-  format,
-} from 'date-fns'
+import { differenceInCalendarDays, format } from 'date-fns'
 import { th } from '@pubsweet/ui-toolkit'
 import PropTypes from 'prop-types'
 import ManuscriptStatus from '../atoms/ManuscriptStatus'
 import media from '../../global/layout/media'
 
-const dashboardDateText = date => {
+export const dashboardDateText = date => {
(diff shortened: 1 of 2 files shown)


app/components/ui/molecules/DashboardListItem.js
```diff
@@ -1,32 +1,41 @@
...
 import ManuscriptStatus from '../atoms/ManuscriptStatus'
 import media from '../../global/layout/media'
 
-const dashboardDateText = date => {
+export const dashboardDateText = date => {
   const diffDays = differenceInCalendarDays(new Date(), date)
-  const diffMonths = differenceInCalendarMonths(new Date(), date)
-  if (diffDays < 0) {
+  if (diffDays < 0 || Number.isNaN(diffDays)) {
     return 'Invalid date'
```
jyenes: I would suggest rather than having multiple return statements refactor this using else clauses and extract the return value to a variable, we have ended up with 7 return paths.

diversemix: @jyenes I see this as a style thing as functionally I don't see any difference. I've worked both ways - but mainly having only one exit point from a function was only used in C family of languages.

Input? @Cooryd @jsms90 @tamlyn @damnhipster

diversemix: FTR: I was just maintaining the previous style of the function

jyenes: I know @diversemix it was doing by Mihail and myself, but I thought it can be refactored in a better way, it's just a suggestion as behaviour wouldn't change.

damnhipster: I'm happy either way, as long as we keep it consistent, and look at enforcing it using linting.

Issue #956, opened 2018-11-02
Static pages not scrolling to the top on load


hdrury1:
**Describe the bug**
When clicking on a link within the author or reviewer guides that take the user to a different section, the browser window does not scroll to the top on load.

**To Reproduce**
1. Go to https://staging--xpub.elifesciences.org/author-guide/editorial-process
2. Click on '(read more)' in first call out box to go to the publication fees section
3. See that the browser loads with the scroll bar at the same place in the page, rather than at the top of the window.

**Expected behavior**
When a new section is navigated to, the browser should scroll to the top of the window (unless link is directing towards a specific anchor point in the section).

**Desktop (please complete the following information):**
 - OS: macOS High Sierra
 - Browser: Firefox

Date: Thu Nov 1 17:53:05 UTC 2018 - Commit: 3411b2a

PR #954, opened 2018-11-01
Fix internal page anchor links

jsms90:
#### Background
At the moment we use 2 (or 3) different types of links in our application:
- `ExternalLink` - our own component: an anchor tag with eLife styling applied
- `Link` from pubsweet - which uses react-router's `Link` under the hood & applies styling)

(Note: elsewhere we also import `Link` directly from react-router, in situations where it's not styled normally i.e. non-blue links in ProfileMenu, DashboardList, etc.)

Until now, we had never linked to a heading within the same page (via an id).

#### What does this PR do?
- Fixes broken links in `EditorialProcess` & `Fees` components by changing `Link` (react-router) to `ExternalLink` (regular anchor tag)
- Renames the `ExternalLink` component to `NativeLink`, to better describe all use cases
- Updates ExternalLInk/Native Link styleguidist description accordingly

#### Any relevant tickets
Fixes #907 
Fixes #955

#### How has this been tested?
Manually clicking the links

PR #849, opened 2018-10-22
simple static content components

jsms90:
#### What does this PR do?
- Creates content components listed in issue #832 - i.e. those under the "Author Guide", "Reviewer Guide" & "Contact Us" headings, which do not require any more sub-components to be built
- Removes underlining from our internal links (`Link` from pubsweet), so they look the same as our external links
- Adds correct font to list items
- Add change margin to padding on all headings (H1, H2, etc) so that they align to the 24px baseline grid (see [typography page in Zeplin](https://app.zeplin.io/project/5bacf042615a8b6e20e84a5f/screen/5bc0cddca55ed15fd3fb3973))
- Add margin-bottom to `Paragraph` & `SmallParagraph` (except when they're the last child), so we have `24px` between elements

Please be aware:
- Internal application links will not lead anywhere until #505 is complete & I anticipate potential issues with the internal page link to the FAQ section - see this issue on `react-router-dom`: https://github.com/ReactTraining/react-router/issues/394

#### Any relevant tickets
fixes #832 
relates #840 

#### How has this been tested?
Visually

PR #964, opened 2018-11-02
Static Pages: not scrolling to the top when click on relative link

jyenes:
#### Background
not scrolling to the top when clicking on the relative link `read more`

What does this PR do?
closes: #956

### DIFF
--- a/app/components/pages/StaticPages/ScrollToTop.js
+++ b/app/components/pages/StaticPages/ScrollToTop.js
@@ -1,0 +1,16 @@
+import React from 'react'
+import { withRouter } from 'react-router-dom'
+
+class ScrollToTop extends React.Component {
+  componentDidUpdate(prevProps) {
+    if (this.props.location !== prevProps.location) {
+      window.scrollTo(0, 0)
+    }
+  }
+
+  render() {
+    return this.props.children
+  }
+}
+
+export default withRouter(ScrollToTop)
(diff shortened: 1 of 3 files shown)


jyenes:
don't merge yet, I realise there are other parts that are not scrolling

jyenes:
ready to merge

jsms90:
I've just realised that perhaps you were following the approach I took in PR #954?

Just to give a bit of explanation - the reason why I chose to change `Link` to `NativeLink` there, is because the anchor tag is directing to the same page that the user is on. So there's no delay in load time.

I imagine that the way this works is that the browser looks at the `href` on the anchor element, sees that it's linking to the same page being served & the browser itself knows not to re-render the entire page, but simply jump to the correct place. As such, that causes no load time issues.

Admittedly, it does mean that we jump the user to the right place on the page instead of scrolling (which really isn't the best user experience). But react-router doesn't seem to have an in-built, simple way of handling these same-page links - see my warning on https://github.com/elifesciences/elife-xpub/pull/849#issue-224641173 & this issue in the react-router repo: ReactTraining/react-router#394

So, given the decent load time with this approach, I figured that implementing this manually or using a library didn't seem worth it, by comparison to using `NativeLink`.

damnhipster:
There's a [guide on scrolling with React Router](https://reacttraining.com/react-router/web/guides/scroll-restoration). I agree with @jsms90, avoid using `NativeLink` unless you're scrolling to a specific section of a page using the `id` attribute of an element.

If we extract out the scrolling logic (given we follow the guide and create it), we could look at replacing `NativeLink` with `Link` again and using the scrolling logic instead.

app/components/pages/StaticPages/ScrollToTop.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { withRouter } from 'react-router-dom'
+
+class ScrollToTop extends React.Component {
```
Cooryd: A component like this could use `React.PureComponent` or even the new `React.memo` from `v16.6.0` for improved performance.

jsms90:
@damnhipster happy with this?

PR #969, opened 2018-11-03
WIP : First stab at an admin page - not to be merged

diversemix:
#### Background

Creates an Admin page

#### Any relevant tickets

Not really - just playing at the weekend

Locally - though I do need to write some tests and break out the administrator list into a separate config file... or consider adding a flag to the database.

### DIFF
--- a/app/components/pages/Dashboard/ManuscriptsQuery.js
+++ b/app/components/pages/Dashboard/ManuscriptsQuery.js
@@ -1,8 +1,8 @@
 import React from 'react'
 import { Query } from 'react-apollo'
 
-const ManuscriptsQuery = ({ query, children }) => (
-  <Query fetchPolicy="cache-and-network" query={query}>
+const ManuscriptsQuery = ({ query, children, variables }) => (
+  <Query fetchPolicy="cache-and-network" query={query} variables={variables}>
     {({ data, loading, error }) => {
       if (loading && !data.manuscripts) {
         return <div>Loading...</div>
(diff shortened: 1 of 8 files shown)


diversemix:
@tamlyn / @Cooryd ... this is an idea I had while Gaby was asleep after our Saturday walk!  - so I thought I'd have a go. I'd be interested in your thoughts as to my approach.

tamlyn:
I think an admin page will be useful/essential. But I would be tempted to either keep it separate from the main app code or fully integrate it with user roles and teams. As it stands it feel like a half way solution.

diversemix:
@tamlyn thanks for the input - this is something @Cooryd has suggested too, yeah I guess this is something we could "slip in quickly" if needed - but what you suggest is defo the longer term solution :+1:

diversemix:
@hdrury1 - I'd like to just close this down but raise a task to create some `rudimentry` Admin pages for release. I propose these are done in the style of xPub and are intended for developers and staff to see what is going on xPub-wize.

/cc @Cooryd

jsms90:
I guess the fact that we've got this PR, is really just highlighting how urgent it is for the staff to have their own dashboard on xpub?

From speaking to @hdrury1 and @chugginselifesciences briefly on Wednesday, it sounds like a staff dashboard is likely to be the next body of work, to be done straight after first release (since staff will need their own view of the dashboard, in order to then go in and complete the QC step).

I suppose the question is, how essential is it for staff to have a "rudimentary" dashboard before first release? Or can we afford to wait until just after first release, given that this is going to be the next piece of work anyway?

diversemix:
Well ... this PR also shows how simple it is to knock one up ... but as Tam pointed out we should probably start thinking about Teams and Roles anyhow. This is an area that we should perhaps book a meeting to throw around some ideas. What do you think @hdrury1  ?

damnhipster:
Awaiting requirements from product

PR #982, opened 2018-11-05
styleguide fix issue with static content

### LINKED ISSUE #913: Styleguide not visually representative

jyenes:
#### Background
static content not being displayed properly on Styleguide, so renamed  to assets

What does this PR do?
closes #913


jyenes:
fixed right now!

jyenes:
at what point this `https://s3.amazonaws.com/ci-elife-xpub-styleguide/913_assets_styleguide/index.html#peoplepicker` is being built?
it seems is not picking the new build changes.

jyenes:
@giorgiosironi directory structure is now correct, the issue we have right now is how the content is being served, is there anyway in S3 to resolve /assets to include the bucket name?

screenshot 2018-11-08 at 10 41 57

giorgiosironi:
From the call:

- can't resolve `/assets` to a branch-specific folder while using a single bucket for all branches
- trying to change the path within the Webpack build
- reserve option is to generate a container (derived from `nginx`) with strange routing rules, but it's much more complex to maintain

styleguide.config.js
```diff
@@ -1,4 +1,17 @@
 const path = require('path')
+const CopyWebpackPlugin = require('copy-webpack-plugin')
```
diversemix: Is there any reason this file lives in the root and not in `webpack` alongside the other config files?

diversemix: Just to be clear the file `styleguide.config.js` ... expected it to be something like `webpack/styleguide.config.js`

jsms90: Does your comment relate to this line of the file?

Or about the location of this whole file?

styleguide.config.js
```diff
@@ -44,13 +57,14 @@ module.exports = {
   ],
   webpackConfig: {
```
jsms90: @diversemix Is your comment suggesting that we move this piece of the styleguidist config out of this file, and put it under the `/webpack` directory?

styleguide.config.js
diversemix: The whole file

styleguide.config.js
diversemix: My question is, if it is webpack config - why doesn't it live in the webpack folder?

styleguide.config.js
jsms90: I think this whole file needs to stay in the root. But we are right now moving just the webpack portion of this whole styleguidist config file into the `/webpack` directory.

Does that sound better?

styleguide.config.js
jsms90: The webpack portion of this config file does now live in the webpack folder. But the rest of the styleguidist config is still in the root.

Let us know what you think after this latest commit :+1:

styleguide.config.js
```diff
@@ -1,4 +1,17 @@
 const path = require('path')
+const CopyWebpackPlugin = require('copy-webpack-plugin')
+const rules = require('./webpack/rules.js')
```
diversemix: Lines 3-14 could live in `webpack/styleguide.rules.js` which in turn could be required on line 47 - this ensures all the webpack config lives in the webpack folder. Let me know if I am missing something?

Issue #917, opened 2018-10-30
Use PubSweet BaseModel


tamlyn:
Refactor `@elifesciences/xpub-model` to use `@pubsweet/base-model` (and therefore Objection.js ORM).

diversemix:
Raised #989 as per our discussion yesterday

diversemix:
@damnhipster / @Cooryd - can we add a `md` file into `docs/developing` to record some of the idiosyncrasies of using `knex` and `objective` we discussed yesterday. This this may prove useful in sharing knowledge and for others especially as we come up to speed with this ORM? thoughts?

damnhipster:
#984 overrides the default `save` operation. We should look at removing this due to the danger introduced of deleting related entities.

damnhipster:
Raised #1006 to address the `save` operation

PR #984, opened 2018-11-06
feat: use pubsweet base model for xpub-model

tamlyn:
#### Background

Replace most of `@elifesciences/xpub-model` with an ORM based on `@pubsweet/base-model`. Outside of that package it's mostly updating usage from stuff like `ManuscriptManager.save(manuscript)` to `manuscript.save()`. 

#### Other changes

- BREAKING: remove `meta,title` columns in favour of a single `meta` JSONB column. There is a migration to change the schema but it doesn't attempt to migrate the data of existing manuscripts.
- Prevent Jest from feeding server code through Babel because it does strange things with classes. This has the side effect that jest mocks no longer get automatically hoisted to the top of the file so you have to ensure the mocks come before the requires.
- Don't have empty reviewers team in `emptyManuscript` otherwise it gets added to every manuscript on instantiation. Instead have a `setDefaults()` method that only adds it in `createManuscript`. 

#### Dragons 🐉

I've found `objection` hard to get to grips with. It doesn't really do anything by itself but it exposes helpers that let you do ORM type things. It has a very large API and the documentation is written in a linear fashion that makes it hard to dip in and out of. Loading of related entities seems to be either buggy or designed to behave in a way that I am yet to comprehend. Unfortunately this is  

The PubSweet `BaseModel` implementation also does some unusual (to me) things like:

- `manuscript.save()` doesn't return `this`, it returns a new instance of the entity which may have different data from the original, e.g. `id` and `created` will have been populated.
- `manuscript.save()` doesn't automatically save related entities like files and teams. They have to be saved individually.

In this PR I replace the default save logic of `BaseModel` with the `graphUpsert` feature of `objection` to "fix" these two issues. But it also comes with an unexpected behaviour of its own in that it actively deletes any files and teams from that are not currently in the `files` and `teams` arrays of the manuscript. That means that if the related entities failed to load and you save the manuscript it will delete all existing related entities with this `manuscriptId`.

Contributes to #917


server/xpub-model/entities/manuscript/index.js
```diff
@@ -16,67 +14,145 @@ const mergeObjects = (...inputs) =>
...
+        decision: { type: ['string', 'null'] },
+        previousVersion: { type: 'uuid' },
+      },
+    }
+  }
+
+  static get jsonAttributes() {
+    // tell objection to serialise arrays as a postgres array rather than JSON
+    return []
+  }
```
damnhipster: This needs looking into, not sure what the impact is.

server/xpub-model/entities/manuscript/index.js
```diff
@@ -16,67 +14,145 @@ const mergeObjects = (...inputs) =>
...
+    const [manuscript] = await this.query().where({
+      'manuscript.id': id,
+      'manuscript.created_by': user,
+    })
+
+    if (!manuscript) {
+      throw new Error(`${this.name} not found`)
     }
+    // todo why does eager loading sometimes not work?
+    await manuscript.$loadRelated('[teams, files]')
```
damnhipster: What is meant by eager loading? Do we expect relations to be resolved without explicitly doing it ourselves?

server/xpub-model/entities/manuscript/index.js
```diff
@@ -16,67 +14,145 @@ const mergeObjects = (...inputs) =>
...
+    // reload related entities
+    await this.$loadRelated('[teams, files]')
+    return this
+  }
 
-    return { ...manuscript, id }
-  },
+  // atomically update the manuscript status
+  static async updateStatus(id, status) {
+    // Can't use objection's patch method as this overwrites existing data with defaults
```
damnhipster: Fair enough.

server/xpub-model/entities/manuscript/index.js
damnhipster: I've managed to reproduce the issue locally.

server/xpub-model/migrations/1540916414-meta-json.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE manuscript
+  ADD COLUMN meta JSONB,
+  DROP COLUMN "meta,title",
+  DROP COLUMN "meta,article_type",
+  DROP COLUMN "meta,article_ids",
+  DROP COLUMN "meta,abstract",
+  DROP COLUMN "meta,subjects",
+  DROP COLUMN "meta,publication_dates",
+  DROP COLUMN "meta,notes";
```
damnhipster: This should also migrate existing rows.

damnhipster:
> manuscript.save() doesn't return this, it returns a new instance of the entity which may have different data from the original, e.g. id and created will have been populated.

This has been captured in a test.

server/xpub-model/entities/manuscript/index.js
```diff
@@ -16,67 +14,145 @@ const mergeObjects = (...inputs) =>
...
+    await Promise.all(
+      manuscripts.map(manuscript => manuscript.$loadRelated('[teams, files]')),
+    )
+    return manuscripts
+  }
+
+  async save() {
+    // save manuscript and all related files and teams
+    // note that this also deletes any related entities that are not present
+    await this.$query().upsertGraphAndFetch(this)
```
damnhipster: We should look at removing this due to the danger of deleting related entities.

server/xpub-model/migrations/1540916414-meta-json.sql
damnhipster: We're going to wipe the databases in existing environments instead.

server/xpub-model/entities/manuscript/index.js
damnhipster: Added to #917

damnhipster:
The manuscript upload issue seems to occur on all review environments.

PR #1016, opened 2018-11-09
Fix manuscripts on dashboard

### LINKED ISSUE #1014: Dashboard shows submissions of other users
**Steps to Reproduce:**

- Go to https://staging--xpub.elifesciences.org/
- Login with the an Orcid sandbox user
- All submissions for all users, including unfinished submissions, are displayed

image 4

- Users are unable to continue submissions initially created by other users.

screen shot 2018-11-09 at 15 52 55

damnhipster:
Closes #1014 

I'm unable to test this outside of Staging as our login mock only supports a single user.

### DIFF
--- a/server/xpub-model/entities/manuscript/index.js
+++ b/server/xpub-model/entities/manuscript/index.js
@@ -129,6 +129,17 @@ class Manuscript extends BaseModel {
     return manuscripts
   }
 
+  static async all(user) {
+    const manuscripts = await this.query().where({
+      'manuscript.created_by': user,
+    })
+
+    await Promise.all(
+      manuscripts.map(manuscript => manuscript.$loadRelated('[teams, files]')),
+    )
+    return manuscripts
+  }
+
   async save() {
     // save manuscript and all related files and teams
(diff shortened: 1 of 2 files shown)


damnhipster:
There currently isn't any functionality where we need a `Manuscript.all` function, or a `File.all` or `User.all`. It's safer to not expose the methods at all until they're needed.

diversemix:
OK, so let's hide `all` by overriding to throw an exception - and do the rename on the existing function as currently its name is misleading.

damnhipster:
I think this requires a different refactor where we query the user for their manuscripts, rather than querying manuscripts by user. I thought it may be worth doing it in a separate pull request.

The existing functions within the `Manuscript` model expect a user to be passed in, such as `.find` and `.findByStatus`. I suggest we keep it consistent until we've managed to perform the above refactor.

diversemix:
Ok ... I'm with you on the application logic of `user -> manuscript` and lets get this in so that its fixed short-term on the understanding that it needs to be changed. :+1:

PR #1017, opened 2018-11-09
[wip] Update MECA package to contain surname and email

### LINKED ISSUE #991: Change the MECA package to use the API's new fields.
see #545

diversemix:
#### Background

Creates a MECA package with `surname` and `email` to allow eJP to link up the users.

#### Any relevant tickets

Closes #991 

Locally - currently fails in CI as this uses the data from the endpoint `https://continuumtest--gateway.elifesciences.org/` which is not yet populates


server/meca/file-generators/article.js
```diff
@@ -1,7 +1,8 @@
 const xmlbuilder = require('xmlbuilder')
 const lodash = require('lodash')
 const config = require('config')
-const { User } = require('@elifesciences/xpub-model')
+const elifeApi = require('@elifesciences/xpub-model/entities/user/helpers/elife-api')
```
damnhipster: I suggest we expose elife-api at the top level of the package.

```suggestion
const { elifeApi } = require('@elifesciences/xpub-model')
```

server/xpub-model/entities/user/helpers/elife-api.js
```diff
@@ -56,4 +83,5 @@ module.exports = {
   person,
   peopleById,
   profile,
+  getEditorsByPersonId,
```
damnhipster: This function should be removed from the `User` model if we're expecting to call it from the API instead

damnhipster: In fact, it may be better off exposed as a static method under the `Team` model.

server/meca/file-generators/article.js
```diff
@@ -110,7 +113,7 @@ async function generateArticleXml(manuscript) {
     return ids
   }, [])
 
-  const editors = await User.getEditorsByPersonId(editorIds)
+  const editors = await elifeApi.getEditorsByPersonId(editorIds)
```
damnhipster:
```suggestion
  const editors = await Team.getEditorsByPersonId(editorIds)
```

server/meca/file-generators/article.js
damnhipster: Even better would be to expose the function as a static method on the `Team` model.

```suggestion
const { Team } = require('@elifesciences/xpub-model')
```

server/meca/test/http-mocks/api.elifesciences.org-443/154169448713464007
```diff
@@ -0,0 +1,28 @@
+GET /people/6ce3402a
+accept-encoding: gzip, deflate
+authorization: xpubsecret
```
damnhipster: This should be removed.

server/xpub-model/entities/user/helpers/elife-api.test.js
```diff
@@ -0,0 +1,43 @@
...
+const request = require('superagent')
+
+describe('eLife API tests', () => {
+  it('sends the Authorization token', async () => {
+    await api.people()
+    const header = request.getHeader()
+
+    expect(Object.keys(header)).toContain('Authorization')
+    expect(header.Authorization).toBe('xpubsecret')
+  })
```
damnhipster: This looks good

damnhipster:
Can we rename obsforcate to obfuscate?

damnhipster:
Resolved by #1065

PR #1028, opened 2018-11-13
639 stateful upload

### LINKED ISSUE #639: refactor FileUpload component to be stateful
We currently have some bugs:
> There's a little bit of weirdness in using `setFieldError` in that the message disappears on the next form validation (e.g. when blurring the cover letter editor). Another issue is that it favours `conversion.error` (i.e. server errors) over form errors. That means if you upload a valid file and it fails on the server, then you try to upload an invalid file and it fails on the client, the server error continues to be displayed rather than the new form error.
>
> I think these are acceptable edge case bugs for now but happy to be convinced otherwise. The alternative is to create a stateful component which tracks errors from various sources and can choose to show the most recent error, for example.

After a quick demo, it was agreed that the error being shown to the user should be the only that has occurred most recently.

Another bug: there are two aspects of the error state UI which are currently working independently of each other - the red border & the contents of the box (icon and message). Right now these are handled by different components.

These are best dealt with by making the component stateful

## First time upload 

- User should see default state on first sight
- onBlur without upload remains default state
- Click next without uploading displays 'Please upload a manuscript' error state
- If user tries to upload one file and files to big then the error state is set to 'That file is not supported' state
- If the user tries to upload one file and it isn't a supported type then the That file is not supported state is set
- If they try to upload a file that is to big then they try to upload an unsupported file then the UI shouldn't change
- If the user tries to upload multiple files at once then the component displays the error state 'Only one manuscript file can be uploaded'
-  If user uploads correct file then user views loading component state
- If an error occurs server side when uploading file the component should change to an error state 'There was a problem uploading your file'
- once upload and conversion has finished the component should display a success state
- During upload if the user navigates back a step in the wizard then returns to upload step, latest progress percentage displays or the success / error state should display
- **What should happen when you press next in the middle of an upload?**
- **What should happen if you navigate to the dashboard in the middle of an upload? Should upload continue or be cancelled?**

## Replacement upload

- **If user tries to replace file with an invalid file, the correct error state for the replacement file should be shown and the previous file should be removed from the submission. Navigating away from the wizard and returning will show the 'Please upload a manuscript'**

### LINKED ISSUE #646: Fix border color when reuploading manuscript
Right now if the manuscript upload fails the border of the dropzone will turn red. If you try to upload a manuscript again, while the text "Manuscript is uploading" is displayed the border will remain red. This doesn't look like something we would like and the color should change back to gray/text-placeholder while the manuscript is uploading.

manuscript-uploading

### LINKED ISSUE #653: Upload Progress does not resume after navigating away
If you begin to upload a file (the bigger the better) then navigate back.... then forward again ... you have the option to replace (as the component is unaware of the upload progress still in play - this can be seen from the ws-frames). 

You can begin to upload a second one and then the upload progress flicks between the two upload percentages ... rather fun ... but not desired :)

Cooryd:
#### Background
The `ManuscriptUpload` component currently has some issues with displaying the incorrect error message when triggering an auto save by changing another field (cover letter for example). The component needs to always show the most recent error message that matches what has actually gone wrong.

When a user uploads a file, going forward or backwards within the wizard closes the progress connection so that returning to the upload step doesn't correctly show that a file is still being uploaded.

The border colour of the `Dropzone` component should be aligned with the current status of the File field. When in an error state it should be red, otherwise it should not.

What does this PR do?

Refactors the `ManuscriptUpload` component so that is now stateful.
Moves the progress subscription up to the `WizardSteps` level so that the upload progress may be tracked when navigating to other wizard steps then back to the upload step.

#### Any relevant tickets

close #639 
close #653
close #646

Unit tests
Visually


jsms90:
@Cooryd I've pulled this down locally and the New Submission button isn't working. Looks like it just needs a rebase, so gonna force push if that's ok with you?

app/components/pages/SubmissionWizard/steps/Files/utils.js
```diff
@@ -0,0 +1,6 @@
+export const errorMessageMapping = {
+  EMPTY: 'Please upload a manuscript.',
+  MULTIPLE: 'Only one file can be uploaded.',
+  UNSUPPORTED: 'That file is not supported.',
+  UNSUCCESSFUL: 'There was a problem uploading your file.',
```
jsms90: "UNSUCCESSFUL" kind of describes all 4 of these situations. Can anyone think of a name which describes this specific scenario, which differentiates it from the others?

Maybe "UPLOAD_UNSUCCESSFUL"? Or "UPLOAD_FAILURE"? Just so it's clear that it's the upload itself that's gone wrong, rather than the user being unsuccessful because the file type is unsupported?

app/components/pages/SubmissionWizard/steps/Files/ManuscriptUpload.js
```diff
@@ -135,34 +133,67 @@ const DropzoneContent = ({ conversion, formError, dropzoneOpen }) => {
...
-        </Box>
-      </CentredFlex>
-    </StyledDropzone>
-  )
+class ManuscriptUpload extends React.Component {
+  state = {
+    errorMessage: null,
+  }
+
+  componentDidUpdate(prevProps) {
```
jsms90: Would be good to have tests that cover all the scenarios in here, since this is something new that's being added with this PR

app/components/pages/SubmissionWizard/index.js
```diff
@@ -13,104 +14,111 @@ import { schema as editorsPageSchema } from './steps/Editors/schema'
...
-              handleAutoSave={updateManuscript}
-              handleButtonClick={updateManuscript}
-              history={history}
-              initialValues={initialValues}
-              nextUrl={`${match.url}/submission`}
-              previousUrl={`${match.url}/author`}
-              step={1}
-              title="Write your cover letter and upload your manuscript"
-              validationSchema={filesPageSchema}
+      <Subscription subscription={ON_UPLOAD_PROGRESS}>
```
jsms90: The change in the PR - moving the subscription up into the Submission Wizard level instead of the File Upload component itself - means that this product requirement is now satisfied:
> During upload if the user navigates back a step in the wizard then returns to upload step, latest progress percentage displays or the success / error state should display

So it would be good for this PR to also include a test that covers this scenario?

app/components/pages/SubmissionWizard/steps/Files/utils.js
Cooryd: `UPLOAD_FAILURE` seems the clearest option to me.

diversemix: I'd be happy with `FAILURE` ... minor point but I think we should group the upload errors ... either with a prefix `UPLOAD_...` or by renaming the object to `uploadErrorMessages` or similar

app/components/pages/SubmissionWizard/index.js
Cooryd: I've left testing for this out of this PR as I think it would fit into an issue for adding more Submission Wizard tests as well as extending the browser test with regards to uploading.

jsms90: Sure thing. I don't blame you for wanting to get the refactor in, so that we can unblock the other issues, if nothing else.

But do you not think it's good to write tests as we go?

PR #1029, opened 2018-11-13
Use internal urls within AWS environments

### LINKED ISSUE #948: In all environments use the internal URLs for server-side calls
When we are in production we can access the endpoints of journal internally by using `elife.internal` - for example `http://prod--gateway.elife.internal`

Ref: https://github.com/elifesciences/builder-configuration/blob/master/pillar/environment-prod-public.sls#L12

I'm anticipating that this is only a change to the config for the API url... but its worth checking. Contact @diversemix or @giorgiosironi  if you have any questions.

diversemix:
#### Background

Use internal urls when running in AWS environments to prevent traffic being routed externally.

#### Any relevant tickets

#991 
Closes #948

Locally


config/default.js
```diff
@@ -42,11 +42,13 @@ module.exports = {
     sha: '',
     ref: '',
   },
-  elife: {
-    isPublic: true,
+  server: {
     api: {
       url: 'https://api.elifesciences.org/',
```
giorgiosironi: this goes [through the CDN](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L431) so to get a similar experience to the other environments you should point to [https://prod--gateway.elifesciences.org](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L379)

config/end2end.js
```diff
@@ -10,9 +10,9 @@ module.exports = {
       disableUpload: true,
     },
   },
-  elife: {
+  server: {
     api: {
-      url: 'https://end2end--gateway.elifesciences.org/',
+      url: 'http://end2end--gateway.elife.internal',
```
giorgiosironi: some of the `server.api.url` values have a trailing `/`, some haven't, not sure if this is alright

diversemix: Ooo - good spot - the tests did work but I think I will change them just to make sure!

diversemix:
@damnhipster ?

PR #1035, opened 2018-11-14
Add selectors for file upload result

giorgiosironi:
Primarily for Selenium end-to-end testing to understand when the file has finished uploading, but only the `completed` value is necessary there.

https://github.com/elifesciences/elife-xpub/blob/a8e5bc83de9021c72500a3167a6ad5d0d84e497f/app/components/ui/molecules/FileUpload.test.js#L29 could rely on these to check the various states instead of on the text; or use them in addition to the text.

https://github.com/elifesciences/elife-xpub/blob/develop/test/submission.browser.js#L85 has a bare sleep that could also rely on a waiting condition like this instead, to improve stability.

### DIFF
--- a/app/components/ui/molecules/FileUpload.js
+++ b/app/components/ui/molecules/FileUpload.js
@@ -85,7 +85,10 @@ const DropzoneContent = ({ conversion, formError, dropzoneOpen }) => {
     return (
       <React.Fragment>
         <StyledUploadIcon percentage={conversion.progress} />
-        <UploadInstruction data-test-id="dropzoneMessage">
+        <UploadInstruction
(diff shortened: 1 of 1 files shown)


diversemix:
@Cooryd ... you may want to be aware of this as you are working in the same area?

Cooryd:
I'll add the same data tags to my work just incase this doesn't make it to my merged in code. I've renamed, moved and refactored this file in my PR so wouldn't want to see these get lost

giorgiosironi:
All naming is up for grabs, I just need a `data-*` attribute to select on but it could be any name and value.

Issue #745, opened 2018-10-08
Allow user to delete an in-progress submission


hdrury1:
**User Story:**

_As an author I want to delete a submission that I have not completed because I know longer wish to submit to eLife._

_As an author I want to delete a submission that I have not completed because I want to tidy up my dashboard._

**Product requirements/Acceptance criteria:**

Functional:

- Clicking on a trash can should remove that entry from a user's dashboard
- Entries below the deleted entry move up one space in the list
- A user should only be able to delete a submission with the 'in progress' status
- Below the 768px breakpoint dashboard list items should not contain the trash can icon
- A user should not have the option to delete a submission in a viewport of 768px and below

**Designs**

Above 768px:

<img width="927" alt="screen shot 2018-11-07 at 16 06 21" src="https://user-images.githubusercontent.com/17023702/48143741-5e0b1300-e2a7-11e8-929c-105356da3206.png">

<img width="910" alt="screen shot 2018-11-07 at 16 08 07" src="https://user-images.githubusercontent.com/17023702/48143754-63685d80-e2a7-11e8-99f0-38101b567069.png">

768px and below (same design as currently):

<img width="393" alt="screen shot 2018-11-07 at 16 09 56" src="https://user-images.githubusercontent.com/17023702/48143850-9ad70a00-e2a7-11e8-96b4-4bb2f9c63ed0.png">

jsms90:
Just to make sure we're on the same page, "In progress" = "Continue submission" now, right?

PR #1043, opened 2018-11-15
745 remove submission

jyenes:
#### Background
allow user to remove submission in progress

What does this PR do?
closes: #745


jyenes:
pending tasks:

- [ ] front end is now broken (incorrect displayed)
- [ ] remove the file from S3
- [ ] unit testing
- [ ] e2e testing

jyenes:
@hdrury1 what happen with the items that have been submitted and it's not possible to be deleted

look at this: 

screenshot 2018-11-15 at 12 43 04

hdrury1:
According to designs this is the expected behaviour (i.e. that the content is mis-aligned due to absence of the trash icon) but I'd welcome @chugginselifesciences input. In the screenshot you posted though @jyenes it looks like there is a divider missing between the first and second items in the list.

**Designs**

screen shot 2018-11-15 at 13 28 39

chugginselifesciences:
> According to designs this is the expected behaviour but I'd welcome @chugginselifesciences input. In the screenshot you posted though @jyenes it looks like there is a divider missing between the first and second items in the list.
> 
> **Designs**
> 
> screen shot 2018-11-15 at 13 28 39

@hdrury1 yes it looks like the "submitted" manuscript is missing the divider, which would then match the designs. But with many submissions I agree that alignment is going to be a problem, so I will update the designs so that submitted manuscripts' list items are aligned but don't have the trash icon.

FYI I hope this will change in future once we no longer need to put submitted manuscripts in an "inactive" state and therefore "not deletable".

jyenes:
> ording to de

The missing divider is something we already know, we wanted to check if alignment on the items submitted was ok

hdrury1:
@jyenes Please see Zeplin for updated designs:

screen shot 2018-11-15 at 14 33 10

As you can see there is now a 'space' where the trash icon would be, meaning that the rest of the item is aligned.

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -78,31 +80,110 @@ const DateBox = styled(Flex)`
...
+            </React.Fragment>
+          ) : (
+            <React.Fragment>
+              <DashboardLink to={`/submit/${manuscript.id}`}>
+                {this.renderItemContent()}
+              </DashboardLink>
+              <TrashButton>
+                <Mutation mutation={DELETE_MANUSCRIPT}>
+                  {deleteManuscript => {
+                    console.log(manuscript)
```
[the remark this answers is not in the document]
damnhipster: System.out.println

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -78,31 +80,110 @@ const DateBox = styled(Flex)`
...
+    })
+  }
+  render() {
+    const { manuscript } = this.props
+
+    return (
+      !this.state.isDeleted && (
+        <Root py={3}>
+          {manuscript.clientStatus === 'SUBMITTED' ? (
+            <React.Fragment>
```
SemraSh: I thins it looks cleaner when you import the exports on top where you import React
```js
import React, { Fragment, Component } form 'react'

<Fragment>
 ... some kind of code here
</Fragment>

```

same for `Component` on line 123

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -78,31 +80,110 @@ const DateBox = styled(Flex)`
...
+      variables: { id: this.props.manuscript.id },
+    }).then(() => {
+      this.setState({ isDeleted: true })
+    })
+  }
+  render() {
+    const { manuscript } = this.props
+
+    return (
+      !this.state.isDeleted && (
```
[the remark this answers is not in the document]
damnhipster: I've not had to delete using a mutation in the past, and surprisingly, it's still a question which [doesn't have a good answer](https://github.com/apollographql/apollo-client/issues/899)

I think @Cooryd's first option, where we manually refresh the manuscripts list and refetch is the best option, since [it's not advised to mess around with the Apollo cache at the moment](https://github.com/apollographql/apollo-client/issues/621).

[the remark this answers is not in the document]
Cooryd:
```
<Mutation refetchQueries={ [{ query: ALL_MANUSCRIPTS }]} /> 
```

That should do the trick. Protects us from damaging the cache but should perform a refresh in the same request as the mutation.

app/components/ui/molecules/DashboardListItem.js
```diff
@@ -78,29 +81,95 @@ const DateBox = styled(Flex)`
...
+  flex-direction: row;
+  justify-content: space-between;
+`
+
 const DashboardListItem = ({ manuscript }) => {
-  const date = new Date(manuscript.created)
+  const onClickHandler = deleteManuscript => {
+    deleteManuscript({
+      variables: { id: manuscript.id },
+      refetchQueries: [{ query: ALL_MANUSCRIPTS }],
```
diversemix: Can be a significant delay for the user - add user feedback

server/xpub-model/entities/file/index.js
```diff
@@ -49,6 +49,28 @@ class File extends BaseModel {
...
+    }
+    return file
+  }
+
+  async deleteContent() {
+    if (!this.id) {
+      throw new Error('File has no ID, or is invalid')
+    }
+
+    return s3
```
diversemix: s3 is not the responsibility of the model - this should be taken out at some point

server/xpub-server/entities/manuscript/resolvers/index.js
```diff
@@ -28,10 +28,20 @@ const resolvers = {
...
       const manuscript = await Manuscript.find(id, userUuid)
+
+      const manuscriptFile = manuscript.files[0]
+      if (manuscriptFile) {
+        try {
+          const file = await File.find(manuscriptFile.manuscriptId)
+          await file.deleteContent()
+          await file.delete()
+        } catch (error) {
+          console.log(error)
```
jyenes: logger instead of console.log

server/xpub-server/entities/manuscript/resolvers/index.js
```diff
@@ -28,10 +28,20 @@ const resolvers = {
...
-    // TODO restrict this in production
     async deleteManuscript(_, { id }, { user }) {
       const userUuid = await User.getUuidForProfile(user)
       const manuscript = await Manuscript.find(id, userUuid)
+
+      const manuscriptFile = manuscript.files[0]
+      if (manuscriptFile) {
+        try {
+          const file = await File.find(manuscriptFile.manuscriptId)
+          await file.deleteContent()
```
diversemix: Architecture ... do we call out to the different areas of concern here (i.e. model /s3) ... or do we introduce another layer (logic) that will handle the orchestrations of permissions, model, s3, etc. ?

PR #1050, opened 2018-11-15
1034 remove manuscript on error

### LINKED ISSUE #1034: Remove Stored Manuscript File on Client Side Errror
**Describe the bug**

Currently when the user attempts to replace a successfully uploaded Manuscript file with an invalid file the the client correctly displays an error message. As the data stored still has reference to the previously uploaded file, the user is then able to click the next button and still go through to the next step of the Submission Wizard. 

**Steps To Reproduce**

1. Create a new manuscript, fill in details and get to the 'Files' step of the wizard.
2. Fill in the cover letter field and upload a valid pdf or docx to the Manuscript upload `Dropzone`
3. Then attempt to drag and drop an invalid file (size over 100MB or an invalid type like png)
4. Once error state shows press next button once

**Expected behavior**

The client side validation issue should send a request to remove the existing Manuscript file from the submission. This should result in the user being unable to move forward in the wizard before uploading another valid manuscript file

Cooryd:
#### Background

When there is a client side validation error on the manuscript upload field, the Submission should delete its current Manuscript file.

What does this PR do?

- Refactors the way Manuscript files are uploaded so that there is only ever one of type `MANUSCRIPT_SOURCE` in the manuscript.files field. 
- Extends tests to check any previous manuscript file gets replaced.
- Adds a removeManuscriptUpload mutation to remove any existing manuscript file
- Adds test for removing an upload
- calls the removeManuscriptUpload mutation when a client side upload validation error occurs 

#### Any relevant tickets

close #1034 

Unit tests

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/index.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/index.js
@@ -39,42 +39,35 @@ function getProgress(loading, data) {
 }
 
 const onUploadValidationError = (
-  fieldName,
-  setFieldError,
+  formFunctions,
   deleteFile,
-  setFieldValue,
   manuscriptId,
   errorMessage,
 ) => {
-  setFieldError(fieldName, errorMessage)
+  formFunctions.setFieldError('files', errorMessage)
   deleteFile({
     variables: { id: manuscriptId },
   }).then(({ data }) => {
-    setFieldValue(fieldName, data.removeUploadedManuscript.files)
+    formFunctions.setFieldValue('files', data.removeUploadedManuscript.files)
   })
 }
 
 export const onFileDrop = (
   files,
-  fieldName,
-  setFieldTouched,
-  setFieldError,
-  setFieldValue,
+  formFunctions,
   deleteFile,
   uploadFile,
   manuscriptId,
 ) => {
   const validationError = errorMessage =>
     onUploadValidationError(
-      fieldName,
-      setFieldError,
+      formFunctions,
       deleteFile,
-      setFieldValue,
       manuscriptId,
       errorMessage,
     )
 
-  setFieldTouched(fieldName, true, false)
+  formFunctions.setFieldTouched('files', true, false)
   if (files.length > 1) {
     validationError(errorMessageMapping.MULTIPLE)
     return
@@ -87,8 +80,8 @@ export const onFileDrop = (
   uploadFile({
     variables: { file, id: manuscriptId, fileSize: file.size },
   }).then(({ data }) => {
-    setFieldValue('meta.title', data.uploadManuscript.meta.title)
-    setFieldValue(fieldName, data.uploadManuscript.files)
+    formFunctions.setFieldValue('meta.title', data.uploadManuscript.meta.title)
+    formFunctions.setFieldValue('files', data.uploadManuscript.files)
   })
 }
 
@@ -134,10 +127,11 @@ const FilesPageContainer = ({
                   onDrop={files =>
                     onFileDrop(
                       files,
-                      fieldName,
-                      setFieldTouched,
-                      setFieldError,
-                      setFieldValue,
+                      {
+                        setFieldTouched,
+                        setFieldError,
+                        setFieldValue,
+                      },
                       deleteFile,
                       uploadFile,
                       values.id,
--- a/app/components/pages/SubmissionWizard/steps/Files/index.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/index.js
@@ -5,7 +5,7 @@ import { Box } from 'grid-styled'
 import CoverLetterEditor from './CoverLetterEditor'
 import ValidatedField from '../../../../ui/atoms/ValidatedField'
 import ManuscriptUpload from './ManuscriptUpload'
-import { errorMessageMapping } from './utils'
+import { errorMessageMapping, manuscriptFileTypes } from './utils'
 
 const UPLOAD_MUTATION = gql`
   mutation UploadFile($id: ID!, $file: Upload!, $fileSize: Int!) {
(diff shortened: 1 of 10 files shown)


giorgiosironi:
Can we merge `develop` into this to cross-check all the Jenkins builds are now working correctly?

giorgiosironi:
Good, http://35.172.178.78:30868/ (from the `continuous-integration/jenkins/deploy-for-review` check) works and has a backing database.

app/components/pages/SubmissionWizard/steps/Files/index.test.js
```diff
@@ -0,0 +1,107 @@
...
+    const values = {
+      files: [],
+    }
+    const setValues = (fieldName, value) => (values[fieldName] = value)
+    const files = [
+      {
+        name: 'document1.docx',
+      },
+    ]
+    await onFileDrop(
```
damnhipster: It's hard to see what's going on here. Can we look at using an object/multiple objects instead for the parameters of `onFileDrop`?

server/xpub-server/entities/manuscript/resolvers/removeUploadedManuscript.test.js
```diff
@@ -0,0 +1,44 @@
...
+  const profileId = userData.identities[0].identifier
+  let userId
+  beforeEach(async () => {
+    await createTables(true)
+    const [user] = await Promise.all([new User(userData).save()])
+    userId = user.id
+  })
+
+  describe('removeUploadedManuscript', () => {
+    it('returns the original Manuscript when there is no file to remove', async () => {
```
damnhipster: It does this in both cases.

server/xpub-server/entities/manuscript/resolvers/removeUploadedManuscript.test.js
```diff
@@ -0,0 +1,44 @@
...
+const { User, Manuscript } = require('@elifesciences/xpub-model')
+const { Mutation } = require('.')
+const { userData } = require('./index.test.data')
+
+describe('Manuscripts', () => {
+  const profileId = userData.identities[0].identifier
+  let userId
+  beforeEach(async () => {
+    await createTables(true)
+    const [user] = await Promise.all([new User(userData).save()])
```
damnhipster: I don't think this needs `Promise.all`

server/xpub-server/entities/manuscript/resolvers/removeUploadedManuscript.js
```diff
@@ -0,0 +1,24 @@
...
+    manuscript.files.splice(manuscriptUploadIndex, 1)
+    await manuscript.save()
+    logger.debug(`Manuscript file removed`, {
+      manuscriptId: vars.id,
+      userId: userUuid,
+    })
+  }
+
+  return manuscript
+}
```
damnhipster: This looks great

app/components/pages/SubmissionWizard/steps/Files/index.js
```diff
@@ -22,10 +22,69 @@ const UPLOAD_MUTATION = gql`
...
+  formFunctions,
+  deleteFile,
+  manuscriptId,
+  errorMessage,
+) => {
+  formFunctions.setFieldError('files', errorMessage)
+  deleteFile({
+    variables: { id: manuscriptId },
+  }).then(({ data }) => {
+    formFunctions.setFieldValue('files', data.removeUploadedManuscript.files)
```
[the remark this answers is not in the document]
Cooryd: So for now it just sets the field value to be an empty array so if you navigate back then forward in the wizard you'll see the empty field state for the `Dropzone` rather than a 'Success' message. I suppose I could hardcode a `[]` but then if in the future someone introduces a bug in the mutation, the client would be out of sync with the data stored server side.

server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
```diff
@@ -110,7 +123,7 @@ async function uploadManuscript(_, { file, id, fileSize }, { user }) {
     })
   }
 
-  manuscript.files.push(fileEntity)
+  addFileEntityToManuscript(manuscript, fileEntity)
```
damnhipster: It looks like we save the file entity separately to begin with [here](https://github.com/elifesciences/elife-xpub/blob/5247cec4eacbce6bf18465c99114d4e25a1cc051/server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js#L50-L55)

`manuscript.save` will also save the file entity to the database. This looks like it was existing behaviour, but if we're expecting there to be a single `MANUSCRIPT SOURCE` on a manuscript, we'll want to rework this logic.

app/components/pages/SubmissionWizard/steps/Files/index.js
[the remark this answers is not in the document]
damnhipster: Is this to support multiple file upload?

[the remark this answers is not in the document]
damnhipster: If so, it's worth adding a test for this

damnhipster:
I wasn't able to test it because of https://github.com/elifesciences/elife-xpub/issues/1064, but it seems to work locally.

server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
Cooryd: I would say we should expect the `file` to be deleted as it is now, however I know there is a concept of manuscript versioning in the future which could mean it would be preferable for the replaced `file` to continue to exist as it'll be linked to a different `manuscript`. I'll remove the initial `File.save()` for now then.

Cooryd: `manuscript.save()` doesn't seem to automatically add the new `file` entity to the db, Errors get thrown when I remove the `.save()` when constructing the File object. Lets merge this in for now and I'll do a bit more digging into how `objection`'s save functionality works as I was also under the impression that it would create and destroy newly related entities on save.

PR #1053, opened 2018-11-15
Install Helm charts as dependencies

giorgiosironi:
Recently https://github.com/elifesciences/elife-xpub-formula/pull/57 introduced a `stable/postgresql` chart as a dependency, to make the review environments work properly providing a database. The script needs to be expanded to resolve this dependency and download it before installation.

### DIFF
--- a/scripts/helm_deploy.sh
+++ b/scripts/helm_deploy.sh
@@ -22,6 +22,6 @@ fi
 
 cd helm/
 cd elife-xpub
-helm dependency update .
+sudo -u elife -H helm dependency update .
 cd -
 sudo -u elife -H helm upgrade --install "$release_name" --set image.tag="${image_tag}" elife-xpub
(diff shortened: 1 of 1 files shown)


giorgiosironi:
This should fix PRs that can't build a review environment on Jenkins because of the `postgres` addition, and make the review environments actually working.

scripts/helm_deploy.sh
```diff
@@ -21,4 +21,8 @@ else
 fi
 
 cd helm/
+cd elife-xpub
+sudo -u elife -H helm init --client-only
+sudo -u elife -H helm dependency update .
+cd -
```
diversemix: personall I prefer `pushd / popd`

giorgiosironi: I think the ideal would be a subshell that isolates from changing the working directory of the main script... Arcane but this will be pulled out of the project when stable

PR #1070, opened 2018-11-19
Use our own db check instead of one from pubsweet

### LINKED ISSUE #1069: Noisy info-level logging
**Describe the bug**
Due to `/ping` healthchecks, an application instance with no real traffic still produces 80-100 log lines per minute.

**Steps To Reproduce**
```
elife@staging--xpub:/srv/elife-xpub$ docker-compose logs -f
...
app_1          | info: Checking if database tables exist.
app_1          | info: Database tables exist.
app_1          | info: ::ffff:172.19.0.1 [19/Nov/2018:16:12:43 +0000] GET /ping 200 - 4 111.744 ms
app_1          | 
... (repeats forever)
```

Most of this (first 2 lines) come from [`pubsweet/db-manager`](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/db-manager/src/helpers/db-exists.js) but since there is a single global [logger](https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/logger) all calls are identical.

**Expected behavior**

With no users working on the application we should probably see no logs being written, given there are no background processes. The logs listed here may be `debug`-level.

Not sure about the `GET|POST` webserver-style log line. We already log this at the nginx level so it may not be necessary in all real environments.

Is there a RFC for logging in pubsweet and principles on what should/should not be done?

diversemix:
#### Background

Remove necessary logging

#### Any relevant tickets
Closes #1069 

Locally ... healthy - missing db - healthy

### DIFF
--- a/server/xpub-server/health.js
+++ b/server/xpub-server/health.js
@@ -1,0 +1,35 @@
+const config = require('config')
+const { db } = require('pubsweet-server')
+const AWS = require('aws-sdk')
+
+const S3_API_VERSION = '2006-03-01'
+
+const s3 = new AWS.S3({
+  ...config.aws.credentials,
+  ...config.aws.s3,
+  apiVersion: S3_API_VERSION,
+})
+
+const checkS3 = () =>
+  new Promise(resolve => {
+    s3.listObjects({}, (err, response) => {
+      if (err) {
+        resolve(err)
+      } else {
+        resolve(response)
+      }
+    })
+  })
+
+const checkDataBase = () =>
+  new Promise(resolve => {
+    db.select('table_name')
+      .from('information_schema.tables')
+      .then(response => resolve(response))
+      .catch(error => resolve(error))
+  })
+
+module.exports = {
+  checkDataBase,
+  checkS3,
+}
--- a/server/xpub-server/routes.js
+++ b/server/xpub-server/routes.js
@@ -1,6 +1,6 @@
 const logger = require('@pubsweet/logger')
 const config = require('config')
-const { dbExists } = require('@pubsweet/db-manager')
+const { db } = require('pubsweet-server')
 const AWS = require('aws-sdk')
 
 const DATABASE_ERROR = 'Database Error'
@@ -37,7 +37,8 @@ const checkS3 = () =>
 
 const checkDataBase = () =>
   new Promise(resolve => {
-    dbExists()
+    db.select('table_name')
+      .from('information_schema.tables')
       .then(response => resolve(response))
       .catch(error => resolve(error))
   })
--- a/server/xpub-server/routes.js
+++ b/server/xpub-server/routes.js
@@ -1,13 +1,10 @@
 const logger = require('@pubsweet/logger')
-const config = require('config')
-const { db } = require('pubsweet-server')
-const AWS = require('aws-sdk')
+const { checkDataBase, checkS3 } = require('./health')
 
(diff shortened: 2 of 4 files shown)


server/xpub-server/routes.js
```diff
@@ -16,37 +13,13 @@ const nocache = (req, res, next) => {
   next()
 }
 
-const s3 = new AWS.S3({
-  ...config.aws.credentials,
-  ...config.aws.s3,
-  apiVersion: S3_API_VERSION,
-})
-
 let lastHealthStatus = null
```
giorgiosironi: I notice this `lastHealthStatus` is used to try to avoid too many `error` logs. I think that should not happen at all, because if the healthcheck fails (several times in a row) the node is disconnected and the load balancer stops sending traffic to it.

giorgiosironi: So could probably be "simplified out"

diversemix: So should the app shutdown rather than attempt to recover itself?

giorgiosironi: No, I'm saying it's not needed to shut down nor to track the previous health status; `/ping` can be stateless and the app can always keep running; there may be an external fix like stopping/starting the RDS database that will make it serve requests again.

diversemix: ... so in that example it would get added back into the `TargetGroup` of the LB? ... I've just worked with too many systems that can flood the logs in failure cases - just want to ensure we never do. Would you be happy if it were to go in?

server/xpub-server/routes.test.js
```diff
@@ -1,9 +1,9 @@
-jest.mock('@pubsweet/db-manager')
+jest.mock('./health')
```
damnhipster: If we're mocking `health`, it's worth writing a unit test for it.

server/xpub-server/health.js
```diff
@@ -0,0 +1,35 @@
...
+    })
+  })
+
+const checkDataBase = () =>
+  new Promise(resolve => {
+    db.select('table_name')
+      .from('information_schema.tables')
+      .then(response => resolve(response))
+      .catch(error => resolve(error))
+  })
```
damnhipster: We should add a unit test for this me thinks.

damnhipster:
We can get rid of https://github.com/elifesciences/elife-xpub/pull/1070/files#diff-6ba991657f2dfe97ffaf0dc262fe7048R13 since we're mocking out `health`

server/xpub-server/routes.test.js
diversemix: Good point!

server/xpub-server/health.js
diversemix: What would it be testing? :/

damnhipster: It's more to highlight what the result of `checkDataBase` would be, but I think you're right, it's too postgres specific.

server/xpub-server/routes.js
damnhipster: The ping check would keep being requested even if its out of the load balancer, to see if the node has recovered. Can you confirm @giorgiosironi?

giorgiosironi: Yes I think you're right. In our typical experience we turn off nginx during deploy, which explains why we do not see logs as the request is refused at the TCP level rather than being answered with a HTTP 500. So this part should be unchanged.

PR #1089, opened 2018-11-22
resume navigation to the last visited page

### LINKED ISSUE #1063: Navigation to the last visited page is broken
latest changes have removed navigation to the last visited page on continue submission

jyenes:
#### Background

What does this PR do?
closes #1063 

there's another ticket to create a proper browser test for that #1088

see #1088

#### UI Changes

- [x] Has been reviewed against Designs

### DIFF
--- a/app/components/pages/Dashboard/operations.js
+++ b/app/components/pages/Dashboard/operations.js
@@ -9,6 +9,7 @@ export const ALL_MANUSCRIPTS = gql`
         title
       }
       clientStatus
+      formState
     }
   }
 `
--- a/app/components/ui/molecules/DashboardListItem.js
+++ b/app/components/ui/molecules/DashboardListItem.js
@@ -155,7 +155,7 @@ const DashboardListItem = ({ manuscript }) => {
         </Fragment>
       ) : (
         <Fragment>
-          <DashboardLink to={`/submit/${manuscript.id}`}>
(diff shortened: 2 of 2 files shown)


SemraSh (approved):
LGTM 👍

app/components/pages/Dashboard/operations.js
```diff
@@ -9,6 +9,7 @@ export const ALL_MANUSCRIPTS = gql`
         title
       }
       clientStatus
+      formState
```
SemraSh: Maybe we can change the name to be something more explanatory, like `previousPage`, `lastVisitedPage` if that makes sense?

diversemix: `lastVisitedPage` :+1:

damnhipster: This would require a migration, which should be automatically applied, but we haven't actually run a migration since we created Production and Staging last month so there may be some manual steps needed.

And in the spirit of getting small tickets/issues through, I suggest moving this change to a separate PR/issue.

jyenes: agree on move this to a separate ticket

jyenes: so new ticket raised #1097 for refactor the field name

PR #1091, opened 2018-11-22
Update tests so dev's can run without secret

### LINKED ISSUE #1090: Allow tests to run without the secret
As part of the MECA package changes. A secret was needed to query the eLife API. The tests were written against the use of this secret. Remove the dummy secret from `config/default.js` and re-write the tests for when developers do not have the secret configured.

diversemix:
#### Background

Fix running tests locally for devs

#### Any relevant tickets

Closes #1090 

Tested locally and updated unit test

### DIFF
--- a/server/meca/file-generators/article.test.js
+++ b/server/meca/file-generators/article.test.js
@@ -1,5 +1,6 @@
 const { createTables } = require('@pubsweet/db-manager')
 const path = require('path')
+const config = require('config')
 const { db } = require('pubsweet-server')
 const generateXml = require('./article')
 const Replay = require('replay')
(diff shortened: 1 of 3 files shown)


damnhipster:
See https://github.com/elifesciences/elife-xpub/issues/1090#issuecomment-441048186

server/meca/file-generators/article.test.js
```diff
@@ -27,6 +28,7 @@ describe('Article XML generator', () => {
   it('generates expected XML', async () => {
     let xml = await generateXml(sampleManuscript)
     xml = obfuscateEmail(xml)
-    expect(xml).toMatchSnapshot()
+    if (config.get('server.api.secret')) expect(xml).toMatchSnapshot()
```
damnhipster: I'd suggest putting this into a separate test so it's not hidden behind an `if` statement.

We can mock out `config` so that there is a secret set for the test.

damnhipster: This should also mean there's no need for obsolete snapshots.

diversemix:
@damnhipster Ready for review again - now have two "custom" snapshots used for with and without a secret.

Issue #1064, opened 2018-11-19
Fix uploading manuscripts on review environments


damnhipster:
Review environments only support xpub up to the point of uploading a manuscript. Review environments need integration with an S3 endpoint for manuscript upload to function correctly.

## Tasks
- [x] create temporary test environment from Helm chart
- [x] reproduce problem and ensure the error is visible
- [x] fiddle with it adding a container simulating S3
- [x] configure app in `ci` to talk to `fakes3` as an endpoint
- [x] add a new stage to the formula build that does some validation

damnhipster:
Closed in favour of #1100

damnhipster:
Reopened to be implemented on Jenkins rather than Gitlab

giorgiosironi:
https://alfred.elifesciences.org/job/test-elife-xpub/537/ is broken on `elife-xpub--ci`, I think because the `ci` environment is overloaded with running the Helm chart and also this stack. To be fair we could drop `elife-xpub--ci`, cutting one stack from the pipeline and relying on `end2end` and `staging` as testing environments.

`elife-xpub--ci` only runs some smoke tests, and `elife-xpub` tests are run on `develop` by the oddly named https://alfred.elifesciences.org/job/test-elife-xpub-source-repository/ so `end2end` would become the first place where we test `elife-xpub-deployment`+`elife-xpub-formula`. Thoughts?

diversemix:
Sounds reasonable to me. Thinking about the testing strategy... this sounds like a good place to run integration tests too. Or do we want them running in CI ?

damnhipster:
Integration tests on CI still sound reasonable to me, since we can clear state easily enough when running local mocks. 

The review environment will be using a combination of real endpoints and containerised fakes, which we'll probably need to have a way to setup and destroy/clear.

Or we rewrite the tests so the test data is unique for each test, which feels like a lot of work.

giorgiosironi:
> integration tests

What is the definition of "integration" in here? `end2end` and `staging` do integration with everything we can, from real services (EC2 instances, ELB, S3 buckets, RDS, Fastly) to other projects (`journal`, `api-gateway`, `journal-cms`).

https://alfred.elifesciences.org/job/test-elife-xpub-source-repository/ (driven by `elife-xpub`) seems to cover running all the tests that are in the project, it could also produce a `develop` review environment identical to the PRs one if we want.
[...]

PR #1117, opened 2018-11-26
Allow S3_ENDPOINT environment variable

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/1064

Useful to get `app` to talk to the `fakes3` container rather than a real S3 bucket. We would run `fakes3` in CI, too, despite not using it by default for now.

Shouldn't affect tests as they have a `test` environment that is separate from `ci` anyway. `ci` would be used to run pull request review environments.

Probably `ci` is used also for `elife-xpub--ci` usage in the `develop` branch, but right now the environment variable should be empty so it should be irrelevant.

### DIFF
--- a/.env.ci
+++ b/.env.ci
@@ -8,6 +8,8 @@ PGUSER=test
 PGPASSWORD=pw
 PGDATABASE=test
 NODE_ENV=test
+NODE_CONFIG_ENV=ci
+PUBSWEET_SECRET=fakesecret
 BROWSER=chrome:headless --no-sandbox
 
 # setup data for postgres image
--- a/.env.ci
+++ b/.env.ci
@@ -13,3 +13,6 @@ BROWSER=chrome:headless --no-sandbox
 # setup data for postgres image
 POSTGRES_USER=test
 POSTGRES_PASSWORD=pw
+
+# configure app to access fakes3
+S3_ENDPOINT=http://fakes3:4569
--- a/.env.ci
+++ b/.env.ci
@@ -13,6 +13,3 @@ BROWSER=chrome:headless --no-sandbox
 # setup data for postgres image
 POSTGRES_USER=test
 POSTGRES_PASSWORD=pw
-
-# configure app to access fakes3
-S3_ENDPOINT=http://fakes3:4569
--- a/config/ci.js
+++ b/config/ci.js
@@ -3,7 +3,10 @@ const AWS = require('aws-sdk')
 module.exports = {
   aws: {
     s3: {
-      endpoint: new AWS.Endpoint('http://fakes3:4569'),
+      endpoint: new AWS.Endpoint(
+        // randomise port to avoid conflicts in parallel test runs
+        `http://localhost:${Math.floor(Math.random() * 50000) + 15000}`,
+      ),
     },
   },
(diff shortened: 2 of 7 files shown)


giorgiosironi:
I cannot get this to run locally, which is not a good sign.

damnhipster:
That part of the config expects an `AWS.Endpoint` instance.

https://github.com/elifesciences/elife-xpub/blob/develop/config/test.js#L33-L35

giorgiosironi:
Good example, thanks. I'll also sort out how the app can be started in Jenkins with the `docker-compose.ci.yml` configuration that we use for tests, since at the moment it can't and it breaks the testing of this.

diversemix:
Failing with:
```
FAIL server/xpub-server/entities/manuscript/resolvers/uploadManuscript.test.js (12.803s)
  ● Manuscripts › uploadManuscript › fails if manuscript doesn't belong to user
    CredentialsError: Missing credentials in config
      at Immediate._onImmediate (node_modules/aws-sdk/lib/http/node.js:83:34)
  ● Manuscripts › uploadManuscript › fails if manuscript doesn't belong to user
    TypeError: Cannot read property 'close' of undefined
      46 | 
      47 |     afterEach(done => {
    > 48 |       s3Server.close(done)
         |                ^
      49 |     })
      50 | 
      51 |     it("fails if manuscript doesn't belong to user", async () => {
```

Probably need these creds: https://github.com/elifesciences/elife-xpub/blob/d9264b1f87088407cc52606fe18c6b297c7d16ca/config/test.js#L24

diversemix:
@giorgiosironi ... you might find this command helpful:
```
NODE_ENV=ci node -e "const c = require('config'); console.log(JSON.stringify(c.get('aws'), null, 4));"
```
(running in the root folder where `package.json` lives)

diversemix:
ok ... @giorgiosironi ... looks like the `uploadManuscript.test.js` actually starts its own server and so having a fakes3 one conflicts ...see: https://github.com/elifesciences/elife-xpub/blob/b82a4896b3b7195338b33a2f29ca226c5e580791/server/xpub-server/entities/manuscript/resolvers/uploadManuscript.test.js#L40

damnhipster:
#1123 may help with this

diversemix:
@damnhipster Can you specify how that would help? This turned out not to be about config.

damnhipster:
I'm able to reproduce the issue locally by running `NODE_ENV=ci yarn test server/meca/export.test.js`. Tests are expected to be run using `NODE_ENV=test`. We can hardcode this.

damnhipster:
> @damnhipster Can you specify how that would help? This turned out not to be about config.

It's hard to tell whether we're loading `config/test.js` which is necessary for the tests to pass. I'm not sure if `config/ci.js` is overriding things or if `config/test.js` is loaded at all. My theory is that it's the latter.

> ok ... @giorgiosironi ... looks like the `uploadManuscript.test.js` actually starts its own server and so having a fakes3 one conflicts ...see:

The server that gets started is expected to be on a random port as per `config/test.js`.

diversemix:
@giorgiosironi ... pushed a fix.

PR #1123, opened 2018-11-27
Logging config sources on startup

hem-brahmbhatt:
#### Background

We've struggled in the past to see what config has been loaded and has been the source of failing builds.

This change makes it easier to see what config has been loaded during the following processes:
- `yarn server`
- `yarn test`
- `node app.js` which is used to start the server across our environments

We should see logs containing the config files loaded, e.g.
screen shot 2018-11-27 at 16 18 32

Please indicate the need for this change to be tested. If not please justify, if so then at what levels. **Reviewers** ensure that these test requirements are also reviewed.

- [ ] Browser tests
- [ ] End2end tests


diversemix:
Which issue does this relate to?

/cc @hdrury1

package.json
```diff
@@ -35,6 +35,7 @@
     "start:styleguide": "docker-compose run --no-deps -p 6060:6060 app yarn run styleguide",
     "stop:services": "docker-compose down -v",
     "styleguide": "styleguidist server",
+    "pretest": "NODE_ENV=test log-config-sources",
```
giorgiosironi: can you post a sample output of this command?

giorgiosironi: (sample log, more than output)

server/xpub-server/index.js
```diff
@@ -1,5 +1,9 @@
 const { merge } = require('lodash')
 const fs = require('fs')
+const logger = require('@pubsweet/logger')
+const { logConfigSources } = require('@elifesciences/util')
+
+logConfigSources(logger.info)
```
giorgiosironi: reasonable log level

package.json
giorgiosironi: ah it's using `console.log` in this case, so actual output

server/util/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@elifesciences/util",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "ISC",
```
diversemix: If we were to release this we need to change the license

damnhipster:
Closing until after initial release.

PR #1131, opened 2018-11-29
Fix failing e2e test

Cooryd:
#### Background

End 2 End test is failing as the test data being returned from the person api does not contain `research` data. This is not a required field according to the api schema so the xpub `elife-api` should be able to handle uesrs with the minimum required data.

What does this PR do?

Casts any undefined `research` value within person data to an empty object to prevent the `elife-api` from throwing an error.

Extends `elife-api.test.js` to run a test against the minimum required details that may be returned from the person api.

### DIFF
--- a/server/xpub-model/entities/user/helpers/elife-api.js
+++ b/server/xpub-model/entities/user/helpers/elife-api.js
@@ -14,7 +14,7 @@ const request = (endpoint, query = {}) => {
     req.header.Authorization = secret
   }
   return req.query(query).catch(err => {
-    logger.error('Failed to fetch from eLife API:', err.message)
+    logger.error('Failed to fetch from eLife API:', err.message, err.stack)
     throw err
   })
 }
--- a/server/xpub-model/entities/user/helpers/elife-api.js
+++ b/server/xpub-model/entities/user/helpers/elife-api.js
@@ -20,13 +20,7 @@ const request = (endpoint, query = {}) => {
 }
 
 const convertPerson = apiPerson => {
-  const {
-    id,
-    name = {},
-    research = {},
-    emailAddresses,
-    affiliations,
-  } = apiPerson
+  const { id, name, research = {}, emailAddresses, affiliations } = apiPerson
   const { focuses = [], expertises = [] } = research
 
   let person = {
--- a/server/xpub-model/entities/user/helpers/elife-api.js
+++ b/server/xpub-model/entities/user/helpers/elife-api.js
@@ -13,16 +13,20 @@ const request = (endpoint, query = {}) => {
   if (secret) {
     req.header.Authorization = secret
   }
-  return req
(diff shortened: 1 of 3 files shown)


server/xpub-model/entities/user/helpers/elife-api.js
```diff
@@ -13,16 +13,20 @@ const request = (endpoint, query = {}) => {
...
     req.header.Authorization = secret
   }
-  return req
-    .query(query)
-    .catch(err => {
-      logger.error('Failed to fetch from eLife API:', err.message)
-      throw err
-    })
+  return req.query(query).catch(err => {
+    logger.error('Failed to fetch from eLife API:', err.message)
```
giorgiosironi: can we log anything more than `err.message`, does the object have a stacktrace or line or file?

server/xpub-model/entities/user/helpers/elife-api.js
```diff
@@ -13,16 +13,20 @@ const request = (endpoint, query = {}) => {
...
+    throw err
+  })
 }
 
 const convertPerson = apiPerson => {
-  const { id, name, research, emailAddresses, affiliations } = apiPerson
+  const {
+    id,
+    name = {},
```
giorgiosironi: wasn't `name` considered mandatory?

server/xpub-model/entities/user/helpers/elife-api.js
damnhipster: The object does have a stacktrace, we should be able to log that.

server/xpub-model/entities/user/helpers/elife-api.test.js
```diff
@@ -15,48 +15,57 @@ jest.mock('config', () => ({
...
     expect(header.Authorization).toBe('xpubsecret')
   })
 
   it('creates the correct person structure', async () => {
+    request.get.mockImplementation(makeGetResponse(person.allDetails))
+    const result = await api.people()
+    expect(result).toHaveLength(1)
+    expect(result[0].firstname).toBe(person.allDetails.name.givenNames)
+    expect(result[0].surname).toBe(person.allDetails.name.surname)
+    expect(result[0].email).toBe(person.allDetails.emailAddresses[0].value)
```
giorgiosironi: should check the fields in which `research` and `affiliations` end up?

server/xpub-model/entities/user/helpers/elife-api.js
diversemix: Can we have a unit test for `convertPerson`

server/xpub-model/entities/user/helpers/elife-api.js
Cooryd: I have included `err.stack` when logging the error now

Issue #1125, opened 2018-11-27
Write tests for new PersonPod modal - full person info


jsms90:
Write tests for Person Info modal (created in issue #635 via PR #1119)

#### Accessing the info modal via the PeoplePickerControl
Renders correctly:
- [ ] When author clicks on the info icon from the editors page (i.e. inside PeoplePickerControl), the modal should display 'Remove editor'

Functionality:
- [ ] Given that an author has opened the info modal from within the PeoplePickerControl, when an author then clicks 'Remove editor', the relevant person pod should be removed from the control
- [ ] When an author clicks 'Cancel', the modal closes and the PeoplePickerControl is unchanged

#### Accessing the info modal via the PeoplePickerModal
Renders correctly:
- [ ] Given that the author has opened the PeoplePickerModal, when the info button is clicked on a PersonPod that has **not** yet been selected, the modal should have '**Add Editor**' text
- [ ] Given that the author has opened the PeoplePickerModal, when the info button is clicked on a PersonPod that **has** been selected, the modal should have '**Remove Editor**' text
- ~Given that maximum number of editors have already been selected in the PeoplePickerModal **and** person has already been selected, when the info button is clicked, the modal should have '**Remove Editor**' text~ - to be completed as part of issue #1128 

Functionality:
- [ ] Given that ... (see 1st point above) ..., when info is clicked, then Add Editor is clicked, modal should close & person should be selected (i.e. tag should appear and icon type should change to a tick)
- [ ] Given that ... (see 2nd point above) ..., when info is clicked, then Remove Editor is clicked, modal should close & person should be deselected (i.e. tag should disappear and icon type should change to +)
- ~Given that maximum number of editors have already been chosen, Add Editor button in modal should be disabled~ - to be completed as part of issue #1128

hdrury1:
@jsms90 in the instance where a user has already reached the maximum number of selections and opens the info modal, the 'Add Editor' button should be disabled. i.e. unclickable and 50% opacity.

PR #1132, opened 2018-11-29
People picker refactor

### LINKED ISSUE #1023: People picker not rendering in correct context
**Describe the bug**
After excluding an editor and then returning to add another suggested editor, the People Picker adds the new editor to the list of exclusions rather than suggestions.

**Steps To Reproduce**

1. Go to https://staging--xpub.elifesciences.org
2. Begin a new submission
3. Navigate to the editors step
4. Select 2 suggested Senior editors
5. Select 1 excluded Senior editor
6. Attempt to choose a third suggested Senior editor
7. Observe that People Picker opens with exclusions

**Expected behavior**
People Picker should open with suggestions

**Screenshots**

screen shot 2018-11-13 at 08 29 53

screen shot 2018-11-13 at 08 31 33

screen shot 2018-11-13 at 08 32 43

Macbook Pro, High Sierra, Firefox

jsms90:
#### Background
Hopefully this should make the picker easier to reason about and therefore easier to maintain, easier to debug & easier to write tests for :crossed_fingers: 

This started while I was breaking `PersonPod` up into smaller pieces, to make it easier to test in #1125.

But working with Semra on adding the modal into the PersonPod, really highlighted how confusing the current separation of components is to jump into:
- It's very unclear from styleguidist which component is responsible for which thing (especially the separation between the `PeoplePicker` & `PeoplePickerModal`)
- Some of the naming adds confusion (e.g. `PersonPod.SelectButton` - this is not a button, but rather a 2nd type of pod, `isSelected` is a function which toggles selection not a boolean value, etc.)
- The files are all spread across the codebase in multiple different places

Given how many bug tickets are being raised on the PeoplePicker right now, including problems which will not allow the user to complete a submission, this seemed like a good time to clean up.

#### What does this PR do?
- Passes `modalName` prop into each `PeoplePickerControl` & down into `ModalHistoryState` in order to differentiate between the 4 pickers on the EditorsPage  - fixes #1023  
- Moves all components that make up the People Picker under the same folder: `/ui/PeoplePicker` (taking the `PeoplePickerModal` & `PeoplePickerControl` out of the submission wizard will also make it easier to add into the staff dashboard when we get to that point)
- Separates PersonPod into 4 separate components, living in separate files - PodContainer, PersonPod & ChooserPod (used to be called PersonPod.SelectButton) & PersonInfoModal (modal bit used to be in PersonPod & info bit used to be its own component)
- Alters the way that a PersonPod's icon is changed - now a string passed into PodContainer 
- Moves ModalOverlay into same file as ModalHistoryState, so that responsibility for opening/closing the people picker modal lives in one component (PeoplePickerControl) instead of 2
- Separates large PeoplePicker file into: `PeoplePickerButtons`, `PeoplePickerBody` & `PersonPodGrid` **this is especially useful for rectifying the performance issues**
- Creates styleguidist files for all new components
- Adds prop types to all files that were missing them


app/components/ui/PeoplePicker/PeoplePickerLogic.js
```diff
@@ -0,0 +1,123 @@
...
+      toggleSelection: person => this.toggleSelection(person),
+    })
+  }
+}
+
+const peopleArrayPropType = PropTypes.arrayOf(
+  PropTypes.shape({
+    id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+    name: PropTypes.string.isRequired,
+    aff: PropTypes.string.isRequired,
```
Cooryd: Does this propType make the following ternary unnecessary?
https://github.com/elifesciences/elife-xpub/blob/c6817a581f39f197a159205254db63048d174c8b/app/components/ui/PeoplePicker/PeoplePickerLogic.js#L75

Cooryd: Also, is there any chance the app may correctly encounter a Person entry that doesn't have a defined value for `aff`, `expertise` or `focuses`?I know these aren't required fields in the Person elife-api schema.

PR #1135, opened 2018-11-29
Change upload progress and sciencebeam timeout

### LINKED ISSUE #839: Extend ScienceBeam extraction timeout and fold feedback into manuscript upload
See [this Slack thread](https://elifesciences.slack.com/archives/C6LKMFK5Z/p1539268174000100) for context.

- We should extend the timeout of extraction to 6 seconds, as currently many manuscripts are displaying a 'Successful' upload without extracting the title
- We should try and embed extraction into the file upload progress: up to 50% for manuscript upload, 50%-100% for ScienceBeam extraction.
- We should avoid displaying the 100% upload for an extended period of time, as this will confuse users
- 'Success' should only display when 1) the manuscript has uploaded and 2) extraction has occurred OR when 1) the manuscript has uploaded and 2) ScienceBeam has timed out.

diversemix:
#### Background

- Increased Sciencebeam timeout
- Used a time-based prediction method for upload progress - see #839 for the maths

#### Any relevant tickets

Closes #839 

Tested locally - already have the end2end and browser tests

### DIFF
--- a/server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
+++ b/server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
@@ -95,6 +95,7 @@ async function uploadManuscript(_, { file, id, fileSize }, { user }) {
   } catch (err) {
     logger.error(`Manuscript was not uploaded to S3: ${err}`)
     await fileEntity.delete()
+    clearInterval(handle)
     throw err
   }
 
--- a/server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
+++ b/server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
@@ -4,7 +4,6 @@ const logger = require('@pubsweet/logger')
 const request = require('request-promise-native')
 const { promisify } = require('util')
 const xml2js = require('xml2js')
-const lodash = require('lodash')
 
 const { ON_UPLOAD_PROGRESS } = pubsubManager.asyncIterators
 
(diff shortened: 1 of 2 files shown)


server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
```diff
@@ -55,18 +55,26 @@ async function uploadManuscript(_, { file, id, fileSize }, { user }) {
   }).save()
 
   const pubsub = await pubsubManager.getPubsub()
-  const reportProgress = lodash.throttle(progress => {
+
+  // Predict upload time
+  const predictedTime = 5 + 4.67e-6 * fileSize
+  const startedTime = Date.now()
+
+  const handle = setInterval(() => {
```
Cooryd: Maybe to make the percentage more accurate (and account for varying upload speeds on the clients side) we could leave the call to this within the `stream.on` but do `Math.floor(progress * 50)` instead then set the interval timer once the files are uploaded to the S3 bucket? That way the first 50% progress is truly representational of the upload progress and the estimation only needs to guess ScienceBeams processing time.

diversemix: Did you read the comments on the issue?

diversemix: Let's  talk f2f... we couldn't go with this approach as most of the time is uploading to S3

PR #1143, opened 2018-11-30
People picker refactor

damnhipster:
Reintroduce the people picker refactor.

This was initially introduced in #1132 
This was initially reverted in #1142


app/components/ui/PeoplePicker/types.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types'
```
damnhipster: We need to ensure the prop types here actually need to be required.

PR #1147, opened 2018-11-30
People picker refactor (2)

### LINKED ISSUE #1141: Change people picker headings from "Choose" to "Suggest" when suggesting and "Exlude" when excluding
The term "choose" may be confusing while authors make suggestions for Senior and Reviewing Editors via the people picker. We should continue using the term "Suggest" instead of "Choose" so that users are clear that it is a suggestion, not a definite selection.

image

Exclude a Senior Editor (can only exclude 1)

image

Exclude Reviewing Editors (2 allowed - 1 min)

image

(please ignore button inactive states above)

jsms90:
PR https://github.com/elifesciences/elife-xpub/pull/1132 + a few fixes for the console errors

Also fixes #1141

### DIFF
--- a/app/components/ui/PeoplePicker/PeoplePickerLogic.js
+++ b/app/components/ui/PeoplePicker/PeoplePickerLogic.js
@@ -15,7 +15,7 @@ class PeoplePickerLogic extends React.Component {
   }
 
   toggleSelection(person) {
-    if (this.isSelected(person)) {
+    if (this.select(person)) {
       this.setState({
         selection: this.state.selection.filter(p => p.id !== person.id),
       })
@@ -30,7 +30,7 @@ class PeoplePickerLogic extends React.Component {
     }
   }
 
-  isSelected(person) {
+  select(person) {
     return this.state.selection.some(p => p.id === person.id)
   }
 
@@ -90,7 +90,7 @@ class PeoplePickerLogic extends React.Component {
       searchOptions,
       filterFunction: this.filterPeople,
       getMatchIndex: this.getMatchIndex,
-      isSelected: person => this.isSelected(person),
+      isSelected: person => this.select(person),
       isValid: this.isValid(),
       selection: this.state.selection,
       onSubmit: () => this.handleSubmit(),
--- a/app/components/ui/PeoplePicker/types.js
+++ b/app/components/ui/PeoplePicker/types.js
@@ -9,8 +9,6 @@ export const personNamePropType = PropTypes.string
 
 export const affiliationPropType = PropTypes.string
 
-export const affiliationsPropType = PropTypes.arrayOf(affiliationPropType)
-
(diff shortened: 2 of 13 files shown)


app/components/ui/PeoplePicker/types.js
```diff
@@ -9,18 +9,16 @@ export const personNamePropType = PropTypes.string
...
 
 export const personPropType = PropTypes.shape({
   id: personIdPropType.isRequired,
   name: personNamePropType.isRequired,
-  aff: affiliationsPropType,
-  expertises: expertisesPropType,
-  focuses: focusesPropType,
+  aff: affiliationPropType,
+  expertises: expertisesPropType.isRequired,
+  focuses: focusesPropType.isRequired,
```
Cooryd: `expertises` & `focuses` still have `isRequired` properties. Has something changed elsewhere that I'm missing to cast these to empty arrays?

jsms90: The shape of the object we create on the server side will always have empty arrays on

```
research: {
  focuses: []
  expertises: []
}
```

https://github.com/elifesciences/elife-xpub/blob/develop/server/xpub-model/entities/user/helpers/elife-api.js#L23-L24

I think we should change the server-side code so this isn't the case, but if we do that it's prob worth doing in a separate issue.

app/components/ui/PeoplePicker/PeoplePickerLogic.js
```diff
@@ -30,7 +30,7 @@ class PeoplePickerLogic extends React.Component {
     }
   }
 
-  isSelected(person) {
+  select(person) {
```
damnhipster: This returns a boolean, and doesn't actually mutate anything. `isSelected` seems to be the correct name.

jsms90: Yeah I totally agree that `isXYZ` is the perfect name for a boolean value. But I think it's useful not to conflate the function which returns a boolean, with the return value itself.

[Line 93, below](https://github.com/elifesciences/elife-xpub/pull/1147/files/c4a5729b5d10a8fed94ef8f0367975485c11d1d9#diff-58e04d7ad244eea02c2d937aec254b5eR93) kind of shows what I mean:
`isSelected` is the return value of calling the `select` function with `person` as an input value

So there's a clear separation between the thing which generates the boolean (select) & the return value which _is_ a boolean (isSelected).

Does that make sense? Or am I missing something?

damnhipster: - Pure functions, like `stringify` describe what the return value of the function would be
- Mutator functions like `update` have less emphasis on their names and more emphasis on the mutation being made 
- Accessor functions like `isConnected` are named after the state being queried for, and aren't expected to have any side effects

In this case, we're working with the third option. I see that the function is `PeoplePickerLogic.isSelected` which is adding to the confusion. `isPersonSelected(person)` would be clearer.

PR #1149, opened 2018-11-30
Remove footer

### LINKED ISSUE #1148: Remove Footer
Chris has asked to remove the footer.

The information is in the header in production builds so I'm happy to do this.

diversemix:
#### Background

Removes the footer with date and commit sha

#### Any relevant tickets

Closes #1148

Tested locally - none needed.


damnhipster:
Is this something we want to leave in place for local development and review environments?

diversemix:
We could ... but maybe we should use the same technique for inclusion as `index-production.js`

damnhipster:
> We could ... but maybe we should use the same technique for inclusion as `index-production.js`

We already put the date and SHA within the page for production, so no need to do anything further.

screen shot 2018-11-30 at 15 38 09

damnhipster:
This change means that there's no footer at all on xpub, so the buttons are right at the bottom of the page. You can see this in the review environment under the checks above

screen shot 2018-11-30 at 15 40 48

@hdrury1 do we have any plans around the footer for the page? What should we be doing in the meantime?

diversemix:
@damnhipster do you have more detailed info like screenshots? It all looked to me. I suggest getting this I  and raising another PR if there are issues

PR #1151, opened 2018-11-30
WIP (not mergeabel): browser tests for info modal

jsms90:
@Cooryd I've created a branch off of the one that the other test PR is created from, since I don't think this addition quite passes the browser tests yet. Also the TO DO comments are still in here.

I'm just pushing this up so you can work off of it on Monday, or simply use it as a reference point :+1:


jsms90:
Oops. Didn't mean to make a PR to develop

PR #1167, opened 2018-12-05
Allow email to be configured with env vars 

### LINKED ISSUE #998: Refactor the way mail is configured
This needs to be accessible to be configured within `config/custom-environment-variables.js`

Anyone with type `leadership` ... the 4 poeple not appearing are: Randy, Detlef, Eve and Anna

@diversemix to complete

diversemix:
#### Background

Tried a number of ways but this is the least friction (take a look at #998)  - probably want a pubsweet change at some point.... I dunno.

#### Any relevant tickets

Closes #998 

This is only part of the configuration setup - this needs to be tested in the pipelines

### DIFF
--- a/config/non-serializable/mailer.js
+++ b/config/non-serializable/mailer.js
@@ -1,6 +1,6 @@
 module.exports = {
   transport: {
-    host: `${process.env.MAILER_HOST}`,
-    port: `${process.env.MAILER_PORT}`,
+    host: `${process.env.MAILER_HOST || 'localhost'}`,
+    port: `${process.env.MAILER_PORT || 25}`,
   },
 }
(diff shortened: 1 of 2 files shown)


config/non-serializable/mailer.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   transport: {
-    host: 'localhost',
-    port: 25,
+    host: `${process.env.MAILER_HOST}`,
```
giorgiosironi: both should have a default value?

diversemix: That might be kinder for the mis-configured cases. :+1:

Issue #939, opened 2018-10-31
Port Styleguide onto Jenkins


giorgiosironi:
Splitted from https://github.com/elifesciences/elife-xpub/issues/675

Simple follow-up to https://github.com/elifesciences/elife-xpub/issues/408 where the target would be the same (a S3 bucket acting as a list of static websites, one per PR); the orchestration would be done on the Jenkins side to further empty out what Gitlab CI does.

giorgiosironi:
Styleguide uses the [Docker image of the project](https://github.com/elifesciences/elife-xpub/blob/develop/.gitlab-ci.yml#L132), but needs to credentials to be passed in, in some way.

giorgiosironi:
https://github.com/elifesciences/elife-xpub/pull/1168 ready for review.

giorgiosironi:
https://github.com/elifesciences/elife-xpub/pull/1172 is the follow up for `develop`.

PR #1168, opened 2018-12-05
Port Styleguide on Jenkins

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/939

Jenkins nodes have their own credentials and `aws` cli preinstalled, so they can easily deal with this use case.


giorgiosironi:
The commit status leads to https://s3.amazonaws.com/ci-elife-xpub-styleguide/1168/index.html now, can someone check it works as expected?

damnhipster:
Can we run this for the main branch as well? i.e. `develop`

giorgiosironi:
> Can we run this for the main branch

Yes, I didn't want to clash with the Gitlab version so I'll do that with a follow-up PR.

giorgiosironi:
> Do we have a similar (or the same) task to tear these down as we do for review environments?

We could, but usually (e.g. `pattern-library`) don't bother as there is no resource limit on S3 and the cost of keeping them there is very low.

Issue #419, opened 2018-07-31
Increase number of editors displayed in People Picker


tamlyn:
**User Story:**

_As an author or editor I want to be able to scroll through a long list of people in order to find the person I want._

**Product requirements/Acceptance criteria:**

Functional:
- Currently the people picker gets quite slow so an artificial limit of 30 results has been imposed
- Remove this limit and ensure rendering stays acceptable with a long (>300) list of people 
- This may just be a case of optimising the rendering or implementing a virtual scrolling behaviour

tamlyn:
Did a bit of profiling and couldn't see any obvious performance problems in the components. Each person pod takes a few milliseconds to render but when you render 300 of them it all adds up. 

![image](https://user-images.githubusercontent.com/115310/43525597-2e655dcc-959a-11e8-854b-e281de99d7bd.png)

You can optimise them to only re-render when needed (e.g. with [onlyUpdateForKeys](https://github.com/acdlite/recompose/blob/master/docs/API.md#onlyupdateforkeys) which helps but initial render is still slow.

Next avenue of exploration should probably be [react-virtualized](https://github.com/bvaughn/react-virtualized).

diversemix:
I know this is not about rendering - but I noticed the people picker makes two graphql queries one with role `senior-editor` and the other with `reviewing-editor` - these requests are made on the page regardless of how you have arrived (via wanting to select senior or reviewing). Is this an oversight? ... @tamlyn

tamlyn:
No, that's as designed. It simplifies the structure of the page and was deemed to be a small cost. But we could change it to load the data only when the modal is shown.

hdrury1:
Can someone confirm if this is still a to-do?

tamlyn:
Yes, this is still to do in order to remove the 30 pod display limit (and even that will be slow on older devices).

damnhipster:
### Background 

Off the back of #419, we decided to put `react-window` in place to improve the performance of the people picker when you have 

### Behaviour of `react-window`
`react-window` doesn't give us flexibility to use window scrolling to control rendering, and instead creates a new scrollable component. It also doesn't allow you to add a header or footer onto the scrollable component, so it's not ideal to render large lists which are part of a page.

You can see it in action here:

![react-window-demo](https://user-images.githubusercontent.com/2573277/49735741-084bd100-fc80-11e8-9dda-3b6548fd1061.gif)

Another issue is that we lose the responsive functionality of the people picker.

### Solution
[...]

PR #1183, opened 2018-12-11
[DEMO] 419 people picker react-window implementation

SemraSh:
This is an attempt to solve [this](https://github.com/elifesciences/elife-xpub/issues/419#issuecomment-409583197) issue using `react-window`.


damnhipster:
**Performance flamegraph**
screen shot 2018-12-11 at 15 03 38

Notice how we're rendering a lot less pods even though the user can scroll through all 300-400 editors.

SemraSh:
### Findings of implementing `react-window`

With this implementation the required functionality is working perfectly fine as @damnhipster described above, but we have difficulties with adding styles. 

#### Behaviour on the browser
react-window-attempt

##### Problems
1. The scroll is not effected outside the PeoplePickerGrid component which holds the react-window's `Grid`. 
The only reason I can scroll on the right hand side is because of this line `width={window.innerWidth}` I've added to avoid the scrolling window to be in the middle of the page. I've added that thinking the expansion will be to both sides, which was solving the problem of being able to scroll just inside that component. 

I have another attempt to implement the older version of react-window called rect-virtualized [here](https://github.com/elifesciences/elife-xpub/pull/1195), as that version has WindowScroller component which is allowing scrolling anywhere in the page and has the same behaviour as react-window.

hdrury1:
Now investigating alternatives to `react-window`, in particular paginating the API. Given what we don't know about `react-window` it is a risky dependency, and the work needed to override the styling is proving an obstacle.

PR #1184, opened 2018-12-11
WIP: recreate branch and add refresh function

### LINKED ISSUE #1162: Autosave feature broken
When uploading a supporting file the autosave does not work. See discussion in agenda 4/12/2018

https://docs.google.com/document/d/1IEYlmkzyBmxEFhn8U0kM8F5cl5OO-2jEpJmVV4VGZus/edit

diversemix:
#### Background

Currently there is a bug in the supporting file uploads mutation that means the manuscript object may contain stale data if an auto-save occurs while a file is still uploading.

As a quick fix for this (pending a potential job queue service which would allow us to queue any changes) we should refresh the object before we make changes to its data but after we have uploaded the file to s3 to make sure any other simultaneous edits are not over-written.

This work adds a refresh function to the `Manuscript` class and implements it before the manuscript save in the `supportingUpload` and `manuscriptUpload` mutations.

Note: This still leaves in bug #1162 

#### Any relevant tickets

closes #1162

Please indicate the need for this change to be tested. If not please justify, if so then at what levels. **Reviewers** ensure that these test requirements are also reviewed.

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/server/xpub-model/entities/manuscript/index.test.js
+++ b/server/xpub-model/entities/manuscript/index.test.js
@@ -358,6 +358,9 @@ const testOverwritingWithOldData = async (userId, refresh = false) => {
       title: 'Version1',
     },
   }).save()
+  expect(manuscriptV1).toHaveProperty('updated')
+  const v1Time = new Date(manuscriptV1.updated).getTime()
+
   const manuscriptV2 = await Manuscript.find(manuscriptV1.id, userId)
   manuscriptV2.meta.title = 'Version2'
 
@@ -365,6 +368,9 @@ const testOverwritingWithOldData = async (userId, refresh = false) => {
   manuscriptV3.meta.title = 'Version3'
   manuscriptV3 = await manuscriptV3.save()
   expect(manuscriptV3.meta.title).toBe('Version3')
+  const v3Time = new Date(manuscriptV3.updated).getTime()
+
+  expect(v1Time).toBeLessThan(v3Time)
 
   if (!refresh) {
     // If you are not refreshing - save() should not work and throw
--- a/server/xpub-model/entities/manuscript/index.test.js
+++ b/server/xpub-model/entities/manuscript/index.test.js
@@ -14,14 +14,23 @@ describe('Manuscript', () => {
   describe('applyInput', () => {
     it('picks only whitelisted properties', () => {
       const manuscript = new Manuscript({
-        meta: { title: 'Apples' },
+        meta: {
+          title: 'Apples',
+        },
         status: 'boo',
         teams: [],
       })
-      const input = { meta: { title: 'Oranges' }, other: 'frank' }
+      const input = {
+        meta: {
+          title: 'Oranges',
+        },
+        other: 'frank',
+      }
       manuscript.applyInput(input)
       expect(manuscript).toMatchObject({
-        meta: { title: 'Oranges' },
+        meta: {
+          title: 'Oranges',
+        },
         status: 'boo',
       })
     })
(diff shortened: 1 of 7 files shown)


hem-brahmbhatt (approved):

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -225,15 +330,54 @@ describe('Manuscript', () => {
           status: 'INITIAL',
         }).save(),
       ).rejects.toThrow())
+
+    it('does not overwrite an updated manuscript', async () => {
```
diversemix: Highlighting new tests as there is a lot of reformatting.

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -178,28 +263,45 @@ describe('Manuscript', () => {
       ).rejects.toThrow('Manuscript not found'))
   })
 
+  describe('refresh()', () => {
```
diversemix: Highlighting new tests as there is a lot of reformatting.

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -225,15 +330,54 @@ describe('Manuscript', () => {
...
+      await new Manuscript({
+        createdBy: secondUserId,
+      }).save()
       const loadedManuscripts = await Manuscript.all(userId)
       expect(loadedManuscripts).toHaveLength(1)
     })
   })
 })
+
+const testOverwritingWithOldData = async (userId, refresh = false) => {
```
diversemix: New function that test the overwriting with old data should throw if the object is not refreshed or work if it is refreshed

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -178,28 +263,45 @@ describe('Manuscript', () => {
       ).rejects.toThrow('Manuscript not found'))
   })
 
+  describe('refresh()', () => {
+    it('refreshes manuscript', async () => {
+      await testOverwritingWithOldData(userId, true)
```
damnhipster: I'm not a fan of having tests like this. It makes it difficult to see what's being tested and what's being asserted on. I use the tests a lot like documentation, and extracting tests out like this makes it difficult to see how to call a function and what the behaviour of it is.

I'm happy extracting mocks and setup code to separate functions, but the invocation of the unit under test and the assertions should be more visible for each test.

If we really want to make our own DSL for the tests to make the more DRY, we'll want to make sure the tests are still readable.

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -225,15 +330,54 @@ describe('Manuscript', () => {
...
+    )
+  } else {
+    await manuscriptV2.refresh()
+    // In memory should have been refreshed from the database
+    expect(manuscriptV2.meta.title).toBe('Version3')
+  }
+
+  // The database should still be at v3
+  const ms = await Manuscript.find(manuscriptV1.id, userId)
+  expect(ms.meta.title).toBe('Version3')
```
damnhipster: The assertions here are all good 👍

server/xpub-model/entities/manuscript/index.test.js
damnhipster: I'm not a fan of having tests like this. It makes it difficult to see what's being tested and what's being asserted on. I use the tests a lot like documentation, and extracting tests out like this makes it difficult to see how to call a function and what the behaviour of it is.

I'm happy extracting mocks and setup code to separate functions, but the invocation of the unit under test and the assertions should be more visible for each test.

If we really want to make our own DSL for the tests to make the more DRY, we'll want to make sure the tests are still readable.

hem-brahmbhatt (changes requested):

server/xpub-model/entities/manuscript/index.test.js
diversemix: Can you point out why you think this is not readable?

giorgiosironi: The given/when/then (arrange/act/assert) phases are not visible, in fact this is correlated to an `if` inside the function being called here. It could be split into pieces which would also avoid the `if`. Very pseudo-code like:
```
v1 = givenManuscriptVersion(1)
v2 = modifyManuscriptVersion(v1)
v3 = modifyManuscriptVersion(v1)
save(v3)
refresh(v2)
assertNoChanges(v2) # or any assert...(...)
```

diversemix: That makes sense ... I guess what you're asking is not to use functions in the (arrange>act>assert) categories rather than in its entirety.

diversemix:
Going to close as its been superceeded with #1213 @damnhipster / @giorgiosironi I've also addressed the testing structure as per the arrange>act>assert pattern.

PR #1191, opened 2018-12-12
Upgrade yarn to 1.10.1

damnhipster:
#### Background

We currently accept any version of yarn greater than 1.2.

1.10 introduced a new field in the `yarn.lock` file, which [gets stripped out when using earlier versions](https://github.com/yarnpkg/yarn/issues/6440)

We generally use the latest version of yarn locally, so this change will throw an error if an earlier version of yarn is used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# built from https://gitlab.coko.foundation/yld/xpub-deployment-config-postgres
 FROM xpub/xpub:base
 
 WORKDIR ${HOME}
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "private": true,
   "engines": {
     "node": ">=8.9.4",
-    "yarn": ">=1.2"
+    "yarn": "^1.10.1"
   },
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


diversemix:
Test failures looks like we have to update yarn in the docker image `xpub/xpub:base`

diversemix:
```
$ docker pull xpub/xpub:base
base: Pulling from xpub/xpub
Digest: sha256:3b10abb05a9c8a3c69adb08fbc27ff5ab83bbb5a3d49014dee61a3e2bc519fd2
Status: Image is up to date for xpub/xpub:base
✔ ~/dev
$ docker run -it xpub/xpub:base bash
root@d327bb54037d:~# yarn --version
1.3.2
root@d327bb54037d:~# 
```

damnhipster:
The [build and push job](https://gitlab.coko.foundation/yld/xpub-deployment-config-postgres/pipelines) doesn't seem to work, will build and push manually.

damnhipster:
https://gitlab.coko.foundation/xpub/xpub is a separate project which also pushes it's images to the same location as our base image. We should look at changing this.

PR #1193, opened 2018-12-12
Add configurable auth to mailer.js

### LINKED ISSUE #1192: Add Auth to mail config
See discussion: https://elifesciences.slack.com/archives/C6LKMFK5Z/p1544625931021100

Need to add username / password to the mail config in `mailer.js`

diversemix:
#### Background

Allows authentication to SMTP server

#### Any relevant tickets

Closes #1192

Visually inspected - Not tested

### DIFF
--- a/config/non-serializable/mailer.js
+++ b/config/non-serializable/mailer.js
@@ -3,8 +3,5 @@ module.exports = {
     host: `${process.env.MAILER_HOST || 'localhost'}`,
     port: `${process.env.MAILER_PORT || 25}`,
   },
-  auth: {
-    user: `${process.env.MAILER_AUTH_USER || ''}`,
-    pass: `${process.env.MAILER_AUTH_PASSWORD || ''}`,
-  },
+  auth: `${process.env.MAILER_AUTH || '{}'}`,
 }
--- a/config/non-serializable/mailer.js
+++ b/config/non-serializable/mailer.js
@@ -2,6 +2,6 @@ module.exports = {
   transport: {
     host: `${process.env.MAILER_HOST || 'localhost'}`,
     port: `${process.env.MAILER_PORT || 25}`,
+    auth: JSON.parse(`${process.env.MAILER_AUTH || '{}'}`),
   },
-  auth: `${process.env.MAILER_AUTH || '{}'}`,
 }
--- a/config/non-serializable/mailer.js
+++ b/config/non-serializable/mailer.js
@@ -4,7 +4,7 @@ module.exports = {
     port: `${process.env.MAILER_PORT || 25}`,
   },
   auth: {
-    user: `${process.env.MAILER_AUTH_USER || 'user'}`,
-    pass: `${process.env.MAILER_AUTH_PASSWORD || 'password'}`,
+    user: `${process.env.MAILER_AUTH_USER || ''}`,
+    pass: `${process.env.MAILER_AUTH_PASSWORD || ''}`,
   },
 }
--- a/config/non-serializable/mailer.js
+++ b/config/non-serializable/mailer.js
@@ -3,4 +3,8 @@ module.exports = {
     host: `${process.env.MAILER_HOST || 'localhost'}`,
     port: `${process.env.MAILER_PORT || 25}`,
   },
+  auth: {
+    user: `${process.env.MAILER_AUTH_USER || 'user'}`,
+    pass: `${process.env.MAILER_AUTH_PASSWORD || 'password'}`,
+  },
 }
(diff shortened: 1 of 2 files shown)


config/non-serializable/mailer.js
```diff
@@ -3,4 +3,8 @@ module.exports = {
     host: `${process.env.MAILER_HOST || 'localhost'}`,
     port: `${process.env.MAILER_PORT || 25}`,
   },
+  auth: {
+    user: `${process.env.MAILER_AUTH_USER || 'user'}`,
```
giorgiosironi: I think the default here should be empty (`''` or `null` or `undefined`?) as our `localhost` servers don't have authentication being unreachable from the outside world.

diversemix: ok - sounds sensible as it could result in an unreadable error message

diversemix: @giorgiosironi From the docs... https://nodemailer.com/smtp/#authentication

> If authentication data is not present, the connection is considered authenticated from the start. Otherwise you would need to provide the authentication options object.

So that makes me wonder how `''`or `null` would be interpreted / sent to the server :thinking:

giorgiosironi: https://github.com/nodemailer/nodemailer/blob/33b62e2ea6bc9215c99a9bb4bfba94e2fb27ebd0/lib/smtp-transport/index.js#L88-L97 seems to say that the default should be `{}` for the whole object? Needs to be tested.

config/non-serializable/mailer.js
```diff
@@ -3,4 +3,8 @@ module.exports = {
     host: `${process.env.MAILER_HOST || 'localhost'}`,
     port: `${process.env.MAILER_PORT || 25}`,
   },
+  auth: {
+    user: `${process.env.MAILER_AUTH_USER || ''}`,
```
giorgiosironi: also, could this just be `MAILER_USER` and `MAILER_PASS`? Simplifies the pillar configuration to just be a `mailer` dictionary

docs/developing/architecture.md
```diff
@@ -1,5 +1,19 @@
 # Architecture
 
-## Logical Architecture
+## System Architecture
```
giorgiosironi: accidentally committed?

diversemix: Yes! Doh! thanks

config/non-serializable/mailer.js
```diff
@@ -3,4 +3,5 @@ module.exports = {
     host: `${process.env.MAILER_HOST || 'localhost'}`,
     port: `${process.env.MAILER_PORT || 25}`,
   },
+  auth: `${process.env.MAILER_AUTH || '{}'}`,
```
giorgiosironi: if ` is what I think [it is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) it can't be used here because it always produces a string as output. The two possible cases [AFAIK](https://github.com/nodemailer/nodemailer/blob/33b62e2ea6bc9215c99a9bb4bfba94e2fb27ebd0/lib/smtp-transport/index.js#L88-L97) are:
- `{}` (empty dictionary or in any case a falsy value)
- `{ 'user': '...', 'pass': '...' }`

damnhipster:
We're using [`@pubsweet/component-send-email`](https://gitlab.coko.foundation/pubsweet/pubsweet/tree/master/packages/components/SendEmail-server) to send emails. 

The package doesn't look like it has a way to support auth, as it [cherry-pick's the config it wants](https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/components/SendEmail-server/src/SendEmail.js#L18) rather than passing through all of it to nodemailer. This requires a change in Pubsweet.

diversemix:
After testing ... this is not going to work in its current state ... it needs a different approach.

diversemix:
Ok, so after testing we can use:
```
export MAILER_AUTH='{ "user": "t5oksg3zpiv6endr@ethereal.email", "pass": "WgDgJYhXPTUCun3sXj"}'
```
Which is parse-able by `JSON.parse()`

diversemix:
@damnhipster ... I agree we should change this - but probably an issue for the backlog for now.

damnhipster:
How are we/will we be testing this?

giorgiosironi:
I can give you some credentials that you can use for local testing? They are safely kept in https://github.com/elifesciences/builder-private/blob/master/pillar/elife-xpub.sls

diversemix:
@damnhipster I've used a local test script for this - do you think its worth adding to `/scripts` ... I think testing of this is a manual task.

diversemix:
@giorgiosironi ... yes please! But the ethermail seems to work ok for me :)

diversemix:
@damnhipster ... what changes are you requesting, before this gets merged? ... Just the testing?

PR #1195, opened 2018-12-12
[DEMO] 419 react-virtualized implementation

diversemix:
This is an attempt to solve [this](https://github.com/elifesciences/elife-xpub/issues/419#issuecomment-409583197) issue using `react-virtualized`.

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -243,9 +243,11 @@ class SupportingUpload extends React.Component {
                   </UploadControl>
                 )}
               </Flex>
-              <ValidationText>
-                Maximum {maxSupportingFiles} supporting files
-              </ValidationText>
+              {successfullyUploadedFiles.length > 9 && (
+                <ValidationText>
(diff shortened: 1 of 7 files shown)


SemraSh:
### Findings of this attempt 

 After having the issues described [here](https://github.com/elifesciences/elife-xpub/pull/1183) with react-window , @damnhipster and I decided to use the older version which is `react-virtualized` and specifically `WindowScroller` component which is giving the opportunity to scroll anywhere in the window (not only inside the List component) as seen in [this](https://bvaughn.github.io/react-virtualized/#/components/WindowScroller) demo. 

#### Behaviour on the browser
kapture 2018-12-13 at 15 48 25

##### Problems
1. The styles are not easy to overwrite, the `+` icon not changing to `selected` is solvable but to make the component responsive we need to go over the way style is implemented inside the react-virtualized as the whole functionality is effected when I try to suppress the existing styles. I've also tried to add the existing styles and add the new styles so to make it look like the current design and have the functionality but I cannot overwrite some of the styles as apparently they've added inline styling. 

2. The major problem with this is that onScroll functionality is broken. I've tried the same code in codesandbox.io and it was working perfectly fine but in our code base re-rendering the current view onScroll is not working, I wasn't able to debug the issue.

PR #1200, opened 2018-12-13
Use a pinned version of the xpub docker image

diversemix:
#### Background

Allows specifying a pinned version of the xpub base image

#### Any relevant tickets

None - this was an Andon moment

Will be tested in the pipeline


giorgiosironi:
Shall be merged

Issue #1199, opened 2018-12-13
Automate rebuilding of the base image


giorgiosironi:
The base image is time saving as every new version of `elifesciences/elife-xpub` only has to rebuild the application-specific stuff rather than starting from scratch from a `node` image. It originally comes from https://gitlab.coko.foundation/yld/xpub-deployment-config-postgres/blob/master/images/xpub-base/Dockerfile but if all the relevant Dockerfiles are committed with special names into the repository, we can automate the rebuilding process and issue generated tags.

Reference also https://hub.docker.com/r/pubsweet/pubsweet-base/tags/ which only has the `latest` tag and no versioning.

giorgiosironi:
Notice also the `base` image is currently huge (858 MB) because it contains a Yarn cache, can take the opportunity to look for ways to both have quick builds and at the same time shrink its size which also results in faster pushes and pulls.

damnhipster:
We can also shave off 300mb if we use `node:8-alpine` rather than `node:8`

giorgiosironi:
> if we use node:8-alpine 

Though it needs to be tested that dependencies like the headless browser or any NPM package work in that environment

damnhipster:
We'll also have to include git and maybe some other basic packages.

PR #1201, opened 2018-12-13
Added base Dockerfile

damnhipster:
#### Background

The image for `elife-xpub` gets [built from `xpub/xpub`](https://github.com/elifesciences/elife-xpub/blob/develop/Dockerfile#L1) which in turn is currently built from [yld/xpub-deployment-config-postgres](https://gitlab.coko.foundation/yld/xpub-deployment-config-postgres/blob/master/images/xpub-base/Dockerfile)

This change relocated the base Dockerfile to the elife-xpub repository.

#### Any relevant tickets

Unblocks #1199 

Manually built and pushed the image, and tested CI for develop and PRs

### DIFF
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8.9.4
 
 ENV HOME "/home/xpub"
 
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -1,0 +1,19 @@
+FROM node:8
+
+ENV HOME "/home/xpub"
+
+RUN mkdir -p ${HOME}
+
+WORKDIR ${HOME}
+
+# Build yarn offline cache
+
+RUN yarn config set yarn-offline-mirror /npm-packages-offline-cache && \
+  yarn config set prefer-offline true && \
+  yarn config set yarn-offline-mirror-pruning true
+
+RUN git clone https://github.com/elifesciences/elife-xpub.git && \
+    cd elife-xpub && \
+    yarn autoclean --init && yarn && \
+    cd && rm -rf elife-xpub
+
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -1,4 +1,4 @@
-FROM node:8.9.4
+FROM node:8.14.0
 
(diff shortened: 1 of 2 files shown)


Dockerfile-base
```diff
@@ -0,0 +1,19 @@
+FROM node:8
```
giorgiosironi: anyone wants to be more specific, like `:8.14` or `:8-jessie`?

Dockerfile-base
```diff
@@ -0,0 +1,19 @@
...
+
+WORKDIR ${HOME}
+
+# Build yarn offline cache
+
+RUN yarn config set yarn-offline-mirror /npm-packages-offline-cache && \
+  yarn config set prefer-offline true && \
+  yarn config set yarn-offline-mirror-pruning true
+
+RUN git clone https://github.com/elifesciences/elife-xpub.git && \
```
giorgiosironi: right, this is not using `xpub-collabra`

damnhipster: Yes, it used to, but it didn't make sense since the specific tag we used was being built by us and being used by us.

Given the location of it, should we change it to clone `xpub-collabra`?

damnhipster:
Since we're hosting this within the repo, what do you guys think of using the commit SHA as the tag for the base docker image? I'm about to push a second image up for the day, and it's something we would want to track.

giorgiosironi:
For frequent changes the commit sha is useful to produce a unique value that is trackable; the disadvantage is that it is not very readable. This [ADR](https://github.com/elifesciences/elife-base-images/blob/master/adr/0001-commit-tags.md) for `elife-base-images` came to the same conclusion, use the commit sha.

damnhipster:
@giorgiosironi Given that `xpub-collabra` use the same convention for their tags, we should hold off changing this and do it as part of #1202. In the meantime, does anyone have any issues, if I overwrite today's image with this new updated one?

giorgiosironi:
> does anyone have any issues, if I overwrite today's image with this new updated one?

Anyone that has already pulled the image, including Jenkins nodes, won't see the new one; hence named it `20181213-1143` or similar

PR #1213, opened 2018-12-17
Fix autosave with manuscript refresh

### LINKED ISSUE #1162: Autosave feature broken

diversemix:
#### Background

- Add refresh() to manuscript to prevent saving stale data
- Adds data integrity checks to manuscript.save()

#### Any relevant tickets

Closes #1162 

Added unit tests for saving - no added browser test or end2end tests as this code will be exercised with the existing tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ test/screenshots/
 .vscode
 .node-gyp
 .idea
+migrations-*
--- a/package.json
+++ b/package.json
@@ -69,8 +69,8 @@
     ]
   },
   "resolutions": {
-    "styled-system": "2.3.6",
-    "@pubsweet/ui-toolkit": "1.2.0"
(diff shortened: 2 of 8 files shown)


giorgiosironi:
If you merge with an instability, the tests that fail would need to be quarantined and allowed to fail so that other changes can be delivered without being blocked by the instability. So either:

- fix the instability before merging this change
- comment out the throwing part if possible, to merge this change (transforming it into a warning log or similar non-fatal behavior)

diversemix:
I'm going to go with the second... as I smell a rabbit hole!!!

Issue #1228, opened 2018-12-20
test:browser hanging


giorgiosironi:
e.g. https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Felife-xpub/detail/PR-1181/6/pipeline hangs apparently forever in the `test:browser` target until the build is killed after a 2 hours timeout.

Situation seems more or less normal:
```
ubuntu@ip-10-0-2-66:~$ docker logs elife-xpub_app_test_browser

> xpub-elife@1.0.0 test:browser /home/xpub
> NODE_CONFIG_ENV=test NODE_ENV=production testcafe "${BROWSER:-chrome}" 'test/**/*.browser.js' "--screenshots" "/tmp/screenshots" "--screenshots-on-fails"

clean-webpack-plugin: /home/xpub/_build/assets has been removed.
 Running tests in:
 - HeadlessChrome 71.0.3578 / Linux 0.0.0

 Submission
Type "TeamMemberMeta" is missing a "__resolveType" resolver. Pass false into "resolverValidationOptions.requireResolversForResolveType" to disable this warning.
Type "Object" is missing a "__resolveType" resolver. Pass false into "resolverValidationOptions.requireResolversForResolveType" to disable this warning.

root@f9bd68cefa51:~# ps faxww
  PID TTY      STAT   TIME COMMAND
  450 pts/0    Ss     0:00 bash
  481 pts/0    R+     0:00  \_ ps faxww
    1 ?        Ssl    0:00 npm
   14 ?        S      0:00 sh -c NODE_CONFIG_ENV=test NODE_ENV=production testcafe "${BROWSER:-chrome}" 'test/**/*.browser.js' "--screenshots" "/tmp/screenshots" "--screenshots-on-fails"
   15 ?        Sl     0:00  \_ node /home/xpub/node_modules/.bin/testcafe chrome:headless --no-sandbox test/**/*.browser.js --screenshots /tmp/screenshots --screenshots-on-fails
   21 ?        Sl     3:16      \_ /usr/local/bin/node /home/xpub/node_modules/testcafe/lib/cli chrome:headless --no-sandbox test/**/*.browser.js --screenshots /tmp/screenshots --screenshots-on-fails
[...]

PR #1233, opened 2018-12-21
TEST-TEST-TEST put in a lock-free pattern

diversemix:
Its nearly Christmas! :package:


diversemix:
Closing out these test - as it is looking to be an infrastructure issue see #1228

Issue #1243, opened 2019-01-04
Add manuscript state to AuditLog


damnhipster:
When a manuscript (in the `base-model`) has a State change - this should get logged into the database `audit_log` table. This includes manuscript create and delete.

- [x] Add a "setter" for `status` onto the `Manuscript` entity in `xpub-model`
- [x] Change everywhere to use the property setter and not directly change the variable.
- [x] Change the setter for `status` to create an `AuditLog`'s log entry with appropriate params.

damnhipster:
Setter will get called when the instance property gets changed. [See the docs.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) 

```
var language = {
  set current(name) {
    this.log.push(name);
  },
  log: []
}

language.current = 'EN';
language.current = 'FA';

console.log(language.log);
// expected output: Array ["EN", "FA"]
```

damnhipster:
> Setter will get called when the instance property gets changed. [See the docs.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set)

Will be using an `updateStatus` function instead, as the `status` keyword is used as a property, which means we can't use `set status` since it's the same keyword.

Raised #1249 to address the issue.

hem-brahmbhatt:
Raised #1391 to store the following details to the AuditLog: *user performing the action on the object*

PR #1256, opened 2019-01-08
1243 manuscript state audit

hem-brahmbhatt:
#### Background

Updates the audit log when manuscript status is updated.

#### Any relevant tickets

#1243 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


server/xpub-model/entities/manuscript/index.js
```diff
@@ -88,6 +88,14 @@ class Manuscript extends BaseModel {
           to: 'team.objectId',
         },
       },
+      audits: {
```
diversemix: Why do we want to do this? Not sure there is a requirement yet that this is part of the Manuscript model, just that they get logged.

server/meca/routes.test.js
```diff
@@ -14,13 +14,17 @@ const makeApp = () => {
   return supertest(app)
 }
 
-describe('MECA HTTP callback handler', () => {
-  const userId = '9A3E046F-8CC9-4880-81E6-309CF0B726F4'
+describe.only('MECA HTTP callback handler', () => {
```
diversemix: Is the only left in by mistake?

server/xpub-model/entities/manuscript/index.js
```diff
@@ -201,16 +209,40 @@ class Manuscript extends BaseModel {
     return this
   }
 
-  // atomically update the manuscript status
+  updateStatus(status) {
+    const audit = {
+      userId: this.createdBy,
+      action: 'UPDATED_STATUS',
```
diversemix: We need to talk about actions - I think it should be an enum of 'CREATED', 'UPDATED', 'DELETED'

server/xpub-model/entities/manuscript/index.js
```diff
@@ -201,16 +209,40 @@ class Manuscript extends BaseModel {
     return this
   }
 
-  // atomically update the manuscript status
+  updateStatus(status) {
+    const audit = {
+      userId: this.createdBy,
```
diversemix: `userId` is incorret - this should be set to the user who initiated the request - not the user who created the manuscript.

server/xpub-model/entities/manuscript/index.js
```diff
@@ -201,16 +209,40 @@ class Manuscript extends BaseModel {
...
+      action: 'UPDATED_STATUS',
+      objectType: 'manuscript',
```
diversemix: Can we have a dotted notiation - like `manuscript.status` ?

server/xpub-model/entities/manuscript/index.js
diversemix: We should also enforce this in the database?

diversemix (changes requested):
Pairing now on changes

diversemix (approved):

diversemix:
Was just testing this and get:
```
error: insert into "audit_log" ("action", "created", "id", "object_id", "object_type", "updated", "value") values ($1, $2, $3, $4, $5, $6, $7) returning "id" - column "value" of relation "audit_log" does not exist message=insert into "audit_log" ("action", "created", "id", "object_id", "object_type", "updated", "value") values ($1, $2, $3, $4, $5, $6, $7) returning "id" - column "value" of relation "audit_log" does not exist, locations=[line=2, column=3], path=[submitManuscript]
error: column "value" of relation "audit_log" does not exist
```
rebuilding my database and starting again :D

diversemix:
All good :+1:

PR #1263, opened 2019-01-08
1163 supporting files tests

### LINKED ISSUE #1163: Tests for Supporting files
Add browser tests for the supporting file uploads

Cooryd:
#### Background

- Adds a browser test for the upload and removal of supporting files. 
- Extends the unit test for removing supporting files to test the files are removed from the manuscript.
- Adds a test for the `uploadSupportingFile` mutation.

#### Any relevant tickets

closes #1163

This adds unit and browser tests

#### UI Changes

None


server/xpub-server/entities/manuscript/resolvers/uploadSupportingFile.test.js
```diff
@@ -0,0 +1,52 @@
...
+      mimetype: 'application/pdf',
+    }
+    expect(manuscript.files).toHaveLength(0)
+    await Mutation.uploadSupportingFile(
+      {},
+      { file: new Promise(resolve => resolve(supportingFile)), id },
+      { user: profileId },
+    )
+    manuscript = await Manuscript.find(id, userId)
+    expect(manuscript.files).toHaveLength(1)
```
diversemix: Also assert its a supporting file and not a manuscript type - and any other attributes?

diversemix:
Browser tests succeed locally :+1:

damnhipster:
- we have a test for removing files within the resolvers where it's actually testing the controller. We should move the test to the controller package, or test it at the lever of the resolver.
- the browser test doesn't have any assertions to make sure we're deleting the files downstream, from the database and from S3. Can we add assertions for those?

damnhipster:
> the browser test doesn't have any assertions to make sure we're deleting the files downstream, from the database and from S3. Can we add assertions for those?

On this point, it doesn't sound correct for browser tests to assert against the server, what do you think?

Cooryd:
I agree, a check like that should probably be put into a unit test. I'm unsure what the best approach to testing the contents of s3 would be though, should we trust the api and just mock it out and assert it's being called or is there a way to query whats in there from a unit test?

PR #1269, opened 2019-01-09
(deprecated) 1220 Loading state for PeoplePicker

### LINKED ISSUE #1220: Create loading state for People Picker
As a short term fix to the people picker performance issue - put up a loady spinny thing to indicate to the user that it might take some while.

Ideally we just show a centred spinner with “Loading”. If we have to show other elements of the modal we should center it in available space
The numbers show in order of preference 1 being most preferred

image

SemraSh:
#### Background
Creates a loading state for People picker. User used to be hanged on the previous page while the response and the rendering is in progress, this change is showing them a loading spinner while they wait for those processes to be done.

What does this PR do?
Creates a new Loading component which has a spinner.
The aim of the ticket was adding a Loading spinner to PeoplePicker, that change is made mainly in PersonPodGrid component because that is the element in the PeoplePicker that holds the elements designer ( @chugginselifesciences ) wants to replace with a loading spinner.

This PR is also replacing all the `<div> Loading... <div/>`s with the new Loading component. 

#### Any relevant tickets
Closes #1220 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
+++ b/app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
@@ -77,8 +77,8 @@ describe('PeoplePicker', () => {
     expect(render).toHaveBeenCalled()
   })
 
-  it('renders as many person pods as people', async () => {
-    const wrapper = await makeWrapper()
+  it('renders as many person pods as people', () => {
+    const wrapper = makeWrapper()
     const buttons = wrapper.find('PersonPod')
 
     expect(buttons).toHaveLength(people.length)
@@ -137,7 +137,7 @@ describe('PeoplePicker', () => {
     expectSelectionLength(wrapper, 0)
   })
 
-  describe('integration with Search Box', async () => {
+  describe('integration with Search Box', () => {
     const searchWrapper = mount(
       <ThemeProvider theme={theme}>
         <PeoplePickerLogic
--- a/app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
+++ b/app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
@@ -127,7 +127,6 @@ describe('PeoplePicker', () => {
   it('clicking on selected item icon removes it', () => {
     const wrapper = makeWrapper()
     expectSelectionLength(wrapper, 0)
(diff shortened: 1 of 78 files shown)


app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
```diff
@@ -132,7 +137,7 @@ describe('PeoplePicker', () => {
     expectSelectionLength(wrapper, 0)
   })
 
-  describe('integration with Search Box', () => {
+  describe('integration with Search Box', async () => {
```
damnhipster: Why is `async` needed here?

Cooryd: Left in by mistake after I removed an await inside this test. This can be removed @SemraSh

hdrury1:
@chugginselifesciences can you take a look at this in the review environment and confirm if you're happy?

app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
```diff
@@ -71,8 +77,8 @@ describe('PeoplePicker', () => {
     expect(render).toHaveBeenCalled()
   })
 
-  it('renders as many person pods as people', () => {
-    const wrapper = makeWrapper()
+  it('renders as many person pods as people', async () => {
+    const wrapper = await makeWrapper()
```
Cooryd: Looking back through this I'm unsure we need this function to be async and await the makeWrapper either.

app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
SemraSh: @damnhipster thank you for this!

app/components/ui/PeoplePicker/PeoplePickerLogic.test.js
SemraSh: 👍

chugginselifesciences:
@hdrury1 Looks good to me. It seems to be loading quite quickly too which is great. Thanks all.

PR #1272, opened 2019-01-09
1220 Loading state for PeoplePicker

### LINKED ISSUE #1220: Create loading state for People Picker

diversemix:
#### Background
Creates a loading state for People picker. User used to be hanged on the previous page while the response and the rendering is in progress, this change is showing them a loading spinner while they wait for those processes to be done.

What does this PR do?
Creates a new Loading component which has a spinner.
The aim of the ticket was adding a Loading spinner to PeoplePicker, that change is made mainly in PersonPodGrid component because that is the element in the PeoplePicker that holds the elements designer ( @chugginselifesciences ) wants to replace with a loading spinner.

This PR is also replacing all the `<div> Loading... <div/>`s with the new Loading component. 

#### Any relevant tickets
Closes #1220 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


Cooryd (approved):
All looks good, @SemraSh could you just look through to make sure theres nothing missing from the recreation of the branch? If not I think it's okay to merge.

PR #1251, opened 2019-01-07
[wip] Re-enable test:browser

giorgiosironi:

PR #1273, opened 2019-01-09
Enable browser tests

hem-brahmbhatt:
#### Background

Runs browser tests within it's own stage in Jenkins. This stops it from being run in parallel alongside linting and unit tests, which allows the app server to start up more quickly for the browser tests, which in turn stops them from hanging.

These changes have been [cherry-picked](https://git-scm.com/docs/git-cherry-pick) from #1251 

#### Any relevant tickets

Related to #1228 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -71,14 +71,12 @@ elifePipeline {
         }
 
         stage 'Browser Tests', {
-            withCommitStatus({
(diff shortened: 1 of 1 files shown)


diversemix:
This is still just a temporary fix. We are all agreed it is Testcafe's issue. Thus ends our ring fenced time. 

@damnhipster if you get round to raising an issue for Testcafe let's create a wrapper issue here in backlog.


/cc @hdrury1

Issue #1164, opened 2018-12-04
MECA support for supporting files


diversemix:
Change the `article.xml` and the MECA package to contain supporting files.

- [x] Update the manifest.xml to contain references to the files #1274
- [x] Update the zip to contain the files #1275
- [ ] Store the mimetype when uploading

PR #1274, opened 2019-01-10
Update the manifest.xml to contain references to the supplementary files

diversemix:
#### Background

What does this PR do?

#### Any relevant tickets

Step 1 of #1164 

- [Unit / Integration tests have been updated.
- Browser tests - not needed
- End2end tests - out of scope - being done on separate ticket

### DIFF
--- a/server/meca/file-generators/manifest.js
+++ b/server/meca/file-generators/manifest.js
@@ -1,7 +1,40 @@
 const fs = require('fs-extra')
 
-async function generateManifest() {
-  return fs.readFile(`${__dirname}/manifest.xml`, 'utf8')
+function supplementaryXml(files) {
+  const keyList = ['id', 'mimeType', 'filename']
+  const fileList = files.filter(
+    fileObject => fileObject.type === 'SUPPORTING_FILE',
+  )
+
+  let supplementaryFileXml = ''
+  const template = fs.readFileSync(
+    `${__dirname}/manifest-supplementary-file.xml`,
+    'utf8',
+  )
+
+  fileList.forEach(file => {
+    supplementaryFileXml += keyList.reduce(
+      (xml, key) => xml.replace(new RegExp(`{${key}}`, 'g'), file[key]),
+      template,
+    )
+  })
+
+  return supplementaryFileXml
+}
+
+function manifestXml(suppFilesXml) {
+  const template = fs.readFileSync(`${__dirname}/manifest.xml`, 'utf8')
+  const xml = template.replace('{supplementaryFiles}', suppFilesXml)
+
+  return xml
+}
+
+function generateManifest(files) {
+  if (!Array.isArray(files)) {
+    throw TypeError(`Expecting array: ${typeof files}`)
+  }
+
+  return manifestXml(supplementaryXml(files))
 }
 
 module.exports = generateManifest
--- a/server/meca/file-generators/manifest.test.js
+++ b/server/meca/file-generators/manifest.test.js
@@ -1,10 +1,83 @@
-const manifest = require('./manifest')
+const generateManifest = require('./manifest')
+const sampleManuscript = require('./article.test.data')
 
 describe('Manifest XML generator', () => {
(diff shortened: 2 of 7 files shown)


server/meca/file-generators/manifest.test.js
```diff
@@ -1,10 +1,83 @@
...
+      '<item id="00000001-6c48-4747-851c-ef806e8486b2" type="supplemental"',
+    )
+
+    // has supplementary file 2
+    expect(xml).toContain(
+      '<item id="00000002-6c48-4747-851c-ef806e8486b2" type="supplemental"',
+    )
   })
 })
+
```
diversemix: Ooops left in this comment!

server/meca/file-generators/manifest.js
```diff
@@ -1,7 +1,40 @@
...
-async function generateManifest() {
-  return fs.readFile(`${__dirname}/manifest.xml`, 'utf8')
+function supplementaryXml(files) {
+  const keyList = ['id', 'mimeType', 'filename']
+  const fileList = files.filter(
+    fileObject => fileObject.type === 'SUPPORTING_FILE',
+  )
+
+  let supplementaryFileXml = ''
+  const template = fs.readFileSync(
```
damnhipster: Given that this is `sync`, did we see any delays when submitting?

server/meca/file-generators/manifest.test.js
```diff
@@ -1,10 +1,43 @@
-const manifest = require('./manifest')
+const generateManifest = require('./manifest')
+const sampleManuscript = require('./article.test.data')
 
 describe('Manifest XML generator', () => {
-  it('returns meaningful content', async () => {
-    const xml = await manifest()
+  it('returns required items', async () => {
+    const xml = await generateManifest(sampleManuscript.files)
```
damnhipster: This is no longer async.

diversemix: The actual export function is async... 
```

async function mecaExport(manuscript, content, clientIp) {
  logger.info(`Starting MECA export`, { manuscriptId: manuscript.id })
  const archive = await generate(manuscript, content, clientIp)
  await upload(archive, manuscript.id)
}
```

server/meca/file-generators/manifest.xml
```diff
@@ -32,4 +32,5 @@
         <instance media-type="text/html" href="cover_letter.html"/>
     </item>
 
+    {supplementaryFiles}
```
damnhipster: This is now a template rather than an xml file, we should indicate that, either by putting this within a templates folder, or using a templating pattern. We can also follow the same pattern as the [`article.js` file](https://github.com/elifesciences/elife-xpub/blob/develop/server/meca/file-generators/article.js)

server/meca/file-generators/manifest.test.js
damnhipster: That isn't the function being tested here, we're testing [`generateManuscript`](https://github.com/elifesciences/elife-xpub/blob/db80c2008a884b40bd17e4b7396d84513ceacc60/server/meca/file-generators/manifest.js#L40)

server/meca/file-generators/manifest.js
diversemix: The actual export function is async... 
```

async function mecaExport(manuscript, content, clientIp) {
  logger.info(`Starting MECA export`, { manuscriptId: manuscript.id })
  const archive = await generate(manuscript, content, clientIp)
  await upload(archive, manuscript.id)
}
```

server/meca/file-generators/manifest.xml
diversemix: Agreed - something for the future

server/meca/file-generators/manifest.test.js
diversemix: This will be fixed in the next PR.

diversemix: see #1275

PR #1275, opened 2019-01-10
Add Supporting files to Zip

diversemix:
#### Background

Adds the supporting files to the zip package.

#### Any relevant tickets

Step 2 of #1164

- Unit / Integration tests - changed
- Browser tests not neccessary
- E2E tests out of scope

### DIFF
--- a/server/meca/Readme.md
+++ b/server/meca/Readme.md
@@ -6,9 +6,12 @@ xPub component to generate and export a Manuscript Exchange Common Approach (MEC
 
     const {mecaExport} = require('@elifesciences/xpub-meca-export')
 
-    mecaExport(manuscriptObject, manuscriptContent, clientIp)
+    mecaExport(manuscriptObject, getContentFunction, clientIp)
         .then(() => console.log('Successfully uploaded archive to SFTP'))
         .catch(err => console.error(err))
+
+The `getContentFunction` has the signiture of one argument, the `File` object which
+will return the file contents of that object.
 
 ## Export callback endpoint
 
--- a/server/meca/export.js
+++ b/server/meca/export.js
@@ -9,24 +9,28 @@ const archiveGenerator = require('./file-generators/archive')
 
 const upload = require('./services/upload')
 
-// const files = new SupportingFiles(dummyStorage, manuscript.id, userId)
-
 async function generate(manuscript, getContent, clientIp) {
   let content = ''
+  const supportingFiles = {}
+
   manuscript.files.forEach(file => {
     if (file.type === 'MANUSCRIPT_SOURCE') {
       content = getContent(file)
+    } else if (file.type === 'SUPPORTING_FILE') {
+      supportingFiles[file.filename] = getContent(file)
     }
   })
 
-  return archiveGenerator({
+  const manditoryFiles = {
     'article.xml': articleGenerator(manuscript),
     'cover_letter.html': coverLetterGenerator(manuscript),
     'disclosure.pdf': disclosureGenerator(manuscript, clientIp),
     'manifest.xml': manifestGenerator(manuscript.files),
     'manuscript.pdf': content,
     'transfer.xml': transferGenerator(''), // auth code not currently used
-  })
+  }
+
+  return archiveGenerator({ ...manditoryFiles, ...supportingFiles })
 }
 
 async function mecaExport(manuscript, getContent, clientIp) {
--- a/server/meca/export.js
+++ b/server/meca/export.js
@@ -10,27 +10,34 @@ const archiveGenerator = require('./file-generators/archive')
 const upload = require('./services/upload')
 
 async function generate(manuscript, getContent, clientIp) {
-  let content = ''
-  const supportingFiles = {}
+  const manditoryFiles = [
+    { name: 'article.xml', content: articleGenerator(manuscript) },
+    { name: 'cover_letter.html', content: coverLetterGenerator(manuscript) },
+    {
+      name: 'disclosure.pdf',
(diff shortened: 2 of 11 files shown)


server/meca/Readme.md
```diff
@@ -6,10 +6,13 @@ xPub component to generate and export a Manuscript Exchange Common Approach (MEC
...
     const {mecaExport} = require('@elifesciences/xpub-meca-export')
 
-    mecaExport(manuscriptObject, manuscriptContent, clientIp)
+    mecaExport(manuscriptObject, getContentFunction, clientIp)
         .then(() => console.log('Successfully uploaded archive to SFTP'))
         .catch(err => console.error(err))
 
+The `getContentFunction` has the signiture of one argument, the `File` object which
+will return the file contents of that object.
+
```
damnhipster: I'm not a huge fan of this interface, as it feels like we're shoehorning in some functionality so the export works.

We could change MecaExport to a class, and do the following:
```
const meca = new Meca({
  getContent: S3Storage.getContent
})
meca.export(manuscript, clientIp)
```

It's also easier to add functionality to a class in the future if we need to fetch content from other places, but more than anything, it's a lot easier to read IMO

server/meca/export.js
```diff
@@ -9,20 +9,33 @@ const archiveGenerator = require('./file-generators/archive')
...
+  const supportingFiles = {}
+
+  manuscript.files.forEach(file => {
+    if (file.type === 'MANUSCRIPT_SOURCE') {
+      content = getContent(file)
+    } else if (file.type === 'SUPPORTING_FILE') {
+      supportingFiles[file.filename] = getContent(file)
+    }
+  })
+
```
damnhipster: May be worth erroring if `content` is null.

server/meca/export.js
```diff
@@ -9,20 +9,33 @@ const archiveGenerator = require('./file-generators/archive')
...
     'article.xml': articleGenerator(manuscript),
     'cover_letter.html': coverLetterGenerator(manuscript),
     'disclosure.pdf': disclosureGenerator(manuscript, clientIp),
     'manifest.xml': manifestGenerator(manuscript.files),
     'manuscript.pdf': content,
     'transfer.xml': transferGenerator(''), // auth code not currently used
-  })
+  }
+
+  return archiveGenerator({ ...manditoryFiles, ...supportingFiles })
```
damnhipster: Depending on the name of the supporting file, it could overwrite the manditory file.

e.g. user uploads a supporting file called `manifest.xml`

We may want to throw an error if this happens.

server/xpub-server/entities/manuscript/resolvers/submitManuscript.test.js
```diff
@@ -103,8 +106,9 @@ describe('Manuscripts', () => {
         actualContent,
         actualIp,
       ] = mecaExport.mock.calls[0]
+
       expect(actualManuscript.id).toBe(id)
-      expect(actualContent).toBe('A real PDF')
+      expect(actualContent).toHaveBeenCalled()
```
damnhipster: No point having this assertion, as we're testing the mock.

`expect(mecaExport).toHaveBeenCalled()` is enough.

server/meca/Readme.md
diversemix: This would be my preference too ... just trying to make the least risky change at this stage of the project.

server/meca/export.js
diversemix: Think this is the responsibility of the storage class ... pretty certain you'd get a rejected Promise if the `S3` storage fails.

server/meca/export.js
diversemix: Fair point

server/xpub-server/entities/manuscript/resolvers/submitManuscript.test.js
diversemix: We also want to test that there has been an attempt to retrieve the content

diversemix: ... this is now done elsewhere so removed.

diversemix:
@damnhipster ... can you re-review pls :+1:

server/meca/export.js
damnhipster: I think it's worth doing for `content` though (the manuscript source), as I think the export shouldn't be allowed to happen if it doesn't exist.

An error message like `'Failed to export MECA package. No manuscript source was provided'` makes it clear what the problem is.

server/meca/Readme.md
damnhipster: Happy for this to go in a separate issue?

server/meca/export.js
diversemix: Sounds good...
@hdrury1 ... should we also fail if a supporting file cannot be put in the zip file? ... or should we carry on regardless?

server/meca/Readme.md
diversemix: Yeah ... I'm seeing this as part of the wider architecture issue. Although the MECA component is planned on being adopted by `pubsweet` ... that is another reason I've chosen to keep it as it is at the moment. But I'm happy this goes in as an issue on its own.

server/meca/export.js
diversemix: Talking with @Cooryd ... we are thinking the best thing to do is to check against file size (storing it in the db) ... as this would get round the zero size file issue. At the moment my feeling is that we should not do this check as we risk a false error rather than a real one.

/cc @hdrury1 ... do you think this warrants a separate issue?

PR #1277, opened 2019-01-10
1122 fix fonts

### LINKED ISSUE #1122: Fix fonts across codebase - differentiate between reading/writing
Right now many areas of the application display the wrong font - serif vs sans. This is because we're using our `Paragraph` (or `SmallParagraph`) atom everwhere, which uses the `fontReading` variable by default - Noto Serif.

Just a couple of examples of this - the `PersonPod` & the disclosure page of the submission wizard.

History:
- There is no Paragraph atom in pubsweet
- Our Paragraph atom was created when we started work on the static pages epic. All static pages need `fontReading`
- We started using `Paragraph` everywhere else in our codebase, without adjusting the font, so there are many situations in which we use serif when we should use sans
- We do need a `Paragraph` atom, rather than using the browser's default `p` tag, because p tags come with margin attached by default & the eLife styleguide calls for different margins & specific line heights to be attached

Other relevant parts of the existing codebase:
- When the size changes, we have entirely different components: `Paragraph` & `SmallParagraph`
- When the colour needs to change, we achieve this through a prop: if `secondary` is passed in, then it uses `colorTextSecondary`, otherwise (by default) it uses `colorText`

We could either:
1. Export 2 types of Paragraph from the same file, like we do with the FormHeadings - i.e. ReadingParagraph and WritingParagraph
2. Choose a default font for `Paragraph` (e.g. `fontWriting`). Then create a sub-component within the same file e.g. `Paragraph.Reading` for when we want the `fontReading` variable (like we do with `PeoplePicker` & `PeoplePicker.Buttons`)
3. Pass a font prop into `Paragraph`. Required this prop, so that you have to specify reading or writing whenever you use the `Paragraph` component

The downside with option 3 is that whenever another atom/molecule uses `Paragraph` under the hood (e.g. `eJPLink`), then we need to pass the same font prop down through there as well. This could easily get out of hand, having to pass props all the way down through many parents.

So I would suggest either of the first 2 options - i.e. a different component.

## Subtasks

TBD

## Assumptions and notes
TBD

SemraSh:
#### Background

What does this PR do?
Fixes fonts across codebase differentiated between reading and writing.

#### Any relevant tickets
Closes #1122 

👀 

#### UI Changes


chugginselifesciences:
@SemraSh @hdrury1 my guess is that these remaining ones are down to `SmallParagraph` still using `Noto Serif` based on the size. 

Longer term I think it would make sense to move to something like `Body` and `SmallBody` for `Noto Sans` and keep `Paragraph` and `SmallParagraph` as `Noto Serif` for large blocks of text. 

If work has gone into using `FontReading` and `FontWriting` for the distinction then perhaps it's not worth it. As long as we can avoid Serif popping up again in strange places then I'm happy however we do it.

hdrury1:
Looks good to me, but I'd like @chugginselifesciences to take a second glance before we merge.

PR #1279, opened 2019-01-11
1010 reviewer suggestion

Cooryd:
#### Background

- Removes the requirement to have three suggested reviewers making all optional (but maintaining the limit of six)

- Reduces the initial number of suggested reviewer inputs displayed from three to one

- Extends the EditorPage unit test to test the functionality of adding new inputs when the user fills in blank suggestedReviewer rows and remove trailing empty detail rows.

- Corrects some typos in the props passed to the People picker modal components.

#### Any relevant tickets

closes #1010
closes #1024

#### How has this been tested?
**Reviewers** ensure that these test requirements are also reviewed.

Please indicate the need for this change to be tested. If not please justify, if so then at what levels. 

- [x] Unit / Integration tests
- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

- [ ] Has been reviewed against Designs
- [ ] Necessary updates to styleguide

PR #1285, opened 2019-01-11
Fix broken validation on submit modal

Cooryd:
#### Background

#1279 removed the validation limits on the suggested reviewers form field however there were still references to the minimum of 3 in the `Joi` validation schema so the app failed when trying to submit the manuscript. This fix allows for a suggested reviewer count of 0.

#### Any relevant tickets

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


hem-brahmbhatt (approved):

server/xpub-model/entities/manuscript/index.js
```diff
@@ -293,12 +293,9 @@ class Manuscript extends BaseModel {
   // would mean applying the input to the manuscript first
   static removeOptionalBlankReviewers(input) {
     const itemIsBlank = item => item.name + item.email === ''
-
     const filteredReviewers = input.suggestedReviewers.filter(
-      (item, index) =>
-        index < Manuscript.MIN_SUGGESTED_REVIEWERS || !itemIsBlank(item),
+      item => !itemIsBlank(item),
```
diversemix, line 294: Not convinced this logic is still correct ... if we have a `MIN_SUGGESTED_REVIEWS` as `2` then this will not leave two empty slots to be filled. Rather it would remove all the ones that are non-blank.

hem-brahmbhatt: I'm not sure what you mean. I've tested it out on the review environment, and it looks fine.

diversemix: So have I ... looks good ... but if you were to change `MIN_SUGGESTED_REVIEWS = 2` it wouldn't work.

diversemix (approved):

Issue #1290, opened 2019-01-14
Add Tests to xpub-controller


diversemix:
Primarily to cover testing of sending email.... this has not been done as part of #1289 as there is a wider question to answer in terms of the structure of the code. As such this will take time and so split into a separate task.

/cc @hdrury1

diversemix:
Closing in preference to #1587

PR #1291, opened 2019-01-14
Fix missing braces in setting originalEmails

### LINKED ISSUE #1289: Spotting bug which will result in multiple emails being sent
Fix the bug...

In `server/xpub-controller/manuscript.js` the value of `originalEmails` is not being set correctly.

diversemix:
#### Background

Stop multiple emails being sent.

#### Any relevant tickets

Closes #1289 

see #1290

### DIFF
--- a/server/xpub-controller/manuscript.js
+++ b/server/xpub-controller/manuscript.js
@@ -18,9 +18,12 @@ class Manuscript {
       )
     }
 
-    const originalEmails = manuscript.getAuthor()
+    const originalEmails = (manuscript.getAuthor()
       ? manuscript.getAuthor()
-      : [].map(author => author.alias.email).join(',')
+      : []
+    )
+      .map(author => author.alias.email)
+      .join(',')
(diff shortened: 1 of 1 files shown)


jyenes (approved):
pretty straight forward

server/xpub-controller/manuscript.js
```diff
@@ -18,9 +18,12 @@ class Manuscript {
       )
     }
 
-    const originalEmails = manuscript.getAuthor()
+    const originalEmails = (manuscript.getAuthor()
       ? manuscript.getAuthor()
-      : [].map(author => author.alias.email).join(',')
+      : []
+    )
```
hem-brahmbhatt, line 24:
```suggestion
  const originalEmails =  manuscript.getAuthor() || []
```

diversemix: Sorry that was prettify that refactored all that :/

Issue #1198, opened 2018-12-12
Investigate: Title extraction working intermitently


hdrury1:
**Describe the bug**
I am only getting the title extracted around 50% of the time I upload a manuscript. Often, extraction will work if I navigate back to the Files step and re-upload the same manuscript file.

**Steps To Reproduce**
Difficult to reproduce, as the failure is intermitent. Keeping this bug ticket as a placeholder- will test extensively and copy terminal logs.

**Expected behavior**
Title extraction should happen successfully in the majority of cases.

hdrury1:
Uploaded the following file:

[elife-40583-v2.pdf](https://github.com/elifesciences/elife-xpub/files/2685648/elife-40583-v2.pdf)

The application gave me a success message, but navigating to the next page the title had not extracted. I navigated back and re-uploaded the same file. This time the title appeared on the next page of the wizard as expected.

hdrury1:
Uploaded the following file with the same results- title only extracting on second upload attempt:

[elife-39723-v2.pdf](https://github.com/elifesciences/elife-xpub/files/2744872/elife-39723-v2.pdf)

And this file wouldn't extract at all:

[elife-38852-v2.pdf](https://github.com/elifesciences/elife-xpub/files/2744871/elife-38852-v2.pdf)

diversemix:
Adding logging into the app to try and figure out what is going on. Couldn't get things to fail locally - and we probably need some of this logging in when we go live to assess how sciencebeam is doing. see #1286

/cc @hdrury1

diversemix:
We can reproduce the problem locally by doing a `docker restart` on the sciencebeam containers, and then requesting an extraction....

```
$ time python tester.py 
[...]

PR #1293, opened 2019-01-14
Double the timeout to see if sciencebeam starts working on staging

diversemix:
#### Background

Increases sciencebeam timeout to see if this solves the failures we see

#### Any relevant tickets

Investigates #1198 

N / A


jyenes (approved):
LGTM

PR #1301, opened 2019-01-16
1001 add leadership type to Senior Editor view of People Picker

### LINKED ISSUE #1001: Deputy editors not displayed in Senior Editors views of People Picker
In the `PeoplePicker` when showing senior editors it also needs to show deputy editors. 

The people picker should show manuscript roles which means that users with global roles of 
Deputy Editors and Senior Editors should be displayed.

# Relevant Code
- `server/xpub-server/entities/user/resolvers.js:13` only ever makes one call to the api - this needs to make two calls (see above)
- `server/xpub-model/entities/user/helpers/elife-api.js`

diversemix:
#### Background

Shows all relevant users in People Picker's senior editor view

#### Any relevant tickets

Closes #1001

Added tests - tested locally


app/components/pages/SubmissionWizard/steps/Editors/index.js
```diff
@@ -19,7 +19,10 @@ const EditorsPageContainer = ({ ...props }) => (
   <Query query={EDITOR_LIST_QUERY} variables={{ role: 'reviewing-editor' }}>
     {({ data: { editors: reviewingEditors = [] } }) => (
       // TODO handle errors
-      <Query query={EDITOR_LIST_QUERY} variables={{ role: 'senior-editor' }}>
+      <Query
+        query={EDITOR_LIST_QUERY}
+        variables={{ role: 'senior-editor,leadership' }}
```
diversemix, line 24: Had a discussion with @Cooryd about the principle of "exposing" the eLife api fields here... we both agreed that because the API is public its not a worry - but acknowledge normal good practise would be to de-couple this,

server/xpub-model/entities/user/helpers/elife-api.js
```diff
@@ -46,16 +46,19 @@ const convertPerson = apiPerson => {
...
   let response
   let page = 1
   do {
-    response = await request('people', {
-      order: 'asc',
-      page,
-      'per-page': 100,
-      type: role,
-    })
+    let query = `order=asc&page=${page}&per-page=100`
```
diversemix, line 67: If anyone wants to suggest a library for doing this feel free... but its pretty simple to do.

app/components/pages/SubmissionWizard/steps/Editors/index.js
hem-brahmbhatt: We can change the server to ensure that the role is validated against a dictionary of allowed roles. The `graphql` package may give us some protection against javascript injection as well.

diversemix: Fair point ... I guess we should at least check there is no `&` in the role string

server/xpub-model/entities/user/helpers/elife-api.js
```diff
@@ -46,16 +46,19 @@ const convertPerson = apiPerson => {
...
+    if (role) {
+      role.split(',').forEach(r => {
+        query += `&type[]=${r}`
```
hem-brahmbhatt, line 73: We can use `qs` to handle query string arrays. [Here's an example.](https://github.com/visionmedia/superagent/issues/629#issuecomment-109197779)

server/xpub-model/entities/user/helpers/elife-api.js
hem-brahmbhatt: We can use `qs` to convert the query object to a string before passing it along to `superagent`. [It handles arrays as well](https://github.com/visionmedia/superagent/issues/629#issuecomment-109197779)

server/xpub-model/entities/user/helpers/elife-api.js
diversemix: That is not using the PHP syntax :( ... like `key[]=value1&key[]=value2`

hem-brahmbhatt: Ah, that's just because it's encoded. You can set `encode` to false:

```javascript
qs.stringify(
  query, 
  { arrayFormat: 'brackets', encode: false }
)
```

Cooryd:
```
    GET https://prod--gateway.elifesciences.org:443/people?order=asc&page=1&per-page=100&type[]=senior-editor refused: not recording and no network access
```

Looks like we need to record another replay

server/xpub-model/entities/user/helpers/elife-api.js
Cooryd: Have discussed this with @diversemix it might not be worth adding a new external library for this one implementation. It doesn't add too much extra code and as we're not having to do this conversion anywhere else in the app yet it doesn't seem necessary to make this conversion process an abstract function. I would leave as is unless we need to do this elsewhere in the future.

server/xpub-model/entities/user/helpers/elife-api.js
diversemix: See conversation: https://github.com/elifesciences/elife-xpub/pull/1301#discussion_r248280837

Cooryd (approved):
Looks good. As commented if we need to any further api calls that convert queries we should look into implementing something like the `qs` library suggested by @damnhipster

PR #1306, opened 2019-01-16
749 Fix button sizes

### LINKED ISSUE #749: Button size tweaks needed
## Fix
- [x] Refactor the way buttons are styled to make this easier
- [x] Change the styles to conform to the design below ...

## The bug
Buttons are not quite the same in size and appearance as the design files, and need slight changes to be closer to intended design. 

The main problem is the height of each buttons is 50px instead of 48px and the cause seems to mostly be the 1px border - I have suggested changes below that should help but may require interpretation to apply correctly in react etc. 

Below shows the current implementation (appears the same in chrome, firefox and safari (IE/edge not tested). Next to it is the screenshot from design files, and next to that my attempts to get as close as possible to the intended designs by editing CSS in browser developer tools. I am not sure if this helps achieve the intended results in the XPub font-end code.

I have also made suggestions to increase the minimum width to 120px, and increased the side padding to give buttons more emphasis. The font should also be set to Noto Sans Semibold if font rendering issues are resolved.

**Screenshots**
image

image

image

 
In addition to above:

- [x] The SUBMIT button should look the same as the NEXT and BACK buttons.
- [x] The NEW SUBMISSION button needs to be centered as the above and needs more padding to the sides.

SemraSh:
#### Background

What does this PR do?
Changes the styles of the existing Button adding another layer of button named ButtonBase which is using the base styles. Changes the use of Button to ButtonBase on the required places.

#### Any relevant tickets
Begins #749 

👀 
style changes

#### UI Changes


SemraSh:
@chugginselifesciences could you confirm whether the `SUBMIT` button needs to be changed. Thank you.

chris-huggins:
> @chugginselifesciences could you confirm whether the `SUBMIT` button needs to be changed. Thank you.

Yes please. The submit button should be the same component as all other blue (primary) buttons

hem-brahmbhatt (commented):
We have a mechanism to override default styles on Pubsweet components such as `Button`. Any styles you add [here](https://github.com/elifesciences/elife-xpub/blob/develop/client/elife-theme/src/elements/ui/Button.js) will get applied to Pubsweet's `Button` component.

We should use the overriding mechanism rather than creating a new `ButtonBase` component otherwise we have to change imports everywhere.

diversemix:
@chugginselifesciences is looking at this now ... as we found that this does not match the sizes in the designs. Chris is just double checking and will get back to you.

Cooryd (approved):
Overrides the `Button` styling, looks good to me

PR #1312, opened 2019-01-17
749 Fix button width

### LINKED ISSUE #749: Button size tweaks needed

SemraSh:
#### Background

What does this PR do?
Fixes the size of the padding from 12px to 18px on the Button

#### Any relevant tickets

Closes #749 

👀 
css change

#### UI Changes

- [x] Has been reviewed against Designs

### DIFF
--- a/client/elife-theme/src/elements/ui/Button.js
+++ b/client/elife-theme/src/elements/ui/Button.js
@@ -60,8 +60,7 @@ const attention = css`
 export default css`
   font-size: ${th('fontSizeBaseSmall')}
   line-height: ${th('lineHeightBaseSmall')};
-  padding: ${th('space.2')};
-  padding-bottom: 9px;
+  padding: 12px 18px 9px 18px;
   text-transform: uppercase;
   font-family: ${th('fontHeading')};
   font-weight: 400;


client/elife-theme/src/elements/ui/Button.js
```diff
@@ -60,8 +60,7 @@ const attention = css`
 export default css`
   font-size: ${th('fontSizeBaseSmall')}
   line-height: ${th('lineHeightBaseSmall')};
-  padding: ${th('space.2')};
-  padding-bottom: 9px;
+  padding: 12px 18px 9px 18px;
```
diversemix, line 63: Should we express these in terms of grid units? @Cooryd ?

jyenes: I think it's ok in that way @diversemix

jyenes (approved):

hdrury1:
@chugginselifesciences and @SemraSh is it worth dealing with the semibold button labels in this PR too? I believe the text should be NotoSans font weight 400?

SemraSh:
> @chugginselifesciences and @SemraSh is it worth dealing with the semibold button labels in this PR too? I believe the text should be NotoSans font weight 400?

@hdrury1 is there a relevant ticket for that ? Could you tag that please I can make the changes in this PR.

chris-huggins:
> @chugginselifesciences and @SemraSh is it worth dealing with the semibold button labels in this PR too? I believe the text should be NotoSans font weight 400?

@hdrury1 Buttons have always been intended to be Noto Sans Semibold at 400. Browsers render them differently so it can look too heavy sometimes.  I believe `-webkit-font-smoothing: antialiased` would improve the look for enabled browsers.

Noto Sans (regular) at 400 would be too lightweight compared to original designs.

In chrome on Mac:

With Noto Sans 400 at 14px

image

With Noto Sans Semibold 400 at 14px 

image

With Noto Sans Semibold 400 at 14px with `-webkit-font-smoothing: antialiased`

image

Do you want to test across different browsers?

chris-huggins:
@hdrury1 

Currently in Firefox on Mac:

image

And safari:

image

chris-huggins (approved):

chris-huggins:
Some context... we historically had issues with inconsistent use of Noto Sans 700 and Noto Sans Semibold 400 to provide the semibold look. This essentially is the result of some weirdness with the typeface in that there is no 500 or 600 "semibold" weight for noto sans, it instead is treated as if it's a different typeface. At one point we ended up combining the 700 weight and the semibold font which ended up with what an ugly "super bold".

https://github.com/elifesciences/elife-xpub/issues/915

SemraSh:
@chugginselifesciences confirmed the above issue is fixed. 
I am merging this PR.

PR #1314, opened 2019-01-17
fix issue not displaying links for supporting files

### LINKED ISSUE #1165: "Add" & "remove all" missing for supporting files
Make sure it works (basic) on mobile... and add the footer containing "Add support files" and  "Removal all"

jyenes:
fix issue not displaying links for add or remove supporting files on mobile

closes #1165

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -5,7 +5,6 @@ import { th } from '@pubsweet/ui-toolkit'
 import styled, { css } from 'styled-components'
 import config from 'config'
 
-import media from '../../../../global/layout/media'
 import Icon from '../../../../ui/atoms/Icon'
 
 const UploadLink = styled.span`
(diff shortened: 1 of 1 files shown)


diversemix (approved):
Looks good - tested on the review environment

chris-huggins:
@jyenes @diversemix @hdrury1 Responding as I was asked to review before the merge.

I don't really like how this looks to be honest. The label wraps onto a second line on mobile which is not ideal.

**Screenshot:**

image

This "Remove all" was added because we were not going to use the trash icons for each item as designed. This label wrapping issue I think is something we'll have to live with until we add the trash icon on each item (and no longer have the "Remove all" option), which makes me think we should aim to do this as soon as we can after launch.

**From designs:**

image

diversemix:
@chugginselifesciences ... yes I saw this too. But this is definitely an improvement to what there is currently. So I decided to merge it. Please raise another ticket if you'd like this changed.

chris-huggins:
@diversemix 
> @chugginselifesciences ... yes I saw this too. But this is definitely an improvement to what there is currently. So I decided to merge it. Please raise another ticket if you'd like this changed.

I agree and was happy for it to be merged - was just adding my feedback as I was requested to review. I'll add a new ticket for the "delete each item" button. That would remove the need for "Remove all" so it would simply no longer be included.

PR #1320, opened 2019-01-18
Do not assume the manuscript is a PDF

### LINKED ISSUE #1311: Manuscript of type docx do not get exported to MECA correctly
This is because the export is assuming the manuscript is a pdf format. This needs to be templated on the format (like supplemental files)

diversemix:
#### Background

Allows the manuscript type to be flexible so it can export `docx` type

#### Any relevant tickets

Closes #1311 

- Unit tests updated
- Tested locally and checked the exported zip in AWS


server/meca/export.js
```diff
@@ -21,21 +21,11 @@ async function generate(manuscript, getContent, clientIp) {
...
-      })
-    }
-  })
+  const uploadedFiles = manuscript.files.map(file => ({
+    name: file.filename,
+    content: getContent(file),
+    type: file.type,
+  }))
 
   return archiveGenerator(manditoryFiles.concat(uploadedFiles))
```
giorgiosironi, line 30: `mandatoryFiles`?

server/meca/export.test.js
```diff
@@ -64,11 +64,11 @@ describe('MECA integration test', () => {
       expect(getFilenames(zip)).toEqual([
         '00000001.pdf',
         '00000002.pdf',
+        '7de61e41-b163-4108-9198-1492e2b54a1f.pdf',
         'article.xml',
```
giorgiosironi, line 68: for what it's worth, end2end tests about to be introduced only rely on `article.xml` so they are robust to these changes

server/meca/export.js
hem-brahmbhatt: These are other files that need to exist within the MECA package, such as the manifest.xml

hem-brahmbhatt (approved):

PR #1335, opened 2019-01-18
Remove function that is self-recursive

diversemix:
Bug found when debugging using this function. Removing it means that `File.find` uses the default implementation from `BaseModel` rather than getting in an infinite loop


Cooryd (approved):
Looks good

PR #1336, opened 2019-01-18
Add tests to confirm the behaviour of the ORM

diversemix:
Confirms the behaviour of the ORM
Put in so we can run on staging against RDS ... which *should* be the same


Cooryd (approved):

hem-brahmbhatt (approved):
This is good stuff, it's a good idea to document and capture the behaviour of the ORM with these tests, since it'll answer common questions every time someone works with it.

jyenes (approved):
LGTM

PR #1342, opened 2019-01-18
Add more tests to confirm memory and db values

diversemix:
More tests please!


diversemix:
Cherry picked onto #1343

PR #1346, opened 2019-01-21
1345 - Fix error thrown when trying to upload invalid file type as manuscript

### LINKED ISSUE #1345: Client error thrown when attempting to upload unsupported manuscript file
**Describe the bug**
An error is thrown within the client when the user tries to upload a file which is not a docx / pdf.

```
Uncaught (in promise) TypeError: Cannot read property 'name' of undefined
    at ProxyComponent.onDrop (ManuscriptUpload.js:287)
    at eval (index.js:253)
```
**Steps To Reproduce**

Drag and drop a png file onto the manuscript upload area. 

**Expected behavior**
Validation message should be shown in the app and the existing file should be removed.

Cooryd:
#### Background

A client side error was being thrown when trying to drag and drop or select via the file finder an unsupported file type. This adds a check on the length of the files array passed to the `onDrop` to prevent the error when file name is being stored.

#### Any relevant tickets

closes #1345

- [x] Browser tests


diversemix (approved):
lgtm

PR #1359, opened 2019-01-21
Enabling Google Analytics debugging

hem-brahmbhatt:
- Suppresses Google Analytics warnings locally and on review environments.
- Prints out debugging information on Google Analytics, including setup and events. 

This is useful for us to test out changes to Google Analytics outside of Staging and Production.

This is enabled by default, and only disabled in Staging and Production, or any environments with a GA ID configured.

The output can be pretty verbose, so I'm open to suggestions. You can view the debugging output by going to the review environment and viewing the console.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -17,7 +17,7 @@ const history = createHistory()
 const store = configureStore(history, {})
 
 const initializeReactGA = googleAnalyticsId => {
-  const debug = googleAnalyticsId === 'test' ? true : false
+  const debug = googleAnalyticsId === 'test'
   ReactGA.initialize(googleAnalyticsId, { debug })
   ReactGA.pageview('/')
 }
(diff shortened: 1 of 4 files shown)


diversemix:
I'm not clear on how this can:
> Prints out debugging information on Google Analytics, including setup and events

Debug only is on when the id is set to `test` - and I'm guessing this does not connect to the real GA ... could you explain?

SemraSh (approved):

hem-brahmbhatt:
@diversemix It's only purpose was to allow us to view tracking events in environments where tracking hadn't been set up. 

This is currently controlled by whether the licence ID has been set. I'll change it so it's controlled via our config, so we can have the debug mode independent of the licence ID, i.e.

```
googleAnalytics: {
  id: 'some-id',
  debug: true
}
```

This should allow us to enable debug mode more granularly per environment

Cooryd (approved):
Looks good

hem-brahmbhatt:
screen shot 2019-01-22 at 11 19 59

diversemix (approved):

PR #1360, opened 2019-01-22
Vim support. Make script work locally too by increasing delays

diversemix:
- Adds commonly used vim files to ignore
- Increase delays and use key strokes where possible to allow the script to run locally.

### DIFF
--- a/scripts/do_submission.sh
+++ b/scripts/do_submission.sh
@@ -10,7 +10,13 @@ set -e
 WID=$(xdotool search --all --onlyvisible --desktop 1 --class "google-chrome")
 CMD="xdotool windowactivate --sync ${WID}"
 WAIT=0.5
-$CMD mousemove 1816 324 click 1
+
+$CMD key F5
+sleep 3
+$CMD key --repeat 7 Tab
+sleep $WAIT
+$CMD key space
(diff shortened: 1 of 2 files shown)


scripts/do_submission.sh
```diff
@@ -10,7 +10,13 @@ set -e
 WID=$(xdotool search --all --onlyvisible --desktop 1 --class "google-chrome")
 CMD="xdotool windowactivate --sync ${WID}"
 WAIT=0.5
-$CMD mousemove 1816 324 click 1
+
+$CMD key F5
+sleep 3
```
giorgiosironi, line 15: [Bare sleeps](https://www.martinfowler.com/articles/nonDeterminism.html#AsynchronousBehavior) are not a robust solution to asynchronous behavior, why is this script not using the same tools as `test:browser` or some standard browser automation with bindings for a programming language?

giorgiosironi: Answered in person: exploratory testing usage rather than automated testing, usage from a laptop against both development and staging

giorgiosironi (approved):
Good for local usage for now, but https://github.com/elifesciences/elife-xpub/issues/1365 will evolve this for the long-term

Issue #1362, opened 2019-01-22
Refreshing page show Supporting file incorrectly


diversemix:
If you refresh the page while Supporting files are uploading then they appear to be successfully uploaded even if they are not. This was demonstrated by going through and submitting the Submission and seeing the error  in #1361 

Depends on #1373 #1374

Need to validate based on the status of the `manuscript.files[ ]` and show the correct state (i.e. spinner accordingly)

diversemix:
@diversemix is now breaking this down

diversemix:
Breakdown done

PR #1387, opened 2019-01-23
1367 validate file size

### LINKED ISSUE #1367: Add validation state for supporting file that is too large
**Describe the bug**
When attempting to upload an invalid file, e.g. one more than 100MB, there is no indication why it cannot be uploaded. The file manager closes with the UI unchanged.

**Steps To Reproduce**
1. Navigate to the file page.
2. Upload a manuscript successfully.
3. Click the link to upload supporting file(s).
4. Observe that file manager closes with no file added.

**Expected behavior**
We should feed back to the user that the file they have attempted to upload is invalid. They should then be able to clear this validation message by deleting the item from the list.

Design:

screen shot 2019-01-22 at 12 46 23

We should cap the number of items displayed to the user at 10, regardless of whether the items are valid or invalid.

The user should be able to continue through the wizard while invalid files are listed.

The user can also re-upload a file with the same name to override the error. (<- How difficult is this techinically? We could potentially drop for launch)

SemraSh:
#### Background
Adds validation for supporting files larger than 100mb

#### Any relevant tickets
Closes #1367

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -43,13 +43,13 @@ const StyledDropzone = styled(({ setRef, ...rest }) => (
       border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorBorder')};
     `};
 `
-const FileBlock = styled(({ uploadError, ...rest }) => <Flex {...rest} />)`
+const FileBlock = styled(({ fileError, ...rest }) => <Flex {...rest} />)`
   border-top: ${th('borderWidth')} ${th('borderStyle')} ${th('colorBorder')};
   ${props =>
-    props.uploadError
+    props.fileError
       ? css`
           margin: -1px -1px 0 -1px;
-          border: ${th('borderWidth')} ${th('borderStyle')} ${th('colorError')};
+          color: ${th('colorError')};
         `
       : css`
           &:first-child {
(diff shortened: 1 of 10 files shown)


diversemix:
I tested this in the review environment - looks good.

Minor point but the text starts with a capital 'M' and not a dash and lower case as per the spec... but I'm sure you know that anyway :wink:

SemraSh:
Confirmed with Chris that the design is as implemented in this pr.

jyenes (approved):
we have created a test browser for this but GitHub doesn't allow us to upload files larger than 100 MB

app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
```diff
@@ -185,40 +239,32 @@ class SupportingUpload extends React.Component {
...
+                {file.success && <UploadSuccessIcon />}
+                {(file.error || file.rejected) && <UploadFailureIcon />}
+              </IconHolder>
+              <FileHolder>
+                <FileName data-test-id="file-block-name">
+                  {file.file.name ? file.file.name : file.file.filename}
+                </FileName>
+                {file.rejected && (
+                  <ErrorMessage data-test-id="file-block-error">
+                    Must be less than 100mb
```
giorgiosironi: should read `config.fileUpload.maxSizeMB`?

diversemix (approved):

hdrury1:
Meets Product reqs. At the moment refreshing the page removes the error but I presume this will be addressed when tackling #1362

hem-brahmbhatt (approved):
Code looks good

jyenes:
> Meets Product reqs. At the moment refreshing the page removes the error but I presume this will be addressed when tackling #1362

@hdrury1 why do have to persist an invalid submission?
I mean, the file wasn't uploaded to S3, and also wasn't persisted in the database, is just a warning to show to the user, from my point of view, we shouldn't persist this status in the database. am I wrong?

hdrury1:
Hi @jyenes yes I think this is the case (at least right now, see https://elifesciences.slack.com/archives/C6LKMFK5Z/p1548337091016000), and design is happy with it. Good to merge!

PR #1400, opened 2019-01-28
1362 - File upload statuses displaying correctly after refresh

### LINKED ISSUE #1362: Refreshing page show Supporting file incorrectly

Cooryd:
#### Background

Currently the default client display for an uploaded supporting file is the upload successful state. When refreshing the app while a transaction is still in progress, it doesn't correctly represent the state in which the file data is in the db so files still in the process of being stored in S3 are showing as successful.

This PR adds in the correct initial display of the files state on, refresh however further work will be needed in a separate piece of work to reconnect to a web-socket to track the progress in real time.
 
#### Any relevant tickets

closes #1362

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -152,7 +152,9 @@ class SupportingUpload extends React.Component {
       files: props.files.map((file, index) => ({
         file,
         id: index,
-        success: true,
+        success: file.status === 'STORED',
+        loading: file.status === 'CREATED' || file.status === 'UPLOADED',
+        error: file.status === 'CANCELLED',
       })),
     }
   }


hdrury1:
Looks good visually- I'm getting an initial error state when first hitting refresh which I presume is the desired outcome.

hdrury1 (approved):

Cooryd:
> Looks good visually- I'm getting an initial error state when first hitting refresh which I presume is the desired outcome.

I wouldn't expect anything to come back in an error state by default. You might need to reset your db to include the status data which won't exist on old file uploads. The best way to test this is to try to upload a large file > 10MB and then refresh straight away. The app should reload and that file should still appear with a loading spinner instead of the completed arrow icon.

app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
```diff
@@ -152,7 +152,9 @@ class SupportingUpload extends React.Component {
       files: props.files.map((file, index) => ({
         file,
         id: index,
-        success: true,
+        success: file.status === 'STORED',
```
diversemix, line 155: Cory and I are making a note - that when we eventually write to the ApolloCache we should cosolidate this into a clientStatus (i.e. calculated from the server status)

diversemix (approved):

PR #1410, opened 2019-01-29
1397 Fix submission page dropdown ui

### LINKED ISSUE #1333: Fix height of subject area picker before/after selection (and other fixes)
**Describe the bug**
The height increases when a selection is made.

**Expected behaviour**
The height would be 48px (including border) before and after selection is made (unless selectedItem(s) are more than 1 line tall - which would increase height by 48px for each line/item.

Also:

- Update text and styling of the placeholder text (should match other inputs)
- Remove dropdown icon/arrow
- Fix selectedItem height

**Current**

image

**Intended**

image

Select Item - should be 36px tall, 5px padding within the input, and 6px between selectedItems

image

### LINKED ISSUE #1397: 48px height for "Subject Areas" and "Article type" pickers/input
**Describe the bug**
Inconsistent height. There are other related issues described in #1333. I've added this ticket to allow us to cover just the height fix in the short term. 

@SemraSh said she will try to find time to make this small fix before launch so I'm adding it to To Do as a reminder.

If it is a quick enough fix I would recommend also fixing the "Subject areas" label to 14px font instead of 16px.

image

SemraSh:
#### Background

What does this PR do?
Fixes the ui to match the designs

#### Any relevant tickets

Closes #1333 
Closes #1397 

👀 

#### UI Changes

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Submission/SubjectAreaDropdown.js
+++ b/app/components/pages/SubmissionWizard/steps/Submission/SubjectAreaDropdown.js
@@ -25,8 +25,11 @@ const Label = styled.label`
   font-size: 14px;
 `
 
-const SelectLimitMessage = styled.p`
+const SelectLimitMessage = styled.div`
   color: ${th('colorSuccess')};
+  min-height: ${th('lineHeightBaseSmall')};
+  margin-top: ${th('space.1')};
+  font-size: ${th('fontSizeBaseSmall')};
 `
 
 const subjectAreas = config.client.majorSubjectAreas
--- a/app/components/pages/SubmissionWizard/steps/Submission/SubjectAreaDropdown.js
+++ b/app/components/pages/SubmissionWizard/steps/Submission/SubjectAreaDropdown.js
@@ -21,6 +21,10 @@ const StyledCrossIcon = styled(CrossIcon)`
 
 const Root = styled.div``
 
+const Label = styled.label`
+  font-size: 14px;
+`
+
 const SelectLimitMessage = styled.p`
   color: ${th('colorSuccess')};
 `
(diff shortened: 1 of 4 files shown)


app/components/pages/SubmissionWizard/steps/Submission/SubjectAreaDropdown.js
```diff
@@ -142,7 +148,7 @@ class SubjectAreaDropdown extends React.Component {
           ),
         },
         backgroundColor: this.props.theme.backgroundColor,
-        minHeight: this.props.gridUnit,
+        minHeight: `48px`,
```
Cooryd: This could use `minHeight: ${th('space.5')}`  instead so it uses the grid sizing still

chris-huggins:
The height looks correct to me, however the article type dropdown has a span with 6px padding which is causing too much space on the inside left edge of the input. If this 6px is removed then I think it's fine.

There are other small issues around the focus state conflicting with validation states but I will raise those in another ticket.

chris-huggins:
Hi @SemraSh - it looks great. I think the only things now that I would ideally see changed are:

- the positive validation text when 2 items are added is larger than the usual error state
- The green validation state wasn't shown when I opened from the dashboard, even thought two subject areas were already added 

The positive state when max is reached - too large/too spaced:

image

Compared to the error state, which looks correct:

image

Thanks

chris-huggins (approved):
Looks good to me.

app/components/pages/SubmissionWizard/steps/Submission/SubjectAreaDropdown.js
```diff
@@ -142,7 +153,7 @@ class SubjectAreaDropdown extends React.Component {
           ),
         },
         backgroundColor: this.props.theme.backgroundColor,
-        minHeight: this.props.gridUnit,
+        minHeight: `${gridUnitValue * 8}px`,
```
Cooryd: This could be a `space.5` instead. I think we give preference to the `space` unit over a multiple of the `gridUnit` elsewhere in the app.

SemraSh: In customReactSelectStyles object both gridUnitValue and gridUnit are used. I didn't wanted to mix both space and gridValue. I am keeping that consistency in the styled definitions but this object is using them as value and number. Are you suggesting I should change all of these to use the space instead and change the gridUnit to something like spaceUnit and use that ?

SemraSh: I can change all of them though ?

SemraSh: Please see the latest changes on that @Cooryd

Cooryd (approved):

Issue #1375, opened 2019-01-22
Update playbook with how to restore the database from a snapshot


diversemix:
Going to be done by @giorgiosironi as the repo might get renamed

giorgiosironi:
Testing out a snapshot on a new server, from https://console.aws.amazon.com/rds/home?region=us-east-1#snapshot:engine=postgres;id=rds:elife-xpub-staging-2019-01-29-03-30

giorgiosironi:
Instance being created at https://console.aws.amazon.com/rds/home?region=us-east-1#database:id=elife-xpub-staging-restore-test;is-cluster=false

giorgiosironi:
Instance ready, trying to generate a `pg_dump` of it.

giorgiosironi:
Either I copied the wrong hostname or there are some visibility problems between subnets:
```
pg_dump: [archiver (db)] connection to database "elife-xpub-staging-restore-test" failed: could not connect to server: Connection timed out
        Is the server running on host "elife-xpub-staging-restore-test.cxyopn44uqbl.us-east-1.rds.amazonaws.com" (10.0.4.4) and accepting
        TCP/IP connections on port 5432?
```

giorgiosironi:
The [security group](https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#SecurityGroups:search=sg-e4a1de81;sort=groupId) seems to allow only traffic from inside the security group. It seems an old one so I'll detach it, must be some default.

giorgiosironi:
Changed to `elife-xpub-staging-VPCSecurityGroup-16L6J45AL7U4W`. This needs better automation.

giorgiosironi:
```
#!/bin/bash
set -e

docker run \
    -e PGHOST=elife-xpub-staging-restore-test.cxyopn44uqbl.us-east-1.rds.amazonaws.com \
    -e PGPORT=5432 \
    -e PGUSER=$PGUSER \
    -e PGPASSWORD=$PGPASSWORD \
    -e PGDATABASE=elifexpubstaging \
    postgres:10 \
    pg_dump
```
got me a dump:
```
elife@staging--xpub:~$ ls -l /ext/tmp/dump.log 
-rw-rw-r-- 1 elife elife 697369 Jan 29 13:02 /ext/tmp/dump.log
```

giorgiosironi:
The `directory` format is better as it's compressed by default and can be restored both while dropping or not dropping the existing tables:
```
[...]

PR #1412, opened 2019-01-29
Add playbook for elife-xpub specific database restore

giorgiosironi:
For https://github.com/elifesciences/elife-xpub/issues/1375

### DIFF
--- a/docs/deploying/database.md
+++ b/docs/deploying/database.md
@@ -116,3 +116,7 @@ The `PGHOST` being used here is the RDS node that has just been created (2nd arg
 ```
 
 The `PGHOST` being used here is the existing RDS node for this environment (2nd argument).
+
+#### 7. Remove the new node
+
+Don't worry about this: when the emergency has been addressed, ask your administrator to clean up the temporary RDS node and its own snapshots.
--- a/docs/deploying/database.md
+++ b/docs/deploying/database.md
@@ -31,3 +31,88 @@ Migrations are [run automatically](https://github.com/elifesciences/elife-xpub-f
 Migrations have to be backward compatible as every deploy has a critical section where the old application code is still running after migrations have been applied.
 
 There is no automated rollback of migrations.
+
+## Backup and restore
+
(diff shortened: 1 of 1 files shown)


docs/deploying/database.md
```diff
@@ -31,3 +31,88 @@ Migrations are [run automatically](https://github.com/elifesciences/elife-xpub-f
 Migrations have to be backward compatible as every deploy has a critical section where the old application code is still running after migrations have been applied.
 
 There is no automated rollback of migrations.
+
+## Backup and restore
+
+RDS snapshots are taken daily and [retained for 4 weeks in `prod`](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L1970).
+
+The scripts `dump-database.sh` and `restore-database.sh`, on PATH in the servers, can respectively be used to perform a new dump from any Postgres host and to restore that dump.
```
diversemix: Where are these scripts? Would be good to add their location here.

docs/deploying/database.md
```diff
@@ -31,3 +31,88 @@ Migrations are [run automatically](https://github.com/elifesciences/elife-xpub-f
...
+
+The `PGHOST` being used here is the RDS node that has just been created (2nd argument).
+
+#### 6. Restore the dump on the existing node
+
+```
+ ./restore-database.sh /ext/tmp/dump-afternoon elife-xpub-staging.cxyopn44uqbl.us-east-1.rds.amazonaws.com
+```
+
+The `PGHOST` being used here is the existing RDS node for this environment (2nd argument).
```
diversemix, line 118: Do we need instructions for removing the "new node" ... which I guess was only done to perform the SQL extraction?

giorgiosironi: Assuming this is done in an emergency, there's no harm in leaving it around, an administrator can remove it later on; didn't want to make this guide even longer.

docs/deploying/database.md
giorgiosironi: they are on the `PATH` so they should be accessible wherever you are, but I'll reword

giorgiosironi: (they are in `/usr/local/bin` but you shouldn't have to know that to use them)

Cooryd (approved):

PR #1416, opened 2019-01-30
Work around permissions problem on formula repository

giorgiosironi:
Avoids review environments deployment breaking on changes to `elife-xpub-formula`.

For full description of the issue: https://github.com/elifesciences/issues/issues/4634

### DIFF
--- a/scripts/helm_deploy.sh
+++ b/scripts/helm_deploy.sh
@@ -17,6 +17,8 @@ if [ ! -d /tmp/elife-xpub-formula ]; then
     cd /tmp/elife-xpub-formula
 else
     cd /tmp/elife-xpub-formula
+    # workaround for https://github.com/elifesciences/issues/issues/4634
+    git reset --hard
     git pull origin master
 fi
 


scripts/helm_deploy.sh
```diff
@@ -17,6 +17,8 @@ if [ ! -d /tmp/elife-xpub-formula ]; then
     cd /tmp/elife-xpub-formula
 else
     cd /tmp/elife-xpub-formula
+    # workaround for https://github.com/elifesciences/issues/issues/4634
+    git reset --hard
```
diversemix, line 21: do we also want to do a `clean` just to make sure?

diversemix (approved):

scripts/helm_deploy.sh
giorgiosironi: the problem seems to be that existing files [change permissions](https://github.com/elifesciences/issues/issues/4634#issue-385640763), hence `git clean` shouldn't be needed as it works on files that are not under version control (and that won't break `git pull` in the next line which is where the script stops now)

Issue #925, opened 2018-10-30
Populate ejp_name (db table) in all environments


diversemix:
This needs to be populated with the exceptions to name splitting done for the reviewers.

Data: https://docs.google.com/spreadsheets/d/1vrL_FgEwAaM54GC_GEut_Fau6uvkzuJ1xCYbjTKdBCA/

There are names containing special characters, and names which are fully upper/lower case. The plan is to process these separately in case any issues arise from them.

### Background knowledge
- [x] find DB table
- [x] understand how table is used
- [x] check if indexes will need updating
- [ ] check if review environment needs updating
- [x] look for distinctions within the names and highlight them to be processed seperately

### Normalising the data set
This has been done on the spreadsheet itself. New sheets have been created for the following:
- [x] filter out names with special characters
- [x] filter out names which are all lower case
- [x] filter out names which are all upper case

### Export the data set
- [x] export spreadsheet to CSV
  - copy values from *first* and *last* columns to new sheet from filtered view
  - File —> Download as… —> Comma-Separated values
- [x] add sequence to generate IDs on ejp_name table
  - http://www.postgresqltutorial.com/postgresql-serial/
- [x] import CSV to ejp_name table
  - `docker cp /path-to-csv.csv elife-xpub_postgres_1:data.csv`
  - `docker exec -it elife-xpub_postgres_1 psql --user hem`
  - `COPY ejp_name FROM '/data.csv' DELIMITERS ',' CSV;`
- [ ] update locally and test

### Update Staging
- [ ] SCP csv file to staging
- [ ] update staging and test
- [ ] do the above for names with special characters, upper case, and lower case

### Documentation
- [ ] add steps to export data
- [ ] add steps to copy the data to the designated environment
- [ ] add steps to import the data to the `ejp_name` table

hdrury1:
As if by magic, the list appeared in the team drive.

diversemix:
Linked to #1370

hdrury1:
Anything that is a FALSE match in this list:

https://docs.google.com/spreadsheets/d/1vrL_FgEwAaM54GC_GEut_Fau6uvkzuJ1xCYbjTKdBCA/edit#gid=190468026

hem-brahmbhatt:
@hdrury1 The list contains some unusable names, I'll filter those out. 

I'll do the names with special characters separately, since it will need a bit more testing.

giorgiosironi:
[...]

PR #1422, opened 2019-01-31
WIP: added migration for auto-incrementing ID in ejp_name

hem-brahmbhatt:
#### Background

Allows IDs to be automatically generated when adding names to the `ejp_names` table. This is useful when exporting thousands of rows given to us by EJP.

#### Any relevant tickets

#925 

Unit tests run the migrations, these should be passing.


diversemix (changes requested):
What's wrong with `gen_random_uuid()` ?

diversemix:
Not necessary

Issue #1383, opened 2019-01-23
Over-write supporting files with the same name


hdrury1:
If a user uploads a second file with the same name as the first, the second file should override it in the database (currently we allow for multiple files with the same filename to be stored, but this does not persist through to export).

On the front end, rather than displaying both the first and second file in the list of uploaded files, only the second would be displayed to the user.

subtask:
- [x] return the id on the upload mutation (#1424)
- [ ] multiple supporting file uploads in parallel (#1449)

hem-brahmbhatt:
Moved into backlog until #1449 is done.

diversemix:
@damnhipster ... could you update with the text you put in Slack please?

hem-brahmbhatt:
> @damnhipster ... could you update with the text you put in Slack please?

We've decided to park this issue. Instead, we thought it's better to do the work after we've refactored the code in #1449 to support multiple parallel uploads.

It should be easier to do this after we've refactored, and we won't have to write this feature two times (once now and once after refactor).

PR #1426, opened 2019-01-31
adding id to the state once uploading a new file

### LINKED ISSUE #1424: return Id on the upload file mutation
is required o overwrite supporting files

jyenes:
we have to add id in order to be able to replace the supporting file, once a new upload is coming with the same name 

closes #1424  
unblocks #1383

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -138,6 +138,7 @@ function* fileUploadGenerator(files, uploadFile) {
     yield {
       upload: uploadFile(files[fileIndex].file),
       fileId: files[fileIndex].id,
+      filename: files[fileIndex].file.name,
     }
     fileIndex += 1
   }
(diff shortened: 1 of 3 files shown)


SemraSh (approved):
Looks good to me 👍

app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
```diff
@@ -185,11 +193,13 @@ class SupportingUpload extends React.Component {
...
   }
 
-  updateFileState = (fileId, newState) => {
+  updateFileState = (fileId, newState, id) => {
     const newFilesState = [...this.state.files]
     const fileIndex = newFilesState.findIndex(file => file.id === fileId)
     if (fileIndex > -1) {
-      newFilesState[fileIndex] = { ...newFilesState[fileIndex], ...newState }
+      const updatingState = { ...newFilesState[fileIndex] }
+      updatingState.file.id = id
```
diversemix, line 201: why do we pass in `id` ? ... How is it different from `fileId`

app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
```diff
@@ -168,10 +169,17 @@ class SupportingUpload extends React.Component {
       } else {
         result.value.upload
           .then(data => {
-            this.updateFileState(result.value.fileId, {
-              success: true,
-              loading: false,
-            })
+            const updatedFile = data.data.uploadSupportingFile.files.filter(
+              file => file.filename === result.value.filename,
```
diversemix, line 173: Isn't the `id` on `result` ?

app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
jyenes: id is the id from the dataBase, we need to pass it to store id within the state, fileId is the array id and it's being used to remove loading spinner, so basically to know wich row is being updated.

Issue #1253, opened 2019-01-07
Create unit tests for xpub-controller classes


Cooryd:
#1203 refactored existing code by creating a new controller collection of files. The merging of this PR issue introduced some bugs which may have been captured by unit tests. Unit tests should be written for each of the classes found in `server/xpub-controller`

Cooryd:
There are already some tests for the remove supporting files function found in the `xpub-server` directory, these should be moved across. (see #1263)

PR #1444, opened 2019-02-04
[SLICING] 1438 - file_status on manuscript

### LINKED ISSUE #1438: Add a filestatus value to manuscript that consolidates the upload progress of all linked files
## Problem / Motivation

Relating to #1361, The manuscript needs to be aware of the upload status of its manuscript and supporting files so that logic can be added to prevent submission to ejp until all files are in S3.

## Proposed solution
A `filestatus` column needs adding to the manuscript table. This will store a enumerated value of `READY` or `CHANGING` which signifies if all related files are uploaded/cancelled or still in the process of uploading. This will default to `READY` when the manuscript is first created and will be updated when a `validate()` function on the `manuscript` object is called. 

`valide()` should be called on the related `manuscript` every time the files state is changed.

Cooryd:
#### Background

Adds a `file_status` column to the manuscript table for consolidating the statuses of all related files. This is set by a `validate()` function on the `Manuscript.

To correctly call this `validate()` function on `Manuscript` objects after every `File.setStatus()` the code within the resolvers for file upload / delete mutations has been moved into its corresponding `xpub-controller` class. Tests are working however most of the tests that lived next to the resolver code should now be ported across to the controller directory as part of #1253

#### Any relevant tickets

Closes #1438

 Server side changes only.


server/xpub-model/entities/manuscript/index.js
```diff
@@ -225,6 +234,18 @@ class Manuscript extends BaseModel {
...
 
+  async validate() {
+    // Check all files are are stored or correctly cancelled
+    let fileStatusReady = true
+    this.files.forEach(file => {
+      if (file.status !== 'STORED' && file.status !== 'CANCELLED')
+        fileStatusReady = false
+    })
+
+    this.fileStatus = fileStatusReady ? 'READY' : 'CHANGING'
```
giorgiosironi, line 245: I like that this is an enum rather than a limited boolean

server/xpub-model/entities/file/index.js
```diff
@@ -25,23 +26,33 @@ class File extends BaseModel {
     }
   }
 
-  async save() {
+  async save(userId) {
     await this.$query().upsertGraphAndFetch(this)
 
+    const manuscript = await Manuscript.find(this.manuscriptId, userId)
```
giorgiosironi: `File` object loading `Manuscript` objects on save, and saving them in turn? This looks like a lot of coupling between the two, as they are transactionally saved together. Thinking... :thinking:

server/xpub-model/entities/file/index.js
```diff
@@ -25,23 +26,33 @@ class File extends BaseModel {
     }
   }
 
-  async save() {
+  async save(userId) {
     await this.$query().upsertGraphAndFetch(this)
```
giorgiosironi: is this the line that actually saves the `File` object?

server/xpub-model/entities/file/index.js
```diff
@@ -25,23 +26,33 @@ class File extends BaseModel {
...
+
+    if (!manuscript) {
+      throw new Error(
+        `Unable to find Manuscript of ID: ${
+          this.manuscriptId
+        } assigned to File: ${this.id}`,
+      )
+    }
+
+    await manuscript.validate()
```
giorgiosironi: suggestion:
- pass in the `Manuscript` object as an argument here (to make the runtime dependency explicit, if the layering/framework allows it)
- transform this into `manuscript.fileSaved(this.filename, this.status)` (probably needs fewer arguments) which will be the domain event handler

This is still transactional on both objects, but eventually it could make the boundary easier to see when supporting parallel file uploads

server/xpub-model/entities/file/index.js
Cooryd: Yep, this is where `objection` sends the db data

server/xpub-model/entities/file/index.js
hem-brahmbhatt: In fact, we've avoided including models with models except for the AuditLog model, which can get included in other models.

If we want to update models in conjunction with each other, we should do this within `xpub-controller`

Cooryd: I'll port all of the uploads / deletes to the `manuscript` controller and call validate at that level

giorgiosironi: Interesting, let's see how this will look like after that

server/xpub-server/entities/manuscript/resolvers/uploadSupportingFile.js
```diff
@@ -1,46 +1,11 @@
...
-    await fileEntity.updateStatus('STORED')
-  } catch (err) {
-    await fileEntity.updateStatus('CANCELLED')
-    await fileEntity.delete()
-    throw err
-  }
-
-  return Manuscript.find(id, userUuid)
+  const files = new SupportingFiles(S3Storage, id, userUuid)
+  const manuscript = await files.upload(file)
```
hem-brahmbhatt: Seems weird that `files.upload` returns a manuscript.

Cooryd: Agreed, I've made it return the `File` object and then we call `find` on `Manuscript`

server/xpub-model/model.test.js
```diff
@@ -34,7 +34,7 @@ describe('related objects behave as we expect', () => {
 
       const file = await File.find(manuscript.files[0].id)
       file.type = 'updated'
-      await file.save()
+      await file.save(userId)
```
Cooryd: This userId isn't required any more so i'll remove it.

diversemix (commented):
1) I think we all need to take a view of whether we try and break this down and merge in slices.

2) I don't see any validation being done on `fileStatus` - or is that coming next?

Cooryd:
Could you elaborate on validation? I thought that was taken care of by the objects schema ?

```
        fileStatus: {
          type: 'string',
          enum: ['READY', 'CHANGING'],
          default: 'READY',
        },
```

server/xpub-controller/index.js
```diff
@@ -1,9 +1,11 @@
 const SupportingFiles = require('./supportingFiles')
 const Manuscript = require('./manuscript')
 const S3Storage = require('./s3Storage')
+const ScienceBeamApi = require('./scienceBeamApi')
```
hem-brahmbhatt, line 4: Would be good to create a `xpub-client` package to contain the `ScienceBeamApi`, and the `S3Storage`, since it's an isolated piece of functionality.

hem-brahmbhatt: It also means we can rework the controller to follow any pattern in the future, including redux or something.

Cooryd:
Closing this without merging. This was split into a number of smaller PR's that are now merged.

PR #1445, opened 2019-02-04
WIP: 1383 overwrite supporting files same name

### LINKED ISSUE #1383: Over-write supporting files with the same name

jyenes:
#### Background
you can add any time the same file, with the same name, this PR is meant to avoid this.

What does this PR do?
Overwrites supporting files with the same name
#### Any relevant tickets
closes #1383 

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -135,14 +135,10 @@ const ErrorMessage = styled.span`
 
 function* fileUploadGenerator(files, uploadFile, removeSingleFile) {
   for (let fileIndex = 0; fileIndex < files.length; ) {
-    console.log('111111111')
-    console.log(files[fileIndex].file)
-    console.log('222222222')
-    console.log(files[fileIndex].file.id)
     yield {
       upload: uploadFile(files[fileIndex].file),
       fileId: files[fileIndex].id,
-      remove: removeSingleFile(files[fileIndex].file),
+      remove: fileId => removeSingleFile(fileId),
       filename: files[fileIndex].file.name,
     }
     fileIndex += 1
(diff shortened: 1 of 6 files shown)


app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
```diff
@@ -162,32 +163,45 @@ class SupportingUpload extends React.Component {
...
+          file =>
+            (file.file.filename || file.file.name) === result.value.filename,
+        )
+
+        if (index > -1) {
+          await result.value.remove(this.state.files[index].file.id)
+          // TODO: now we have to upload the new one.
+          // console.log('now we have to upload the file again!')
+          // console.log(response)
+        } else {
```
hem-brahmbhatt: This may not need to be an `else`.

We can remove the file if it's a duplicate, and always upload the new file the user is trying to upload.

app/components/pages/SubmissionWizard/steps/Files/index.js
```diff
@@ -190,27 +197,43 @@ const FilesPageContainer = ({
                   {uploadSupportFiles => (
                     <Mutation mutation={DELETE_SUPPORTING_FILES_MUTATION}>
                       {removeSupportFiles => (
-                        <SupportingUpload
-                          data-test-id="supportingFilesUpload"
-                          files={submissionFiles.filter(
-                            file => file.type === SUPPORTING_FILE,
+                        <Mutation mutation={DELETE_SINGLE_SUPPORTING_FILE}>
```
hem-brahmbhatt, line 200: I think we're better off renaming to the following:
`DELETE_ALL_SUPPORTING_FILES_MUTATION`
`DELETE_SUPPORTING_FILE_MUTATION`

app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
```diff
@@ -133,11 +133,12 @@ const ErrorMessage = styled.span`
   font-size: 12px;
 `
 
-function* fileUploadGenerator(files, uploadFile) {
+function* fileUploadGenerator(files, uploadFile, removeSingleFile) {
```
diversemix, line 136: Could you explain why the client needs to know about "removeSingleFile" ? - thanks

SemraSh: Client needs to know whether one file is removed ( replaced for now ) or all of them, as the only option to remove the files for now is to remove them all. But when we upload a repeated file we want to remove that single file and then upload the new version of that, we also need to show the user that this file is changed by adding the loading state as the normal behaviour. Also there are errors that are only defined in the client side, such as `max size exceeded` we want to show that when they try to upload a large repeated file. We could have also name these `removeFile` and `removeFiles`... if we will keep both in the future.

hdrury1: We'll eventually be replacing the 'remove all' functionality with the option to remove one file at a time.

hem-brahmbhatt:
Closed until #1449 has been done.

PR #1446, opened 2019-02-05
[EXT] Fix for error message not displayed when upload fails

erezmus:
#### Background

What does this PR do?

Fixes lack of error message if upload of manuscript fails.

#### Any relevant tickets

None

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/ManuscriptUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/ManuscriptUpload.js
@@ -90,21 +90,6 @@ const DropzoneContent = ({
   dropzoneOpen,
   fileName,
 }) => {
-  if (conversion.converting) {
-    return (
-      <React.Fragment>
-        <StyledUploadIcon percentage={conversion.progress} />
-        <FileName>{fileName}</FileName>
-        <UploadInstruction
-          data-test-conversion="converting"
-          data-test-id="dropzoneMessage"
-        >
-          Manuscript is uploading {conversion.progress}%
-        </UploadInstruction>
-      </React.Fragment>
-    )
-  }
-
   if (errorMessage) {
     return (
       <React.Fragment>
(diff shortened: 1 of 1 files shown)


app/components/pages/SubmissionWizard/steps/Files/ManuscriptUpload.js
```diff
@@ -123,6 +108,21 @@ const DropzoneContent = ({
       </React.Fragment>
     )
   }
+  if (conversion.converting) {
```
Cooryd, line 111: While the reasoning behind moving this down seems straight forward the component isn't clearing down the `errorMessage` when a new valid file is being uploaded so while the replacement file is being uplaoded you are still shown the error message with this change.

Cooryd (changes requested):
This change introduces or re-introduces (as I remember this being a problem previously) a visual bug when uploading a valid file after uploading a failed file. With a working S3 instance (or working mocked S3 instance) the component already behaves as expected. I think we should re-look at this component and its message display logic to either make it clearer whats happening or just generally improve it.

erezmus:
I'll need to get the fakes3 working on my machine to explore that. Any suggestions as to how to make it work?

Another possibility is to add some tests to reproduce the behaviour.

Cooryd:
It may be that the config file isn't looking in the right place when attempting to send files to the service https://github.com/elifesciences/elife-xpub/blob/develop/docs/developing/quick-start.md#aws-s3 details setting up a `local-development` config file and explains how to configure the app to look at the `fakes3` container. 

Hint - If you're running the server while changing a config file you'll need to stop the server (cmd + c) then re-run the server command `npm run server` to register the change to the config as there is a 'compile' process that runs over the config files at start up

Cooryd:
Closing pending further investigation as part of #1490

PR #1451, opened 2019-02-06
refactor and make the uploads running on parallel

### LINKED ISSUE #1449: Multiple "Supporting File" uploads in parallel

jyenes:
#### Background
we need the uploads running in parallel

What does this PR do?
closes #1449 

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
+++ b/app/components/pages/SubmissionWizard/steps/Files/SupportingUpload.js
@@ -133,17 +133,6 @@ const ErrorMessage = styled.span`
   font-size: 12px;
 `
 
-function* fileUploadGenerator(files, uploadFile) {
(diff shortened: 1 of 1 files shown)


hem-brahmbhatt (approved):
Code looks good

diversemix:
Does this work? I had though we did it sequentially for a reason ... that being a race condition on the server. @hdrury @BlueReZZ  @cooryd ?

I will check ASAP

diversemix:
OK ... we have no tests for this :sob: ... But by code inspection is *should* be ok.

Cooryd:
Race conditions were the reason we were doing this sequentially originally, seeing as we are calling `find()` before every operation and save over the manuscript though I would assume this would work in most scenarios now. This should be thoroughly tested (manually) though to make sure there aren't any possible error scenarios ie: some files failing or the page reloading during upload.

PR #1454, opened 2019-02-06
1452 Refactor module resolver

### LINKED ISSUE #1452: Refactor paths to resolve for ui and global components
As part of *issue for refactoring paths* and future refactoring, we'll be moving files around. We're currently using relative paths for client-side components, which is a pain to update when refactoring.

We should use Webpack's resolver to create top level paths, so we don't have to rely on relative paths to commonly used components, such as `ui/` and `global/`

See [here](https://github.com/elifesciences/elife-xpub/blob/develop/app/components/pages/SubmissionWizard/steps/Editors/EditorsPage.js#L8-L12) for an example of some relative pathing we want to avoid.

SemraSh:
#### Background

The first step of refactoring, using ui and global as modules within the app.

#### Any relevant tickets

Closes #1452


diversemix (approved):
Looks good to me

diversemix:
Let's merge in the morning... as this is the first after release @hdrury

.eslintrc
```diff
@@ -3,34 +3,49 @@
...
   "parser": "babel-eslint",
   "rules": {
     "global-require": 0,
-    "class-methods-use-this": ["error", {
-       "exceptMethods": ["storage",
-        "render",
-        "componentDidMount",
-        "shouldComponentUpdate"]
-     }],
+    "import/no-unresolved": "off",
```
Cooryd, line 17: Is there any way to whitelist this rule? I feel we'll be losing some useful checks by disabling it globally just to allow aliases for the ui components. Maybe we can look into creating `package.json` files for the ui molecules similar to how we're using the `xpub-server` and `xpub-model` in the server?

diversemix: Fair point ... I did scratch my head over this when reviewing ... glad you picked it up @Cooryd

hem-brahmbhatt: I think this actually can be whitelisted, it was initially put in as an attempt to achieve the changes we wanted but then we found another way by specifying the actual resolver.

SemraSh: Creating a pr to fix this.

PR #1474, opened 2019-02-07
1467 - Move logic for uploading manuscript out of the resolver and into the controller

### LINKED ISSUE #1467: FileStatus: Move uploadManuscript into controller

Cooryd:
#### Background

Moves the logic around uploading a manuscript from the resolver layer found in `xpub-resolver` to the controller layer `xpub-controller`. This also makes services such as the `scienceBeamApi` and the `pubsub` a dependency of the manuscript resolver so that they can be be mocked and injected when further controller tests are written as part of #1253

#### Any relevant tickets

Closes #1467 

No new functionality, just a code move. Further testing of the controller classes with be carried out as part of #1253.

#### UI Changes

Server changes only


server/xpub-controller/manuscript.js
```diff
@@ -1,17 +1,192 @@
...
-  constructor(config, user) {
+  constructor(config, user, storage, scienceBeamApi, pubsubManager) {
     this.user = user
     this.config = config
+    this.storage = storage
+    this.scienceBeamApi = scienceBeamApi
+    this.pubsubManager = pubsubManager
+  }
+
+  async upload(manuscriptId, file, fileSize, userId) {
```
Cooryd: So currently the client requires the `ws` message to be `${ON_UPLOAD_PROGRESS}.${userId}` as seen on line 55. I don't really like having to pass `userId` down as a param as we already have `this.user` available BUT these are two differently derived values as seen in https://github.com/elifesciences/elife-xpub/blob/eb62007acfced2a188bfa72466316d66ce5b20bf/server/xpub-model/entities/user/index.js#L48

I don't think the client has access to this so it's not possible to change these around 👎 any idea's on how we might be able to escape the need to pass this?

diversemix: Maybe there should be a check before calling this that `manuscript.user == userId` ?

diversemix:
I'm taking a look at this

server/xpub-controller/manuscript.js
hem-brahmbhatt: It looks like we should be able to pass the entire `user` when instantiating the manuscript, and then call `User.getUuidForProfile` within the controller. It does mean that the resolver is basically empty, and delegating everything to the controller, which is defeating the purpose of having this separation...

The only place we need to look at the full user, is this section: 
```
    // ensure user can view manuscript
    let manuscript = await ManuscriptModel.find(manuscriptId, this.user)
```

We can leave this in the resolver, which means we have resolvers authorizing the call, and the controller doing the processing.

@diversemix what do you think?

server/xpub-controller/manuscript.js
```diff
@@ -1,17 +1,195 @@
...
     this.config = config
+    this.storage = storage
+    this.scienceBeamApi = scienceBeamApi
+    this.pubsubManager = pubsubManager
+  }
+
+  async upload(manuscriptId, file, fileSize) {
+    const { ON_UPLOAD_PROGRESS } = this.pubsubManager.asyncIterators
+
+    const userId = await UserModel.getProfileForUuid(this.user)
```
hem-brahmbhatt: This means another call to the database, which I'm not adverse to.

What's your opinion on https://github.com/elifesciences/elife-xpub/pull/1474#discussion_r255058801

diversemix:
@damnhipster ... sorry just picked up your comments after making some changes. The idea looks good about passing in the user to the constructor of manuscript. Not sure how wide the reprocussions of these changes would be tho? Do you have an idea?

server/xpub-controller/manuscript.js
diversemix: Let's leave architectural conversations till next week. I would argue that a `find` function is not the way we should be checking permissions.

hem-brahmbhatt: Semra and I went over this and we came to the same conclusion.

server/xpub-controller/manuscript.js
```diff
@@ -1,17 +1,195 @@
...
+      `Manuscript Upload Time, Actual (${actualTime}) , Predicted (${predictedTime}) | ${manuscriptId}`,
+    )
+
+    return ManuscriptModel.find(manuscriptId, this.user)
   }
 
   async update(data) {
-    const userUuid = await UserModel.getUuidForProfile(this.user)
-    const manuscript = await ManuscriptModel.find(data.id, userUuid)
+    const manuscript = await ManuscriptModel.find(data.id, this.user)
```
hem-brahmbhatt: Why did this have to change? I thought we were just moving `uploadManuscript` into the controller...

If we have to change `this.user` to now be the `userId`, we should rename the variable.

server/xpub-server/entities/manuscript/resolvers/uploadManuscript.js
```diff
@@ -1,181 +1,19 @@
...
 const pubsubManager = require('pubsweet-server/src/graphql/pubsub')
-const logger = require('@pubsweet/logger')
-const scienceBeamApi = require('../../../../xpub-controller/client/scienceBeamApi')
-
-const { ON_UPLOAD_PROGRESS } = pubsubManager.asyncIterators
-
-const { Manuscript, User } = require('@elifesciences/xpub-model')
-const { File } = require('@elifesciences/xpub-model')
-const { S3Storage } = require('@elifesciences/xpub-controller')
+const ScienceBeamApi = require('../../../../xpub-controller/client/scienceBeamApi')
```
hem-brahmbhatt: This path should be `@elifesciences/xpub-controller`, but the API hasn't been exported yet. We should do this in a separate PR.

hem-brahmbhatt: Addressed by https://github.com/elifesciences/elife-xpub/pull/1485

server/xpub-controller/manuscript.js
hem-brahmbhatt: @diversemix if you're happy with this, I'll make the change

diversemix: @damnhipster ... let me know what you think!

hem-brahmbhatt (approved):

server/xpub-controller/manuscript.js
```diff
@@ -1,17 +1,195 @@
...
     this.config = config
+    this.storage = storage
+    this.scienceBeamApi = scienceBeamApi
+    this.pubsubManager = pubsubManager
+  }
+
+  async upload(manuscriptId, file, fileSize) {
+    const { ON_UPLOAD_PROGRESS } = this.pubsubManager.asyncIterators
+
+    const profileId = await UserModel.getProfileForUuid(this.userId)
```
diversemix, line 22: Add comment

diversemix (approved):
Added an explanitory comment

SemraSh (approved):

server/xpub-controller/manuscript.js
```diff
@@ -17,6 +17,8 @@ class Manuscript {
   async upload(manuscriptId, file, fileSize) {
     const { ON_UPLOAD_PROGRESS } = this.pubsubManager.asyncIterators
 
+    // This is necessary as the config of the pubsub connection relies on the
+    // profileId being used. Should be revisted as part of #1493
```
SemraSh: *revisited* :)

Issue #1500, opened 2019-02-11
Upgrade testcafe to 1.0.0


hem-brahmbhatt:
- Upgrade testcafe to [version 1.0.0](https://github.com/DevExpress/testcafe/releases/tag/v1.0.0).
- Remove the lodash resolution in `package.json`, as the upgraded testcafe now relies on a newer version of lodash, so the resolution introduced in #1501 is no longer needed.

PR #1501, opened 2019-02-11
Locked down the version of lodash to 4.17.11

### LINKED ISSUE #1488: Update lodash
two repositories are affected by the new lodash vulnerability:
https://nvd.nist.gov/vuln/detail/CVE-2018-16487

Fastly-iiif
xpub-responsive-proto

See #4739

hem-brahmbhatt:
#### Background

Lodash 4.17.10 contains a vulnerability as described in #1488

Most of our dependencies don't rely on this specific version and accept newer versions of lodash, so a simple update should suffice.

Unfortunately, we have the following dependency chain within `elife-xpub`

`testcafe` --> `testcafe-hammerhead` --> `lodash@4.17.10`

This means that `lodash@4.17.10` gets used as a common denominator for a bunch of dependencies which depend on lodash.

Upgrading `testcafe` or `testcafe-hammerhead` would fix the issue, but it would also [introduce breaking changes](https://github.com/DevExpress/testcafe/releases/tag/v1.0.0), so our current option is to lock down the version of lodash instead.

This change uses [resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) to force a specific version of lodash for all dependencies within the chain.

#1500 has been created to upgrade testcafe going forward.

#### Any relevant tickets

Closes #1488

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,9 @@
     "client/*",
     "server/*"
   ],
+  "resolutions": {
+    "lodash": ">=4.17.11"
+  },
   "dependencies": {
     "@elifesciences/elife-theme": "^1.0.0",
     "@pubsweet/ui": "^9.0.3",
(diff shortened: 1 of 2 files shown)


SemraSh (approved):

jyenes (approved):
LGTM

Cooryd (approved):

PR #1504, opened 2019-02-11
1487 - upload manuscript subscription key

### LINKED ISSUE #1487: Quick fix -  Stuck at 99% uploading
Issue came in via user: Date: Fri, 8 Feb 2019, 22:56

Hello Susanna,

... We are trying to submit a manuscript to eLife, and we received an email to use the new interface. However, while we were trying to upload our manuscript, we were stuck on the 99% uploading. Then we tried to delete the submission request and create a new submission, the manuscript uploading text box is stuck on 99% even though we haven't upload anything yet. Please see attached screenshot.

---

As a quick fix for this the progress connection created when a user starts a manuscript should be specific to that manuscript, not the user. This should allow the user to create a new submission without displaying the progress % of the stuck manuscript.

Cooryd:
#### Background

Currently the `pubsub` subscription that gets created when uploading a manuscript to track it's upload progress uses the key `${ON_UPLOAD_PROGRESS}.${userId}` as an identifier for its messages. This causes a bug when creating two manuscripts and uploading a file to both as the client has no way to differentiate between the progress for each file. 

A more serious bug was found when an upload fails causing the progress message to stick at 99%. When this is the case, even creating a new manuscript shows the upload in progress view in the wizard preventing the user from uploading another file.

This PR replaces the `userId` of the subscription key with the `manuscriptId` so ewach manuscript can have its own message key.

#### Any relevant tickets

Closes #1487

### DIFF
--- a/server/xpub-server/entities/manuscript/resolvers/index.js
+++ b/server/xpub-server/entities/manuscript/resolvers/index.js
@@ -2,6 +2,10 @@ const logger = require('@pubsweet/logger')
 const { Manuscript, User, File } = require('@elifesciences/xpub-model')
 const elifeApi = require('@elifesciences/xpub-model/entities/user/helpers/elife-api')
 const { S3Storage } = require('@elifesciences/xpub-controller')
+const {
+  getPubsub,
+  asyncIterators,
+} = require('pubsweet-server/src/graphql/pubsub')
 
 const removeUploadedManuscript = require('./removeUploadedManuscript')
 const submitManuscript = require('./submitManuscript')
@@ -9,6 +13,8 @@ const updateManuscript = require('./updateManuscript')
 const uploadManuscript = require('./uploadManuscript')
 const uploadSupportingFile = require('./uploadSupportingFile')
 const removeSupportingFiles = require('./removeSupportingFiles')
+
+const { ON_UPLOAD_PROGRESS } = asyncIterators
 
 const resolvers = {
   Query: {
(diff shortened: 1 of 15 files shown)


Cooryd:
Will need to pull and fix conflicts as the server logic now sits in `xpub-controller`

hem-brahmbhatt (approved):

server/xpub-server/entities/manuscript/resolvers/index.js
```diff
@@ -80,6 +86,15 @@ const resolvers = {
     },
   },
 
+  Subscription: {
+    manuscriptUploadProgress: {
+      subscribe: async (_, vars, context) => {
+        const pubsub = await getPubsub()
+        return pubsub.asyncIterator(`${ON_UPLOAD_PROGRESS}.${vars.id}`)
```
Cooryd, line 93: For context: This is where we're changing the subscriptions 'listener' key to use the manuscript id. Heres the code from `pubsweet`s version of the `Subscription` resolvers.

`node_modules/pubsweet-server/src/graphql/definitions/upload.js`

```
 Subscription: {
    uploadProgress: {
      subscribe: async (_, vars, context) => {
        const pubsub = await getPubsub()
        return pubsub.asyncIterator(`${ON_UPLOAD_PROGRESS}.${context.user}`)
      },
    },
  },
```

Unfortunately I couldn't find any way to override graphql resolvers so I couldn't redefine `uploadProgress`, instead I've had to create a new resolver called `manuscriptUploadProgress` which is then added to the server and clients graphql  schemas

jyenes (approved):
the code looks good, I was testing it on the review environment, and the spinner most of the time jumps directly in the 100%,  or behaves like this 1,2, 3, 4, 5, 6, 7, 100

I think is ok at least I notice is not getting stuck

Cooryd:
This is probably due to file size and upload speed. As long as we're seeing something other than 0% it means the connection is correctly working. Best way to also check is to look into the `ws` tabs of the network inspector and have a look at the frames 

screenshot 2019-02-13 at 10 43 44

PR #1509, opened 2019-02-13
Fix blank disclosure

### LINKED ISSUE #1507: Fix blank disclosure.pdf
The MECA package is exporting blank disclosure pdf(s)

- Add tests 
- Then fix

diversemix:
#### Background

Prevents the conversion of the Buffer to a string which corrupts the PDF

#### Any relevant tickets

Closes #1507 

New Test added to properly test the disclosure PDF

### DIFF
--- a/server/meca/file-generators/disclosure.test.js
+++ b/server/meca/file-generators/disclosure.test.js
@@ -1,36 +1,41 @@
 const PDFParser = require('pdf2json')
 const generatePdf = require('./disclosure')
-const fs = require('fs')
 
 describe('Disclosure PDF generator', () => {
   it('returns a string of a PDF', async () => {
     const document = await generatePdf({
       meta: {},
-      teams: [{ role: 'author', teamMembers: [{ alias: {} }] }],
-    })
-
-    // save the document
-    //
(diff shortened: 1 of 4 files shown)


server/meca/file-generators/disclosure.test.js
```diff
@@ -1,12 +1,41 @@
...
     const document = await generatePdf({
       meta: {},
-      teams: [{ role: 'author', teamMembers: [{ alias: {} }] }],
+      teams: [
+        {
+          role: 'author',
+          teamMembers: [
+            {
+              alias: {
+                givenNames: 'J. Edward',
```
Cooryd, line 14: We could turn these values into variables so that the assertions can't become out of sync with the pdf values in future?

Cooryd (approved):
Looks good, just left a suggestion

Edit: Suggestion was going to be more complicated than it looked. I'm happy with this as is.

server/meca/file-generators/disclosure.test.js
diversemix: @Cooryd I'm happy you making these changes

jyenes (approved):
LGTM

PR #1515, opened 2019-02-13
1511 fix prod disclosure

diversemix:
#### Background

Adds a script to regenerate pdf

#### Any relevant tickets

Does not fully close any ticket

locally - its a script not part of production

### DIFF
--- a/scripts/generateDisclosure.js
+++ b/scripts/generateDisclosure.js
@@ -1,0 +1,49 @@
+#!/usr/bin/env node
+const fs = require('fs')
+const generateDisclosurePdf = require('../server/meca/file-generators/disclosure')
+const { Manuscript } = require('@elifesciences/xpub-model')
+const { db } = require('pubsweet-server')
+
+async function regenerate(id) {
+  console.log(id)
+
+  const [manuscript] = await Manuscript.query().where({
+    'manuscript.id': id,
+  })
+
+  let doc = null
+  try {
+    await manuscript.$loadRelated('[teams]')
+    doc = await generateDisclosurePdf(manuscript, 'no-ip-address')
+
+    const filename = `disclosure-${id}.pdf`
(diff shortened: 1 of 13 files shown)


server/meca/file-generators/disclosure.test.js
```diff
@@ -1,25 +1,10 @@
...
-              alias: {
-                givenNames: 'J. Edward',
-                surname: 'XXYYZZ',
-              },
-            },
-          ],
-        },
-      ],
-    })
+    const document = await generatePdf(manuscript)
```
hem-brahmbhatt, line 7: Yeah, this is better 👍

scripts/generateDisclosure.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const id = process.argv[2]
+
+if (id.length !== 36) {
+  console.error('That does not look like a manuscript id')
+  process.exit(1)
+}
+
+regenerate(id)
```
hem-brahmbhatt, line 49: What's the need for this script?

hem-brahmbhatt (approved):

Cooryd (approved):

scripts/generateDisclosure.js
hem-brahmbhatt: I assume this script is to be executed on a production instance. Is this something that's been tested?

hem-brahmbhatt: If not, we can :shipit: and test it afterwards.

Cooryd: It's a manual running of the generation process to create a disclosure pdf for those that are missing it because of the bug. I agree this should have a test written for it to make sure it's doing what we assume it should.

diversemix: Every regenerated pdf will be eyeballed anyway so don't think tests are needed.

Issue #1492, opened 2019-02-11
Breakdown and fix the server `upload` function


diversemix:
See #1487 

When an error occurs in the is function the final graphQL subscription message should always be sent. Currently we are not handling these errors correctly and so the final message is not being sent in this case.

Adding the necessary error handling will turn this huge function into a mother-of-all-sperm-whales of a function... 

... so it would be good to split it up so we can see where error handling should be done and wrapping the pubsub management around this.

/cc @SemraSh  ... this ok?

PR #1516, opened 2019-02-13
WIP: reordering the checks, to check errors on first instance

jyenes:
moving up the check for errors, so this is the first check it doesn't matter if the upload is still in progress


Cooryd (changes requested):
https://github.com/elifesciences/elife-xpub/blob/54f1f44c764f8eb121da7418b237644a3dac7b2c/app/components/pages/SubmissionWizard/steps/Files/ManuscriptUpload.js#L93

This PR will introduce the same graphical bug as this previous PR if i'm understanding it correctly.

jyenes:
closing this one as all has been sorted out on the backend 
see
#1492, #1517, #1535

PR #1517, opened 2019-02-13
1492 refactor manuscript upload

### LINKED ISSUE #1492: Breakdown and fix the server `upload` function

SemraSh:
#### Background
Refactors the update function of Manuscript to make it easier to debug.

#### Any relevant tickets
Closes #1492

- [x] Passing the current tests


server/meca/routes.test.js
```diff
@@ -93,7 +93,7 @@ describe('MECA HTTP callback handler', () => {
     await mailer._sendPromise
     expect(mailer.getMails()).toHaveLength(1)
     expect(mailer.getMails()[0]).toMatchObject({
-      text: expect.stringContaining(manuscript.id)
+      text: expect.stringContaining(manuscript.id),
```
Cooryd, line 96: This looks like a linting change has snuck in. Is this from the pre-commit hook linting or something running locally?

SemraSh: Yes it was added when I committed, looking to how to get rid of that setting.

server/xpub-controller/manuscript.js
```diff
@@ -99,88 +82,44 @@ class Manuscript {
...
       } | ${manuscriptId}`,
     )
 
-    clearInterval(handle)
+    endFileProgress(progress)
     pubsub.publish(`${ON_UPLOAD_PROGRESS}.${manuscriptId}`, {
       manuscriptUploadProgress: 100,
     })
+
+    // -->
```
Cooryd: Comments can be removed

server/xpub-controller/manuscript.js
```diff
@@ -46,37 +36,21 @@ class Manuscript {
...
-      })
-    }, 200)
+    const progress = FilesHelper.startFileProgress(
+      pubsub,
+      ON_UPLOAD_PROGRESS,
+      fileSize,
+      startedTime,
+      predictedTime,
+      manuscriptId,
+    )
```
Cooryd, line 46: Could everything from this line (47) to where the manuscript has been successfully uploaded on line 99 be factored out into a new Manuscript Helper class which sits along side the File helper class. This class would then have a single function called uploadManuscript. This would split the process of Tracking the progress of an upload apart from the actual uploading of the file. This would then make it easier for us to wrap the whole uploading process in a try catch to clear the progress interval if something goes wrong. 

To elaborate further some sudo-code:
```
upload() {
  FileHelper.checkFileSize();
  FileHelper.startFileProgress();
  try {
    await ManuscriptHelper.uploadManuscriptFile();
  } catch (error) {
    FileHelper.endFileProgress()
    throw error
  }
  FileHelper.endFileProgress()
  return manuscript
}
```

server/xpub-controller/manuscript.js
```diff
@@ -99,84 +73,36 @@ class Manuscript {
...
 
     logger.info(
       `Manuscript Upload Manuscript::saved ${
         manuscript.meta.title
       } | ${manuscriptId}`,
     )
 
-    clearInterval(handle)
+    FilesHelper.endFileProgress(progress)
     pubsub.publish(`${ON_UPLOAD_PROGRESS}.${manuscriptId}`, {
```
Cooryd: Could `endFileProgress` be mixed with the publish 100%? I don't think we should be calling one without the other anywhere.

server/xpub-controller/manuscript.js
jyenes: agreed

Cooryd: Maybe leave out the try catch added for #1521 to take care of so this remains just a refactor task

jyenes: makes sense

server/xpub-controller/helpers/manuscript.js
```diff
@@ -0,0 +1,79 @@
...
+        await this.storage.putContent(fileEntity, fileContent, {
+          size: fileSize,
+        })
+        await fileEntity.updateStatus('STORED')
+      } catch (err) {
+        logger.error(
+          `Manuscript was not uploaded to S3: ${err} | ${manuscriptId}`,
+        )
+        await fileEntity.delete()
+        FilesHelper.endFileProgress(progress)
```
Cooryd: Just a note for #1521 this should probably be removed from this helper function (which would mean no longer requiring progress as a param). The reject will then be handled at the controller level to end file progress.

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,151 @@
+const logger = require('@pubsweet/logger')
+const FileModel = require('@elifesciences/xpub-model').File
+
+class FilesHelper {
```
diversemix, line 4: Maybe worth a comment on the class that although this is a generic "FilesHelper" it is very specific to MANUSCRIPT_SOURCE ... in the future it would be good to handle the supplementary files too.

diversemix (approved):
Looks lovely - also agree with Cory's suggestion of moving pubsub into the FileHelper.

Cooryd (approved):
Great work, this is much more readable.

Issue #1462, opened 2019-02-07
Tests on develop branch to fail if coverage decreases


diversemix:
We are going to go through a stage of re-factoring and we should at no time decrease the code coverage. Make sure we cannot commit code that reduces coverage.

- In order to achieve this we will store the overall test coverage in a file within the repo.
- This will get updated when tests are run on the develop branch if the coverage increases (or the same)
- If the coverage decreases this should result in the tests failing

diversemix:
On second thoughts this should be perhaps done on merging to develop

Cooryd:
Agreed, per commit might make it harder to get work into a branch for visibility. Could become more of a hinderance than anything else. Could this be checked in CI

diversemix:
@jyenes / @damnhipster / @Cooryd 

Apologies for the not-so-well-thought-out ticket ... I've updated this ticket (totally my bad)

hem-brahmbhatt:
I think it's useful to have this when running against PRs, rather than just on develop, where we'll then need to create another PR to fix develop. Having this run at the end of the tests should give us enough visibility and the chance to fix the coverage while giving us the opportunity to commit unfinished pieces of work.

PR #1518, opened 2019-02-13
add pre-commit hook to test the coverage

jyenes:
adds a pre-commit hook, that avoids performing any commit if the coverage gets down or the tests are broken

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,14 @@
       "<rootDir>/{client,app,server}/**/*.js",
       "!**/*.{config,test}.js"
     ],
+    "coverageThreshold": {
+      "global": {
+        "branches": 60,
+        "functions": 61,
+        "lines": 65,
+        "statements": -1000
+      }
+    },
     "projects": [
       "<rootDir>/client",
       "<rootDir>/app",
@@ -157,6 +165,7 @@
     "lint-staged": "^7.2.2",
     "mime": "^2.3.1",
     "pdf2json": "^1.1.8",
+    "pre-commit": "^1.2.2",
     "prettier": "^1.8.2",
     "pubsweet": "^3.0.13",
     "react-styleguidist": "^7.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -171,5 +180,8 @@
     "webpack": "^3.8.1",
     "webpack-dev-middleware": "^3.1.3",
     "webpack-hot-middleware": "^2.22.1"
-  }
+  },
+  "pre-commit": [
+    "test"
+  ]
```
hem-brahmbhatt: Not sure about running tests before every commit... I can see that getting in the way while developing. It should be enough if it fails the build on CI.

diversemix: I agree... I've commented here https://github.com/elifesciences/elife-xpub/issues/1462#issuecomment-463545996

jyenes: we can do it in pre-push, but I think is necessary before to upload new code, to run the tests and be sure there are no tests broken

package.json
```diff
@@ -64,6 +64,14 @@
       "<rootDir>/{client,app,server}/**/*.js",
       "!**/*.{config,test}.js"
     ],
+    "coverageThreshold": {
```
diversemix, line 67: What does this achieve?

jyenes: this is the way to configure jest, with a threshold, so if the coverage goes down to this threshold the test will fail

package.json
jyenes: removed pre-commit, with this config, now the test will be executed on the CI, and will fail if the coverage decrease, from the one defined in the threshold, which matches with our current coverage.
the idea behind this is to update the threshold config periodically, so will ensure the coverage increases

jyenes:
removed pre-commit, with this config, now the test will be executed on the CI, and will fail if the coverage decrease, from the one defined in the threshold, which matches with our current coverage.
the idea behind this is to update the threshold config periodically, so will ensure the coverage increases

package.json
diversemix: ok - so how does this achieve not being able to decrease coverage? Do we have to continually update this ?

hem-brahmbhatt: Yes

hem-brahmbhatt: We can automate the coverage threshold in a separate issue.

hem-brahmbhatt (approved):

diversemix (approved):

Issue #1519, opened 2019-02-14
Test email alert when MECA import fails


hem-brahmbhatt:
Test the results of #1496.

_Can this be tested in Staging?_
Yes, we have the outgoing mail server setup in Staging.

_How can we simulate a MECA import failure?_
Request the callback URL with a failure state.

hem-brahmbhatt:
I've been unable to test in Production and Staging so far. In both environments, I get the following error:

_In Staging:_
```
elife@staging--xpub:~$ docker exec -it xpub_postgres_1 psql --user xpub
psql (10.5 (Debian 10.5-1.pgdg90+1))
Type "help" for help.

xpub=# select id from manuscript;
                  id
--------------------------------------
 6a6460f5-83e7-471d-a331-813844d95fbe
 e808c513-63bf-4ae8-b873-2117840ff372
(2 rows)

root@a0adac2b7529:~# ./scripts/send-meca-result.sh 6a6460f5-83e7-471d-a331-813844d95fbe
{"error":"Manuscript not found"}
```

I've executed the script directly on Staging to rule out any access issues. I get the same error when executing the script from my local machine.

hem-brahmbhatt:
I've been unable to reproduce the issue locally. Instead I get the following error:

```
hem@hems-macbook ~/Projects/elife/elife-xpub (develop) $ docker exec -it elife-xpub_postgres_1 psql --user hem
psql (10.4 (Debian 10.4-2.pgdg90+1))
Type "help" for help.

hem=# select id from manuscript;
                  id
--------------------------------------
 dab686dd-b0d2-4a89-93fa-127666784112
(1 row)

hem@hems-macbook ~/Projects/elife/elife-xpub (develop) $ MECA_API_KEY=abcd1234 ./scripts/send-meca-result.sh dab686dd-b0d2-4a89-93fa-127666784112
```
_Logs from local server:_
```
info: MECA callback received manuscriptId=dab686dd-b0d2-4a89-93fa-127666784112, result=failure, fixedby=elife-developer
info: Saved AuditLog with UUID a5025f25-989d-4b16-a9d2-02a0d9c23171
info: POST /meca-result/dab686dd-b0d2-4a89-93fa-127666784112 204 - - 200.144 ms

error: Failed to process MECA callback manuscriptId=dab686dd-b0d2-4a89-93fa-127666784112, error=No recipients defined
[...]

PR #1523, opened 2019-02-14
1519 doesn't send an error response when email fails

hem-brahmbhatt:
#### Background

We currently have an unhandled promise in the MECA callback. We return a successful response back, but if the subsequent email fails, we attempt to send a second response back even though we've already sent a success. This gives us the unhandled promise error.

#### Any relevant tickets

Addresses https://github.com/elifesciences/elife-xpub/issues/1519#issuecomment-463616414

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/server/meca/routes.js
+++ b/server/meca/routes.js
@@ -37,7 +37,14 @@ module.exports = app => {
 
     Manuscript.updateStatus(manuscriptId, status)
       .then(() => res.sendStatus(204))
-      .then(() => {
+      .catch(err => {
+        logger.error('Failed to process MECA callback', {
+          manuscriptId,
+          error: err.message,
+        })
+        res.status(500).send({ error: err.message })
+      })
+      .finally(() => {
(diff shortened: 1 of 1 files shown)


server/meca/routes.js
```diff
@@ -37,7 +37,14 @@ module.exports = app => {
...
+      .catch(err => {
+        logger.error('Failed to process MECA callback', {
+          manuscriptId,
+          error: err.message,
+        })
+        res.status(500).send({ error: err.message })
+      })
+      .finally(() => {
         if (status === Manuscript.statuses.MECA_IMPORT_FAILED) {
           return mailer.send({
```
Cooryd, line 49: Could this be placed in the `catch` after line 45?

Cooryd: My miss-understanding this is a catch for updating the status of the manuscript. This makes sense.

Cooryd (approved):

diversemix (approved):
Took a while for me to get my head round this - looks great :+1:

PR #1529, opened 2019-02-15
1519 fixing Promise error for MECA import emails

hem-brahmbhatt:
#### Background

#1523 introduced a silent error when calling the MECA callback route due to Promises with handles being left open (which is my best guess). Unfortunately this didn't fail the unit tests.

The impact of this change is that we now wait for the email to finish sending before we send back a response. This means that the request currently takes up to 5 seconds, from the results when testing locally.

The output is currently:
```
root@b0a8f28ed4b0:~# ./scripts/send-meca-result.sh 6a6460f5-83e7-471d-a331-813844d95fbe
{"message":"Manuscript.updateStatus(...).then(...).catch(...).finally is not a function"}
```

With this fix, the output should now be empty, since we return a status of 204 with an empty body.

#### Any relevant tickets

Relates to #1519 

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/server/meca/routes.js
+++ b/server/meca/routes.js
@@ -35,32 +35,20 @@ module.exports = app => {
         ? Manuscript.statuses.MECA_IMPORT_SUCCEEDED
         : Manuscript.statuses.MECA_IMPORT_FAILED
 
-    const updateStatus = () => new Promise((resolve, reject) => {
-      console.log('--- updating manuscript status to ', status)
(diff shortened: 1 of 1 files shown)


Cooryd (approved):

diversemix:
Tests worked
image 2

diversemix (approved):

PR #1530, opened 2019-02-15
Added vim to Dockerfile

hem-brahmbhatt:
#### Background

Adds Vim to Dockerfile. This makes it easier to edit files within environments such as Staging and Production. This is useful when editing scripts within the `/scripts` directory, since a lot of these are hardcoded to work only within specific environments.

For example, https://github.com/elifesciences/elife-xpub/blob/develop/scripts/send-meca-result.sh is hardcoded to request production, even if it's executed from Staging.

At the moment, I manually install vim within the docker container when needed, but this isn't persisted between deployments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR ${HOME}
 RUN curl -sL http://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
 RUN echo 'deb http://dl.google.com/linux/chrome/deb/ stable main' >> /etc/apt/sources.list.d/google.list
 RUN apt-get update \
-    && apt-get install -y google-chrome-stable socat \
+    && apt-get install -y google-chrome-stable socat vim \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 


diversemix:
Why are you running these scripts from within the container?
I would rather leave the container as is - and where necessary run scripts from the dev's machine - tunnelling ports where necessary.

Open to a discussion on the topic...

hem-brahmbhatt:
That's fine by me. I was running scripts from environments to avoid tunneling and copying over secrets to my local machine, but I can see that it's dangerous running these scripts directly on the environments.

Do we have any instructions in the docs on running these scripts? If not, I can create an issue for it.

hem-brahmbhatt:
@giorgiosironi would be good to get your opinion on this PR.

giorgiosironi:
Editing code directly on `prod` or `staging` environments sounds like a smell since it may affect the running application and/or run untested code. If the scripts are not flexible enough, we extend them so that they can be reused. I'd rather not expand the container image which already contains Google Chrome and similar, adding time to every deployment.

I can see that scripts may have to be run in a real environment rather than locally because:

- as you discovered, secrets should not be copied and accidentally forgotten on laptops
- there may be firewalls stopping you from e.g. accessing the RDS database

diversemix:
@BlueReZZ / @giorgiosironi ... Do we have a "elife" process for handing such situations? At the moment I have suggested using `~/.xpub_secrets` with the user tool (see tools/user-tool/USER-TOOL.md )

This promotes using secrets on your own machine for accessing staging/prod environment. I also am using these to query the prod database for ongoing stats.

I think it would be good to have a unified approach.

hem-brahmbhatt:
Closing the pull request, happy for conversation to continue in this issue.

I avoid copying secrets to my machine and try to use whatever tools the host has to do my investigations and testing. I have it in mind to create a way to secure secrets on my local machine, but at the moment, I just rely on my encrypted hard drive.

giorgiosironi:
> This promotes using secrets on your own machine for accessing staging/prod environment. I also am using these to query the prod database for ongoing stats.

RDS is a bit excluded by this due to its credentials being generated by AWS and only stored on the servers, but as for credentials in general you could read them with `vault kv get` once it's setup locally.

PR #1532, opened 2019-02-15
Bug fix file progress

SemraSh:
Fixes the wrong identity of endFileProgress happened during the refactoring in #1492


jyenes (commented):
we are closing this PR, a new one will be created with the original code, plus the fix.

PR #1535, opened 2019-02-15
1492 bug fix

SemraSh:
fixes the bigs introduced during the refactor on #1492


hem-brahmbhatt:
Closed in favour of #1536

Issue #1521, opened 2019-02-14
Clear the interval set to send upload progress when a process fails on manuscript upload


Cooryd:
## Subtasks

As part of #1492 the `upload` function on the `manuscript` controller was broken down into smaller helper functions for each stage of the upload process. One of the reasons for doing this is there is currently a bug in the code that causes the app to continue sending the client 99% progress messages when one of the processes fails but the interval being used to send progress messages is not cleared.

The refactoring should have made it easier to wrap tasks that can throw an Error in a `try..catch` and ensure the interval is cleared when that error occurs. 

## Assumptions and notes

- The most likely culprit for the 99% bug is at the scienceBeam extraction stage as this is wrapped in a try catch already but doesn't clear the interval. 

- ~~Only those functions that throw an error or end the request in some way need to be wrapped~~

- The simplest solution would be to wrap the entire process after calling `startFileProgress`  in a try catch that then call the following when an error is thrown.

```
    } catch (error) {
      endFileProgress(progress)
      pubsub.publish(`${ON_UPLOAD_PROGRESS}.${manuscriptId}`, {
        manuscriptUploadProgress: 100,
      })
      throw error
    }
```

Issue #1520, opened 2019-02-14
Write tests for controller helper functions


Cooryd:
## Background

As part of #1492 the `manuscript` controllers `upload` function was simplified by splitting out the functionality into smaller steps and placing these in helper functions. One of the aims of this re-factoring task was to make the individual pieces of functionality easier to test as they now sit in the stand alone functions rather than the controller classes function.

This task involves writing the tests for these helper functions that have been pulled out of the `manuscript` controller. 

## Assumptions and notes

It may be worth looking through existing tests for the `manuscript` controller function and `uploadManuscript` resolver function as they may already contain code for testing some of the functionality of these new helper functions and could be reused / moved as appropriate.

SemraSh:
This ticket will be closed after testing part of the functions in xbup-controller helpers

```
validateFileSize
generateFileEntity
publishPredictedProgress // tested in a previous pr
endFileProgress
```

the rest of the functions are tested in these tickets --> #1554 and #1555

PR #1536, opened 2019-02-15
1538 reintroduce refactor, along with fix

### LINKED ISSUE #1538: Fix the server `upload` function
This was originally done as part of #1492, but had to be reverted due to a bug related with the upload progress, which was captured in #1537.

We should reintroduce the fix, and fix the upload progress issue that had been introduced as part of the original fix.

- [x] re-introduce #1492 code
- [x] fix the bug
- [ ] create test that covers the failure

hem-brahmbhatt:
Fixes the bigs introduced during the refactor on #1492

Closes #1538


Cooryd (approved):
Works as it did before

diversemix:
Could someone identify the fix and the tests that cover this now pls? :)

server/xpub-controller/manuscript.js
```diff
@@ -11,172 +11,54 @@ class Manuscript {
...
 
-    manuscript.files[pendingFileIndex].type = 'MANUSCRIPT_SOURCE'
-
-    logger.info(`Manuscript Upload Manuscript::save ${title} | ${manuscriptId}`)
-    manuscript.meta.title = title
-    await manuscript.save()
-
-    const sourceList = manuscript.files.filter(
-      f => f.type === 'MANUSCRIPT_SOURCE',
+    await this.manuscriptHelper.uploadManuscriptFile(
```
diversemix, line 49: What happens if this function throws ? ... this should be wrapped in a try-catch

jyenes: this will be introduced as part of the #1521

jyenes:
> Could someone identify the fix and the tests that cover this now pls? :)

tests will be done as part as #1520

diversemix:
I'm sort of reluctant to let this go back into production without tests /cc @hdrury1

jyenes (approved):

Cooryd:
Compromise, maybe we try to write a test for the `startFileProgress` function this morning (@jyenes , @SemraSh  and myself can pair on this). This'll add coverage for the breaking issue in the previous merge

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  static startFileProgress(
+    pubsub,
+    ON_UPLOAD_PROGRESS,
+    startedTime,
+    predictedTime,
+    manuscriptId,
+  ) {
+    return setInterval(() => {
```
Cooryd: To make this function easier to test, lets remove the `setInterval` at this level and re-name this something like `publishPredictedProgress` ? The `setInterval` could then be called at the controller level.

jyenes (dismissed):
include the test suggested by @diversemix, so currently following the suggestion of @Cooryd to get this PR in

server/xpub-controller/helpers/files.js
```diff
@@ -33,22 +33,22 @@ class FilesHelper {
...
-  static startFileProgress(
+  static publishPredictedProgress(
     pubsub,
     ON_UPLOAD_PROGRESS,
     startedTime,
     predictedTime,
     manuscriptId,
   ) {
-    return setInterval(() => {
+    return () => {
```
diversemix: Do you think its worth putting some validation on the parameter here.... for example ...
```
preditctedTime > startedTime
manuscriptId.length == 36
typeof pubsub.publish == 'function' 
```
... just a question?

A bit of defensive programming here wouldn't go amiss seeing its the No1 experience for a user at the moment. Then we could check some failure cases too :)

Cooryd: I like this idea. Then we could write some pretty simple tests ensuring an error is thrown when an invalid value is passed

jyenes: predicted time, is not a date, is just the time predicted in seconds, i.e 3.8907 seconds, so this condition is not valid, agree with the other ones.

Cooryd: preditctedTime > 0 would work

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,161 @@
...
+      type: 'MANUSCRIPT_SOURCE_PENDING',
+      mimeType,
+    }).save()
+    return {
+      stream,
+      fileEntity,
+    }
+  }
+
+  static publishPredictedProgress(
```
hem-brahmbhatt, line 36: Is it worth renaming this to `publishPredictedProgressThunk`?

We're expecting to delay the execution of this function, so it may be worth making that clearer.

And if we are using thunks, the validation usually happens within the function, and the wrapping function is left very thin. 

You can read more about thunks [here](https://en.wikipedia.org/wiki/Thunk)

hem-brahmbhatt: Actually, I don't have enough experience with thunks to recommend them here. I'd say avoid them unless someone else wants to give their input on the topic.

server/xpub-controller/manuscript.js
```diff
@@ -11,172 +11,57 @@ class Manuscript {
     this.storage = storage
     this.scienceBeamApi = scienceBeamApi
     this.pubsubManager = pubsubManager
+    this.filesHelper = new FilesHelper(this.config, this.scienceBeamApi)
+    this.manuscriptHelper = new ManuscriptHelper(
+      this.config,
+      this.userId,
+      this.storage,
+      this.filesHelper,
+    )
```
hem-brahmbhatt, line 20: I find it strange that we have classes called `FilesHelper` and `ManuscriptHelper`. A class says to me that we're expecting to have instances of it, for example, `new User()` or `new File()` which are objects we'd expect to keep state in. Classes called helper sound like they shouldn't be keeping state and should be left as pure functions.

It's worth changing this to one of the following **at some point in the future**:
- declare `FilesHelper` as a bunch of functions that get exported
- have a class called `File` which contains the helper functions

server/xpub-controller/manuscript.js
```diff
@@ -11,172 +11,57 @@ class Manuscript {
...
-    }).save()
+    // ensure user can view manuscript
+    const manuscript = await ManuscriptModel.find(manuscriptId, this.userId)
 
-    const fileId = fileEntity.id
-
-    logger.info(
-      `Manuscript Upload Size: ${filename}, ${fileSize} | ${manuscriptId}`,
-    )
+    const fileData = await FilesHelper.generateFileEntity(file, manuscriptId)
```
hem-brahmbhatt, line 31: This should be called `fileEntity` rather than `fileData`, since the function we're calling is `generateFileEntity`.

hem-brahmbhatt: It looks like the functions returns back the data as well. It's worth renaming the function to `getFileData`

server/xpub-controller/helpers/files.js
hem-brahmbhatt: This would've been easier to add in a separate PR. It's making it more difficult to review at the moment given there's new functionality embedded in the refactor.

server/xpub-controller/manuscript.js
SemraSh: @damnhipster I agree your point, but I think Javi thinks the opposite, he suggested making it class will be more declarative as it has the `FileHelper.functionName` pattern. 
I also think this should be consistent in the whole codebase, the other places that use helpers are exporting functions as you suggested. I am happy to have a further discussion on that and I will follow up with this, for now I need to merge this one to be able to add the tests for the rest of the functions.

Thank you for this input.

server/xpub-controller/manuscript.js
SemraSh: `generateFileEntity` returns `{ fileEntity , stream }` not just the fileEntity

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,161 @@
...
+    pubsub,
+    ON_UPLOAD_PROGRESS,
+    startedTime,
+    predictedTime,
+    manuscriptId,
+  ) {
+    if (
+      typeof pubsub.publish === 'function' &&
```
hem-brahmbhatt, line 44: This `typeof` check looks like type checking to me. 

orc mischief

So far we've avoided type checking and I think we should continue avoiding it at this point until we have an agreed upon method of type checking.

The rest of the validation looks 👌

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,161 @@
...
+      logger.info(
+        `Manuscript Upload Manuscript::save ${title} | ${manuscript.id}`,
+      )
+      manuscript.meta.title = title
+      await manuscript.save()
+      resolve(manuscript)
+    })
+  }
+
+  static validateManuscriptSource(manuscript) {
```
hem-brahmbhatt, line 147: This should live in `ManuscriptHelpers`

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,161 @@
...
+          manuscript.files[oldFileIndex].filename
+        } | ${manuscript.id}`,
+      )
+      const oldFile = await FileModel.find(manuscript.files[oldFileIndex].id)
+      manuscript.files.splice(oldFileIndex, 1)
+      await oldFile.delete()
+    }
+  }
+
+  static async setManuscriptMetadata(initialManuscript, title) {
```
hem-brahmbhatt, line 131: This should live in `ManuscriptHelpers`

server/xpub-controller/helpers/files.js
```diff
@@ -0,0 +1,161 @@
...
+      logger.warn('Manuscript conversion failed', {
+        error: errorMessage,
+        manuscriptId,
+        filename,
+      })
+    }
+    return title
+  }
+
+  static async cleanOldManuscript(manuscript) {
```
hem-brahmbhatt, line 114: This should live in `ManuscriptHelpers`

hem-brahmbhatt (commented):
Code looks fine to go in, nothing hugely important that can't be changed later.

Has anyone done any testing in the review environment?

hem-brahmbhatt:
Everything seems to be working fine in the review environment. 

We don't have a sciencebeam container running for review environments so I haven't been able to test out the title extraction. 

But this is covered by the tests on the end2end environment, which does have a running sciencebeam container deployed alongside xpub.

SemraSh (approved):

server/xpub-controller/manuscript.js
hem-brahmbhatt: You could do something like:
```
// FilesHelper.js
module.exports = {
  validateFileSize = function (config, api, ...) { }
}

// controller.js
validateFileSize = FilesHelper.validateFileSize.bind(this.config, this.scienceBeamApi)
validateFileSize(...)
```

PR #1547, opened 2019-02-19
Increase time out for jest before/after hooks

### LINKED ISSUE #1546: Figure out what is wrong with the tests
Putting I  a setTimeout of 200ms into createTables causes all sorts of catastrophes.

This should of course work. I'm figuring this is the same issue as the pipeline due to it being timing in the environment.

diversemix:
#### Background

Prevent the tests from timing out in the pipeline.

#### Any relevant tickets

Closes #1546

This would be a bit "meta" :stuck_out_tongue_closed_eyes:

### DIFF
--- a/server/jest-setup.js
+++ b/server/jest-setup.js
@@ -4,6 +4,7 @@ const replaySetup = require('../test/helpers/replay-setup')
 replaySetup('success')
 
 // The pipeline complains at the default of 5s - so double it!
+// https://jestjs.io/docs/en/jest-object#jestsettimeouttimeout
 jest.setTimeout(10000)
 
 jest.mock('@pubsweet/component-send-email', () => ({
--- a/server/jest-setup.js
+++ b/server/jest-setup.js
@@ -3,6 +3,9 @@ const replaySetup = require('../test/helpers/replay-setup')
 
 replaySetup('success')
 
+// The pipeline complains at the default of 5s - so double it!
+jest.setTimeout(10000)
+
 jest.mock('@pubsweet/component-send-email', () => ({
   _sendPromise: null,
   mails: [],


diversemix:
Third time of running - if it works then I think its good to go

server/jest-setup.js
```diff
@@ -3,6 +3,9 @@ const replaySetup = require('../test/helpers/replay-setup')
 
 replaySetup('success')
 
+// The pipeline complains at the default of 5s - so double it!
```
Cooryd, line 6: Might be useful to comment what is meant by default. or something like `//This increases the interval between the before and after hooks and tests from 5s to 10s.`

Reviewers see:
https://jestjs.io/docs/en/jest-object#jestsettimeouttimeout

diversemix: Ooops - I did plan on putting that link in the comment - let me do that... 

What do you mean by:
> Might be useful to comment what is meant by default

Cooryd (approved):
Solution looks good and makes sense. Is this scalable over time? or will we likely need to increase CI resource / increase this `setTimeout` value as the product and test suites expand?

server/jest-setup.js
Cooryd: As in what the time relates to. I suppose the docs link would be sufficient explanation

diversemix:
My best guess is that its a function of the infrastructure and the number of tests in a given file - given that `afterAll` does a `db.destroy()`

server/jest-setup.js
diversemix: Added the link ... the other I think is a matter of style so will leave the rest of the comment intact

hem-brahmbhatt (approved):

PR #1550, opened 2019-02-20
1521 error catch

### LINKED ISSUE #1521: Clear the interval set to send upload progress when a process fails on manuscript upload

Cooryd:
#### Background

 Currently there is a bug that causes the progress of a manuscript upload to get stuck at 99% if an error is thrown due to the setInterval continuing to broadcast progress messages to the client.

This PR wraps the entire `uploadManuscriptFile` process in a `try catch`, clearing the interval and sending a 100% message to the client so that it knows the process has ended (albeit in an error state). The `uploadManuscriptFile` function originally had this clearing functionality at its scienceBeam extraction stage. This has been removed so that `uploadManuscriptFile` no longer has knowledge of the progress tracking task and this is all taken care of in the controller.

#### Any relevant tickets

Closes #1521

### DIFF
--- a/server/xpub-controller/helpers/files.js
+++ b/server/xpub-controller/helpers/files.js
@@ -58,6 +58,26 @@ class FilesHelper {
     throw new Error('Invalid parameters to calculate the upload file progress.')
   }
 
+  static startFileProgress(
+    pubsub,
+    ON_UPLOAD_PROGRESS,
+    startedTime,
+    predictedTime,
+    manuscriptId,
+    interval,
+  ) {
+    return setInterval(
+      FilesHelper.publishPredictedProgress(
+        pubsub,
+        ON_UPLOAD_PROGRESS,
+        startedTime,
+        predictedTime,
+        manuscriptId,
+      ),
+      interval,
+    )
+  }
+
   static endFileProgress(pubsub, ON_UPLOAD_PROGRESS, progress, manuscriptId) {
     pubsub.publish(`${ON_UPLOAD_PROGRESS}.${manuscriptId}`, {
       manuscriptUploadProgress: 100,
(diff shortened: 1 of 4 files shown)


hem-brahmbhatt (commented):
Let's add a unit test for this.

diversemix (approved):

server/xpub-controller/helpers/files.js
```diff
@@ -58,6 +58,26 @@ class FilesHelper {
...
 
+  static startFileProgress(
+    pubsub,
+    ON_UPLOAD_PROGRESS,
+    startedTime,
+    predictedTime,
+    manuscriptId,
+    interval,
+  ) {
+    return setInterval(
```
Cooryd, line 69: Added this function to let me mock the calling of `setInterval` in the  maniuscript upload controller unit test without using jest.useFakeTimers as this was throwing errors. see #1553 for a little more context.

hem-brahmbhatt (approved):
I'm not sure how to reproduce the bug so I haven't been able to test if it's fixed or not. Upload progress still works in the review environment

PR #1552, opened 2019-02-20
1520 Test xpub controller helpers

### LINKED ISSUE #1520: Write tests for controller helper functions

SemraSh:
#### Background
Adds test coverage for helpers in xpub controller

Closes #1520 

this is adding tests

### DIFF
--- a/server/xpub-controller/helpers/files.js
+++ b/server/xpub-controller/helpers/files.js
@@ -24,7 +24,7 @@ class FilesHelper {
       manuscriptId,
       url: `manuscripts/${manuscriptId}`,
       filename,
-      type: 'MANUSCRIPT_SOURCE_PENDING', // TODO figure out what this is for
+      type: 'MANUSCRIPT_SOURCE_PENDING',
       mimeType,
     }).save()
     return {
--- a/server/xpub-controller/helpers/files.js
+++ b/server/xpub-controller/helpers/files.js
@@ -24,7 +24,7 @@ class FilesHelper {
       manuscriptId,
       url: `manuscripts/${manuscriptId}`,
       filename,
-      type: 'MANUSCRIPT_SOURCE_PENDING',
+      type: 'MANUSCRIPT_SOURCE_PENDING', // TODO figure out what this is for
       mimeType,
     }).save()
     return {
--- a/server/xpub-controller/helpers/files.test.js
+++ b/server/xpub-controller/helpers/files.test.js
@@ -142,10 +142,12 @@ describe('FilesHelper', () => {
   })
 
   describe('endFileProgress', () => {
+    const manuscriptId = uuid()
+    const ON_UPLOAD_PROGRESS = 'ON_UPLOAD_PROGRESS'
+    jest.useFakeTimers()
     it('publishes progress as a 100 percent', () => {
-      const ON_UPLOAD_PROGRESS = 'ON_UPLOAD_PROGRESS'
-      const manuscriptId = uuid()
       const pubsubMock = {
+        // TODO use jest mocks
         publish: (target, message) => {
           expect(target).toBe(`${ON_UPLOAD_PROGRESS}.${manuscriptId}`)
           expect(message.manuscriptUploadProgress).toEqual(100)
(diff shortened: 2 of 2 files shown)


server/xpub-controller/helpers/files.js
```diff
@@ -24,7 +24,7 @@ class FilesHelper {
       manuscriptId,
       url: `manuscripts/${manuscriptId}`,
       filename,
-      type: 'MANUSCRIPT_SOURCE_PENDING',
+      type: 'MANUSCRIPT_SOURCE_PENDING', // TODO figure out what this is for
```
Cooryd: To differentiate between an already existing `MANUSCRIPT_SOURCE` file in the db and one that is pending being uploaded. The removal of the old file doesn't happen till later in the upload function so this is to prevent the db ever having two `MANUSCRIPT_SOURCE` files for one manuscript.

server/xpub-controller/helpers/files.test.js
```diff
@@ -1,25 +1,75 @@
...
 const FilesHelper = require('./files')
+const ManuscriptModel = require('@elifesciences/xpub-model').Manuscript
+
+const filesHelper = new FilesHelper(config)
+
+describe('FilesHelper', () => {
+  // beforeEach(async () => {
+  //   manuscriptId = uuid()
+  //   await createTables(true)
+  // })
```
hem-brahmbhatt: Not sure is this is a typo?

hem-brahmbhatt (approved):
I paired with @SemraSh on this so it's worth @diversemix and @Cooryd to review as well.

hem-brahmbhatt:
There's more tests left to add but they can come along in a separate PR.

server/xpub-controller/helpers/files.test.js
```diff
@@ -90,4 +138,39 @@ describe('FilesHelper Test', () => {
...
+        manuscriptId,
+      )
+      expect(pubsubMock.publish).toHaveBeenCalledWith(
+        `${ON_UPLOAD_PROGRESS}.${manuscriptId}`,
+        { manuscriptUploadProgress: 100 },
+      )
+    })
+
+    it('clears the progress interval', () => {
+      const progress = setInterval(() => {}, 10)
```
diversemix, line 165: I thought setInterval returned a timeout Object? ... maybe rename

server/xpub-controller/helpers/files.test.js
```diff
@@ -1,25 +1,70 @@
...
 const FilesHelper = require('./files')
+const ManuscriptModel = require('@elifesciences/xpub-model').Manuscript
+
+const filesHelper = new FilesHelper(config)
+
+describe('FilesHelper', () => {
+  describe('validateFileSize', () => {
+    const maxSize = config.get('fileUpload.maxSizeMB')
+    const invalidFileSize = (maxSize + 1) * 1e6
+    const validFileSize = (maxSize - 1) * 1e6
```
diversemix, line 13: Why the minus one? Would be good to check the max size works.

server/xpub-controller/helpers/files.test.js
```diff
@@ -1,25 +1,70 @@
...
+        mimetype: 'mimetype',
+      }
+      const filePromise = Promise.resolve(file)
+
+      const { stream } = file
+      const fileData = await FilesHelper.generateFileEntity(filePromise, id)
+      expect(fileData.stream).toEqual(stream)
+      expect(fileData.fileEntity).toMatchObject({
+        manuscriptId: id,
+        id: expect.any(String),
```
diversemix, line 42: We could check the length here?

diversemix (approved):

server/xpub-controller/helpers/files.test.js
SemraSh: I agree, but I think even checking the length is not enough, we should be able to check if this is the exact id we've created, so this is going to change as soon as we start jest mocking instead of using the actual models. I will change it to check the length for now.

server/xpub-controller/helpers/files.test.js
SemraSh: in the function endFileProgress gets progress as argument, I think progress makes sense to keep clear what is it that we are ending.

PR #1556, opened 2019-02-21
Switch to using PDF for the cover letter

### LINKED ISSUE #1497: Export cover letter as PDF
Currently, we convert the author's cover letter into a HTML file before export in the MECA package, and this is resulting in some unwanted behaviour in eJP. We should export the file in PDF format instead.

Note that, once we are no longer integrating with eJP, we will want to retain the HTML format so that we can display the cover letter within the application.

### LINKED ISSUE #1583: Co-ordinate merge of #1556 with eJP
When eJP switch over the cover letters to using PDFs - merge:

#1556

diversemix:
#### Background

Switches to using a PDF for the cover letter

#### Any relevant tickets

Closes #1583

Unit tests have been updated


Cooryd (approved):

server/meca/file-generators/coverLetter.test.js
```diff
@@ -1,16 +1,39 @@
...
+    pdfParser.on('pdfParser_dataReady', pdfData => {
+      expect(errors).toBe(0)
+      expect(pdfParser.data).not.toBe(null)
+
+      const text = pdfParser.data.Pages[0].Texts.map(item =>
+        item.R.map(t => t.T).reduce((prev, curr) => prev + curr),
+      ).reduce((prev, curr) => prev + curr)
+
+      const jsonData = JSON.stringify(text, null, 2)
+      expect(jsonData).toBe('"%20C%20overLetterPickMe!I\'msimplythebest"')
```
hem-brahmbhatt, line 36: Does the following work? 
https://nodejs.org/api/buffer.html#buffer_buf_includes_value_byteoffset_encoding
```suggestion
      expect(document.includes('"%20C%20overLetterPickMe!I\'msimplythebest"')).toBe(true)
```

Cooryd: `jsonData` is a string because of `stringify` so I think this should be okay no?

diversemix: yeah the pdf2json library decodes the PDF stream which is not as easy as you might think. Then the `map-map-reduce-reduce` has to do some funky things to get the text out of the json.

hem-brahmbhatt (approved):

hem-brahmbhatt (approved):
Looks good, happy for this to go in if everything works on the review environment.

diversemix (commented):
# Not to be merged

jyenes (approved):
Looks Good

PR #1558, opened 2019-02-21
suppressing the cover letter and the name of the hotjar

### LINKED ISSUE #1557: Suppress cover letter input and profile menu from HotJar recordings
It's come to our attention that HotJar screen recordings are not suppressing input into the cover letter editor or the name displayed in the profile menu, which means that certain recordings can be linked to a named user.

This is undesirable behaviour and we should work to completely anonymise these recordings.

See documentation here: https://help.hotjar.com/hc/en-us/articles/115012439167-Suppressing-Content-in-Recordings

jyenes:
suppressing the cover letter and the name of the profile of the hotjar info recorded

closes: #1557


app/components/global/AppBar/ProfileMenu.js
```diff
@@ -144,7 +144,9 @@ class ProfileMenu extends React.Component {
             <PopOverPanel>
               <Menu>
                 <MenuItem>
-                  <MenuHeading>{user.identities[0].name}</MenuHeading>
+                  <MenuHeading data-hj-suppress="">
```
hem-brahmbhatt: May be worth adding a test for this. It'll help us understand that it's for hotjar.

diversemix: Agreed.... if its more than 20mins then use another PR :+1:

jyenes: Not sure what do you think it can be tested.
As far  as I know the issue is, hotjar is recording customer navigation, so what is required is in those recording to suppress certain fields, this behaviour cannot be reproduced in our website rendering the page, the only thing can be saw is that the property `data-hj-suppress` is present in the dom object, as you can see in the attached screenshot.

To check if this is actually suppressing the field, I'm afraid the only way is going to the hotjar website and checking whether the recordings don't contain the suppressed fields.

screenshot 2019-02-22 at 09 40 30

hem-brahmbhatt: I think if we assert that the menu heading containing the name of the user has that attribute, it allows us to write a test explaining it's for suppressing the value from hotjar, i.e. `it('should be suppressed from hotjar')`

We can assume the library is tested already and will respect the attribute.

diversemix (approved):

app/components/global/AppBar/ProfileMenu.js
```diff
@@ -144,7 +144,7 @@ class ProfileMenu extends React.Component {
             <PopOverPanel>
               <Menu>
                 <MenuItem>
-                  <MenuHeading data-hj-suppress="">
+                  <MenuHeading data-hj-suppress="" data-test-id="profile-name">
```
Cooryd, line 147: May be useful to also include a unit test for this as well as the browser test? there's already a `ProfileMenu.test.js` file so it would just need to shallow render this and assert the attribute is in the dom tree.

jyenes: done

Cooryd (approved):

hem-brahmbhatt (approved):

PR #1562, opened 2019-02-21
Remove unicode chars for ejp consumption

### LINKED ISSUE #1561: MECA: Remove unicode from filenames
EJP cannot store unicode in the database - so we need to remove any unicode from the filenames on export

diversemix:
#### Background

As title

#### Any relevant tickets

Closes #1561

Tests have been updated - still needs testing with eJP

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,6 +13,7 @@
   "extends": ["pubsweet"],
   "parser": "babel-eslint",
   "rules": {
+    "no-control-regex": 0,
     "global-require": 0,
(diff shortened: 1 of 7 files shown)


diversemix:
As mentioned in standup - this has a slight issue that when the only uniqueness in the filename is the emoji - we can potentially overwrite files. This is going to be fixed in another issue.

Cooryd (approved):

hem-brahmbhatt (approved):

PR #1572, opened 2019-02-25
1563 remove file on error

### LINKED ISSUE #1563: Clean db of manuscript source files if upload fails
**Describe the bug**

Currently we run a function called `validateManuscriptSource` when uploading a manuscript file which throws an error if there are more than one files linked to the manuscript with each of the type `MANUSCRIPT_SOURCE` and `MANUSCRIPT_SOURCE_PENDING`. This is correct as there should never be more than on manuscript source and one file in the process of replacing that file.

Currently if the upload process being carried out by `uploadManuscriptFile` fails before the file has been successfully uploaded, a file gets 'stuck' in the `MANUSCRIPT_SOURCE_PENDING` state. On attempting to re-try, we create a new file with type `MANUSCRIPT_SOURCE_PENDING` which then causes `validateManuscriptSource` to throw another error as there are now 2 `MANUSCRIPT_SOURCE_PENDING`.

**Steps To Reproduce**

- Edit the code to simulate an error occurring when uploading to S3 or extracting the title in scienceBeam
- Attempt to upload a file, it should error.
- Remove the simulated error 
- retry uploading a file

**Expected behavior**

An error being thrown should not cause the manuscripts files to be in an 'invalid' state. On catching an error, any `MANUSCRIPT_SOURCE_PENDING` file related to the manuscript should be removed.

**Screenshots**

Cooryd:
#### Background

Adds a new manuscript helper function that deletes any `File` entities with the type `MANUSCRIPT_SOURCE_PENDING` that are related to a passed in manuscript. This function is then called if the `uploadManuscriptFile` function ever throws an error to ensure the manuscript is in a valid state to re-try uploading a new file.

#### Any relevant tickets

Closes #1563 

### DIFF
--- a/server/xpub-controller/helpers/manuscript.test.js
+++ b/server/xpub-controller/helpers/manuscript.test.js
@@ -29,7 +29,7 @@ describe('clearPendingFile()', () => {
       url: '/',
     })
     await fileEntity.save()
-    await fileEntity2.save()
+    const sourceFile = await fileEntity2.save()
     manuscript = await Manuscript.find(id, userId)
     expect(manuscript.files).toHaveLength(2)
 
@@ -37,5 +37,7 @@ describe('clearPendingFile()', () => {
     manuscript = await Manuscript.find(id, userId)
 
     expect(manuscript.files).toHaveLength(1)
+    expect(manuscript.files[0].id).toEqual(sourceFile.id)
+    expect(manuscript.files[0].type).not.toEqual('MANUSCRIPT_SOURCE_PENDING')
   })
 })
--- a/server/xpub-controller/helpers/manuscript.test.js
+++ b/server/xpub-controller/helpers/manuscript.test.js
@@ -1,0 +1,41 @@
+const { createTables } = require('@pubsweet/db-manager')
+const { User, Manuscript, File } = require('@elifesciences/xpub-model')
+const ManuscriptHelper = require('./manuscript')
+
+describe('clearPendingFile()', () => {
+  let userId
+
+  beforeEach(async () => {
+    await createTables(true)
+    const profileId = 'ewwboc7m'
+    const identities = [{ type: 'elife', identifier: profileId }]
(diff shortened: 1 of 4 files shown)


server/xpub-controller/manuscript.js
```diff
@@ -56,6 +55,8 @@ class Manuscript {
         progress,
         manuscriptId,
       )
+      const manuscript = await ManuscriptModel.find(manuscriptId, this.userId)
+      await ManuscriptHelper.clearPendingFile(manuscript)
```
hem-brahmbhatt, line 59: It may be worth doing this in a `finally` statement. If for some reason an error gets thrown when we call `endFileProgress`, the file would be left in a pending state.

hem-brahmbhatt: Happy for this to happen in a separate PR btw.

server/xpub-controller/helpers/manuscript.test.js
```diff
@@ -0,0 +1,41 @@
...
+    })
+    await fileEntity.save()
+    await fileEntity2.save()
+    manuscript = await Manuscript.find(id, userId)
+    expect(manuscript.files).toHaveLength(2)
+
+    await ManuscriptHelper.clearPendingFile(manuscript)
+    manuscript = await Manuscript.find(id, userId)
+
+    expect(manuscript.files).toHaveLength(1)
```
hem-brahmbhatt, line 39: We should add an assertion to make sure the remaining file has the same id as file2.

hem-brahmbhatt (approved):

server/xpub-controller/manuscript.js
Cooryd: As we're going to be re-thinking this controller code lets leave this till then. This case fixes the bug that is currently in production and we can consider where else we might need the clean up task ran in the controller.

diversemix (approved):

PR #1573, opened 2019-02-25
Fix bad manifest file 

### LINKED ISSUE #1571: eJP / MECA issues in production
Hi Peter and Hannah,

At least the last three submissions using xPub have come through without the article file. The one that failed first time yesterday also came without an article file, so it seems there is a more general problem. 

Is it possible to explore and correct the issue ASAP (and it's likely to be an eJP issue?) and/or dial down the percentage (0%?) until submissions come complete with the article file again?

Unfortunately at the moment authors will need to complete the submission using eJP, as we QC fail papers without an article file. 

Thanks! 

Andy

diversemix:
#### Background

- Adds a test now for the manifest file
- Fixes the issue

#### Any relevant tickets

Closes #1571 

New test written


server/meca/export.test.js
```diff
@@ -50,6 +51,20 @@ describe('MECA integration test', () => {
...
+      const zip = await JsZip.loadAsync(
+        sftp.mockFs.readFileSync(`/test/${finalName}`),
+      )
+
+      const manifest = await zip.files['manifest.xml'].async('string')
+      const manifestJson = JSON.parse(xml2json.toJson(manifest))
+
+      expect(manifestJson).toMatchSnapshot()
+    })
+
```
hem-brahmbhatt, line 67: We already have unit tests for the manifest generator [here](https://github.com/elifesciences/elife-xpub/blob/develop/server/meca/file-generators/manifest.test.js). Does this add any more coverage than the existing unit tests?

hem-brahmbhatt: If it does, it may be worth moving the coverage to the generator tests mentioned above.

diversemix: Good point - those tests (obviously) don't work! I'm going to be looking at them and sort them out if its reasonably straight forward as part of this PR.

diversemix: Been thinking about these tests - and these tests do things at a lower layer with the `manuscript.files` objects. This is distinctly different to the tests in export that test the entire export process with the `manuscript` object.

Cooryd (approved):

jyenes (approved):
Looks good

PR #1620, opened 2019-02-28
update titles on pages

### LINKED ISSUE #1586: Update page titles
## Problem / Motivation
The current page titles for all https://reviewer.elifesciences.org pages are 'xpub'. As this isn't relevant for authors and shows up in various places (Tab labels, Google SERPS etc.) I would like the page titles to be updated.

## Proposed solution
- Update the page titles to conventions that are the same as the main eLife site
- e.g.
"Submit | eLife" for https://reviewer.elifesciences.org/login
"Author guide | eLife" for https://reviewer.elifesciences.org/author-guide/editorial-process

- Can we also add an icon that appears in the tab?

---

## Tasks

- [ ] A description of a task
- [ ] A description

## Technical notes
- More notes or suggestions

## User interface / Wireframes

@hdrury1

jyenes:
update titles on the pages

closes: #1586

### DIFF
--- a/app/trackedRoute.js
+++ b/app/trackedRoute.js
@@ -5,7 +5,6 @@ import config from 'config'
 
 export default class TrackedRoute extends React.Component {
   trackPage(page) {
-    document.title = config.titles[page.split('/')[1]] || 'eLife'
     this.lastTracked = page
     ReactGA.set({
       page,
@@ -15,6 +14,7 @@ export default class TrackedRoute extends React.Component {
 
   render() {
     const { path } = this.props
+    document.title = config.titles[path.split('/')[1]] || 'eLife'
     if (path !== this.lastTracked) {
       this.trackPage(path)
     }
--- a/app/trackedRoute.js
+++ b/app/trackedRoute.js
@@ -5,7 +5,7 @@ import config from 'config'
 
 export default class TrackedRoute extends React.Component {
   trackPage(page) {
-    document.title = config.titles[page.split('/')[1]]
+    document.title = config.titles[page.split('/')[1]] || 'eLife'
     this.lastTracked = page
     ReactGA.set({
       page,
--- a/app/trackedRoute.js
+++ b/app/trackedRoute.js
@@ -1,9 +1,11 @@
 import React from 'react'
 import { Route } from 'react-router-dom'
 import ReactGA from 'react-ga'
+import config from 'config'
 
 export default class TrackedRoute extends React.Component {
   trackPage(page) {
+    document.title = config.titles[page.split('/')[1]]
     this.lastTracked = page
     ReactGA.set({
       page,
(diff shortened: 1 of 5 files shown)


app/trackedRoute.js
```diff
@@ -1,9 +1,11 @@
 import React from 'react'
 import { Route } from 'react-router-dom'
 import ReactGA from 'react-ga'
+import config from 'config'
 
 export default class TrackedRoute extends React.Component {
   trackPage(page) {
+    document.title = config.titles[page.split('/')[1]]
```
hem-brahmbhatt: I don't feel this is the correct function to put this in, since this function is to do with page tracking. Instead, I think this is something that can happen on `render` or on `componentDidMount`.

jyenes: I agree on that, it shouldn't be on the trackPage function, render should be the place.

hdrury1:
This matches product requirements... quick query though @jyenes - I notice that in the review environment 'xpub' appears momentarily before the page completes load the the updated title appears. Will this be an issue in production do you know?

diversemix:
I notice that too ... and also with a bad page: http://35.172.178.78:32592/asd (for example)

hem-brahmbhatt:
> This matches product requirements... quick query though @jyenes - I notice that in the review environment 'xpub' appears momentarily before the page completes load the the updated title appears. Will this be an issue in production do you know?

Yes, this will be an issue.

The initial title you see gets set [here](https://github.com/elifesciences/elife-xpub/blob/develop/app/index-production.html#L12). This title gets set when the app gets built during deployment. It's not trivial to change this. The options I can see are:

1. Don't set a title by default, and rely on the title being set when the React app has finished loading it in the browser
2. Use a placeholder title, such as _Loading..._
3. Load the title with a custom script before the rest of the app has a chance to load. This will set the initial title to be based on the current URL. We can insert the script [here](https://github.com/elifesciences/elife-xpub/blob/develop/app/index-production.html#L10).

app/trackedRoute.js
```diff
@@ -13,6 +14,7 @@ export default class TrackedRoute extends React.Component {
 
   render() {
     const { path } = this.props
+    document.title = config.titles[path.split('/')[1]] || 'eLife'
```
hem-brahmbhatt: This is something which could stop the rest of the page rendering if an exception is thrown, which I've seen happen in the past. I think it's worth adding some of the following around `config.titles[path.split('/')[1]]`:
- error handling
- null checks
- setting more defaults

jyenes: added error handling

jyenes:
I have set the initial title to an empty string so now should be ok

hdrury1 (approved):
lgtm

hem-brahmbhatt (approved):

PR #1625, opened 2019-03-01
add auditLog on ejp callback

### LINKED ISSUE #1616: Add an Audit Log entry for the eJP callback
## Problem / Motivation

- WHO wants this? Developers / Editorial
- WHEN and WHERE do they want it? ASAP - this is 
- WHAT do they want? To be able to track the xPub ID against the eJP ID
- WHY do they want this? To make it easier for the developer to trace issues.

## Proposed solution
Add an audit entry when a success callback comes in with the result - this should include the eJP id

jyenes:
add AuditLog on EJP Callback

closes #1616


server/meca/routes.js
```diff
@@ -37,23 +37,37 @@ module.exports = app => {
...
+        .catch(err => {
+          throw new Error(`MECA import failure email failed to send. ${err}`)
+        })
 
     updateStatus()
       .then(() => {
         if (status === Manuscript.statuses.MECA_IMPORT_FAILED) {
           return sendEmail()
         }
+        auditlog()
```
Cooryd: I don't think the audit will happen if the mailer above throws an error? Should this be done before in its own `.then`?

Cooryd: Didn't realise this was a draft so you may have picked up on this already. Can you put `WIP:` in the title still so it's a bit clearer on the PR page?

jyenes: It's ok good feedback, anyway!

server/meca/routes.js
```diff
@@ -37,20 +37,34 @@ module.exports = app => {
...
-      throw new Error(`MECA import failure email failed to send. ${err}`)
-    })
+        })
+        .catch(err => {
+          throw new Error(`MECA import failure email failed to send. ${err}`)
+        })
 
     updateStatus()
       .then(() => {
+        auditlog()
```
hem-brahmbhatt: `auditLog()` returns a promise, so you'll want to wrap this within a separate `.then`

server/meca/routes.js
```diff
@@ -37,19 +37,33 @@ module.exports = app => {
...
-      from: config.get('meca.email.sender'),
-      subject: `${config.get('meca.email.subjectPrefix')}MECA import failed`,
-      text: `
+    const auditlog = () =>
+      new AuditLog({
+        action: 'UPDATED',
+        objectId: manuscriptId,
+        objectType: 'ejp.id',
+        value: body.id,
+      }).save()
```
Cooryd: Is it possible to add a unit test to ensure `.save()` is called on the `AuditLog` object?

hem-brahmbhatt: It's easier to do the same as https://github.com/elifesciences/elife-xpub/blob/develop/server/xpub-server/entities/manuscript/resolvers/uploadSupportingFile.test.js for now. I don't think we have an example of a test where we mock out `.save()`

Cooryd (approved):
Looks good

server/meca/routes.test.js
```diff
@@ -107,4 +107,24 @@ describe('MECA HTTP callback handler', () => {
...
+      .expect(204)
+    const updatedManuscript = await Manuscript.find(manuscript.id, userId)
+    expect(updatedManuscript.status).toBe(
+      Manuscript.statuses.MECA_IMPORT_SUCCEEDED,
+    )
+    const audits = await AuditLog.all()
+    expect(audits).toHaveLength(2)
+    expect(audits[0].value).toBe('MECA_IMPORT_SUCCEEDED')
+    expect(audits[1].objectType).toBe('ejp.id')
+    expect(audits[1].objectId).toBe(manuscript.id)
```
hem-brahmbhatt: I don't think we should have a separate log entry for this alongside the existing one.

Currently, the AuditLog is meant to show changes made to the database. Recording the `ejp.id` in a separate AuditLog entry goes against that pattern.

Instead, I think we should be able to contain the `ejp.id` within the initial AuditLog entry. We can put it within either the `value` field, or add an additional column for `userType` to store a value like `EJP`, and the `userId` field would contain the EJP id.

@diversemix would be good to get your opinion on this.

hem-brahmbhatt: Current AuditLog entries are as follows:
```
   [ 
      AuditLog {
        id: 'a9d27bc9-40d6-4736-851b-3bb35bc67b8a',
        created: 2019-03-04T14:47:21.311Z,
        userId: null,
        objectId: '5db164cc-9c52-4948-8f0f-323f53e6bd49',
        objectType: 'manuscript.status',
        updated: 2019-03-04T14:47:21.311Z,
        value: 'MECA_IMPORT_SUCCEEDED',
        action: 'UPDATED' },
      AuditLog {
        id: 'fb5ac23b-7f1c-4345-b2c9-f99368b035d5',
        created: 2019-03-04T14:47:21.384Z,
        userId: null,
        objectId: '5db164cc-9c52-4948-8f0f-323f53e6bd49',
        objectType: 'ejp.id',
        updated: 2019-03-04T14:47:21.384Z,
        value: null,
        action: 'UPDATED' } 
    ]
```

jyenes: agree, let's hold this on till have a chat with Peter, to see what he meant

diversemix: Ok, so my point of view - happy for it to be challenged....

First AuditLog entry - this should be done in the REST route of the POST `/meca-result` to record the fact an external entity has called into the application.

Second AuditLog entry I would expect to be the recording of that consequences of that request - in this case on the Manuscript's `state`

hem-brahmbhatt: The first AuditLog entry is captured by our logging, and I think that's the correct place for us to be logging that information. 

I prefer to have the AuditLog used to track database changes, and the database to contain all the information/IDs we need to then use in other systems to continue tracing any external services we're using, like S3.

diversemix: Talking with @hdrury1 ... It is a product requirement to be able to track (audit) the lifecycle of a manuscript for the lifetime of the product (i.e. beyond the 30 day limit of the logs). Ideally we do want to also track S3 and SFTP interactions as well in the future.

server/meca/routes.js
```diff
@@ -29,6 +29,17 @@ module.exports = app => {
       return
     }
 
+    try {
+      await new AuditLog({
+        action: 'RECORD',
```
diversemix: This action should be something like `POST` or `MECA_RESULT`

server/meca/routes.js
```diff
@@ -29,6 +29,17 @@ module.exports = app => {
...
+        objectId: manuscriptId,
+        objectType: 'ejp',
```
diversemix: This should be `manuscript` as `objectId` is `manuscriptId`

server/meca/routes.js
diversemix: The subject is the client that made the request (`userId`) - i.e. `EJP`

diversemix: ... maybe we need to create a User for ejp?

server/xpub-model/entities/auditLog/index.js
```diff
@@ -12,20 +12,18 @@ class AuditLog extends BaseModel {
         userId: { type: 'uuid' }, // Currently not used
         action: {
           type: 'string',
-          enum: ['CREATED', 'UPDATED', 'DELETED'],
+          enum: ['CREATED', 'UPDATED', 'DELETED', 'RECORD'],
```
diversemix: After a discussion here ... we're thinking that:
 -  'CREATED', 'UPDATED', 'DELETED' applies to database operations
 - Any other action should be explicit i.e. `MECA_RESULT`, `S3_UPLOADED`, `SFTP_UPLOADED` etc

diversemix (changes requested):

server/meca/routes.js
jyenes: I think we don't need the userId, because action MECA_RESULT can only come from EJP right?
so adding userId, would be in somehow redundant

SemraSh (approved):

diversemix (approved):

PR #1630, opened 2019-03-04
1609 Move client directory in xpub controller

### LINKED ISSUE #1609: Move the client directory in xpub-controller out into its own top level server directory
## Background
As part of #1587 we want to relocate the client directory found in `xpub-controller`.

Current `/client` sits inside the `/xpub-controller` and contains interfaces such as `S3Storage` and `ScienceBeamApi` which are used by controllers. Within `/xpub-model` we have another 'client' defined as `ElifeApi` which is used by the model. It would be useful to group these clients together in one place however our model collection and controller collection should be kept separate. 

## Proposed solution
We move `client` up to a top level `/server` directory (alongside `xpub-model` and `xpub-controller`. Maybe with a name like `xpub-client`? 

This will also require the changing of all `require` statements for the client functions / classes throughout the codebase to look for the newly located client directory.

For context see : #1084

SemraSh:
#### Background
This pr is moving the client folder to it's own folder outside xpub-controller, named xpub-clinet

What does this PR do?

Partial work of refactoring xpub-controller

#### Any relevant tickets
Closes #1609 

Refactor | existing tests passed


hem-brahmbhatt (approved):

Cooryd:
Can we continue to add a `WIP:` prefix on draft PRs please? As it's a new feature it's still not clear when you click through from the PR list that you'll be looking at a work in progress vs a completed task

Cooryd (approved):
Looks good

Issue #1479, opened 2019-02-07
Coverage reporting generates errors


diversemix:
Typical Error:
```
Failed to collect coverage from /home/peter/dev/elife-xpub-2/app/components/pages/StaticPages/AuthorGuide/RequiredInfoTable.js
ERROR: 'import' and 'export' may appear only with 'sourceType: "module"' (1:0)
```

For fix see
https://github.com/facebook/jest/issues/6053#issuecomment-383632515

SemraSh:
This has became priority because it is making the debugging harder and slowing the team down.

Issue #1524, opened 2019-02-14
Investigate: Adding coveralls for coverage reports


hem-brahmbhatt:
We currently manage our own coverage reporting to let us know if our test coverage is below accepted levels. Coveralls is a service which manages this for you.

https://coveralls.io/

- Integrates with Github
- Free on open source repositories
- Means we don't have to manage our own coverage reporting

Investigate the following:
- [ ] what steps are needed to add this to `elife-xpub`?
- [ ] what steps are needed to remove existing test coverage code?
- [ ] what reporting/information do we get from coveralls?

SemraSh:
We have decided to use Codecov instead, to check the reasons please see #1638  
I will close this ticket

PR #1634, opened 2019-03-05
[WIP] Investigate ignore the errors in code coverage

SemraSh:
This is an investigation of how to keep seeing the code coverage without seeing the errors in development stated in #1479 .

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: '70...100'
+
+  status:
+    project: yes
+    patch: yes
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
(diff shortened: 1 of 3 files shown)


diversemix:
@SemraSh ... how does this relate to #1524  ?

SemraSh:
@diversemix fixing this issue will require another ticket which is to upgrade the babel version. With this ticket I aimed to keep the coverage in the ci and remove it from development. Codecov is giving a good report that will keep the same amount of information. I will close this ticket and create a new one for adding code coverage in the ci.

SemraSh:
As we are using `monorepo-structure` we need to convert the `.babelrc` to `babel.config.js` To solve the issue described in #1479 we need to upgrade the version of babel-core and convert the .babelrc file to be babel.config.js. Please see [this](https://babeljs.io/docs/en/config-files#monorepos) for more information on babel configuration in monorepos. 
The issue that follows this is #1639 
I have also created a issue for code coverage report to be added in ci, please see #1638 .

PR #1636, opened 2019-03-05
Index unicode removed files.

### LINKED ISSUE #1565: Fix potential overwriting of files with emojis
Further to #1561 ... there is a potential possibility that files could get overwritten, for example this file list:

- 👍`File1`
- 👉`File1`

Would just become one file `File1`

Suggested solution is to index the files when we replace unicode ... so in fact the above list would become:
- `1-File1`
- `2-File1`

diversemix:
#### Background

Indexes unicode removal so as to not overwrite files with the same name.

#### Any relevant tickets

Closes #1565

Tests have been written / updated


jyenes (approved):
Looks good, the only point would be maybe adding a unit test for removeUnicode.js

Cooryd (approved):

PR #1644, opened 2019-03-06
1590 create pubsub client class

### LINKED ISSUE #1590: Create a pubsub client class
## Problem / Motivation

As part of the refactoring of the controller (see #1587) it would be beneficial to have a `PubsubClient` class which could be accessed from the controllers client collection (same place as the `scienceBeamApi.js` and `S3Storage.js` client files). This client should make it easier to test processes relating to a pubsub connection, including allowing us to test without a database backing the tests (as this is required through using the `pubsweet` pubsub manager). It would also allow us to store the pubsub instance and any intervals (for sending periodic messages to the app client) as internal properties of the client instance.

### Location
❌ 
```
/xpub-controller
  /client
    pubsub.js
```

✅ 
```
/xpub-client
  ...
  pubsub.js
  pubsub.test.js
```

### Sudo-code
```
class PubsubClient {
  constructor(pubsubManager){
    this.manager = pubsubManager
    this.pubsub = null
    this.pubsubInterval = null
  }
  initialize() {
    this.pubsub = await this.manager.getPubsub()
  }
  publish(message, data) {
    this.pubsub.publish(message, data)
  }
  startInterval(message, data, interval) {
    this.pubsubInterval = setInterval(this.publish(message, data), interval)
  }
  clearInterval() {
    clearInterval(this.pubsubInterval)
    this.pubsubInterval = null
  }
}
```

This should have a full test suite written so its initializer can focus on asserting function calls rather than functionality

It does not need to be implemented anywhere yet as the full controller refactor hasn't taken place yet.

SemraSh:
#### Background

Adds `PubsubClient` class in `xpub/pubsub/index.js` which could be accessed from the controllers

#### Any relevant tickets

Closes #1590 


Cooryd:
I think we should move these out of a directory and have them in a top level `pubsub.js` directly in `/xpub-client`

jyenes (approved):
Looks good

server/xpub-client/pubsub/index.js
```diff
@@ -0,0 +1,29 @@
...
+      interval,
+    )
+  }
+
+  stopPublishingOnInterval(message, dataCallback) {
+    clearInterval(this.publishingInterval)
+    this.publishingInterval = null
+
+    if (message && dataCallback) {
+      this.pubsub.publish(message, dataCallback())
```
Cooryd: Client consumers may want to call publish on its own without an interval. Could we create a `publish` function on the object that just calls `this.pubsub.publish` so that calling it is less verbose

diversemix (approved):

diversemix (approved):

server/xpub-client/pubsub.test.js
```diff
@@ -0,0 +1,118 @@
...
+      const pubsub = createPubsub(mockGetPubsub)
+      await pubsub.initialize()
+
+      expect(mockGetPubsub).toBeCalled()
+      expect(pubsub.pubsub).toBe(mockPubsub)
+    })
+  })
+
+  describe('publish', () => {
+    it('calls pubsub.publish and uses the callback to get the message value', async () => {
```
Cooryd, line 26: These tests relate to the start and stop interval. maybe a new test for publish would simply check it calls `publish` on its `pubsub` property and correctly forwards the params on?

hem-brahmbhatt (approved):

server/xpub-client/pubsub.js
```diff
@@ -0,0 +1,33 @@
...
+    this.pubsub = null
+    this.publishingInterval = null
+  }
+
+  async initialize() {
+    this.pubsub = await this.manager.getPubsub()
+  }
+
+  publish(message, dataCallback) {
+    this.pubsub.publish(message, dataCallback())
```
Cooryd, line 13: Could `dataCallback` be changed to a value here and in the `stopPublishingOnInterval` ? The reason `startPublishingOnInterval` is passed a callback instead of a value is that the value is likely to change each time the interval is called ie: progress. The first time this is called, progress may be 0 but you'll want this value to change over time. Publish (and the extra message called in `stopPublishingOnInterval` ) are one off calls so the value will not change.

Cooryd: This can be raised as a new PR later

PR #1646, opened 2019-03-06
1528 - ScienceBeam feedback

### LINKED ISSUE #1528: Implementing ScienceBeam Feedback loop
Whenever there is a change to the manuscript title log the change out (along with the manuscript id).
This will allow the logs to be parsed to generate a set of incorrectly extracted titles to train sciencebeam with.

I think this is the bear minimum - we could also store the extracted title in the database - that way we can maintain the history. 

/cc @de-code

Cooryd:
#### Background

Adds a SemanticExtraction model to the app. An instance of this extraction is created and saved when ScienceBeam extracts a manuscripts title. This information will then be used in conjunction with the `Manuscript` title once it has been fully submitted to provide feedback to guide improvments.

#### Any relevant tickets

closes #1528 

### DIFF
--- a/server/xpub-controller/helpers/files.js
+++ b/server/xpub-controller/helpers/files.js
@@ -1,5 +1,7 @@
 const logger = require('@pubsweet/logger')
 const FileModel = require('@elifesciences/xpub-model').File
+const SemanticExtractionModel = require('@elifesciences/xpub-model')
+  .SemanticExtraction
 
 class FilesHelper {
   constructor(config, scienceBeamApi) {
@@ -40,10 +42,7 @@ class FilesHelper {
     predictedTime,
     manuscriptId,
   ) {
-    if (
-      manuscriptId.length === 36 &&
-      predictedTime > 0
-    ) {
+    if (manuscriptId.length === 36 && predictedTime > 0) {
       return () => {
         const elapsed = Date.now() - startedTime
         let progress = parseInt((100 * elapsed) / 1000 / predictedTime, 10)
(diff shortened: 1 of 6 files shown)


server/xpub-model/migrations/1551872573-semantic-extraction.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE semantic_extraction (
+    id UUID PRIMARY KEY,
+    created TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT current_timestamp,
+    updated TIMESTAMP WITH TIME ZONE,
+    manuscript_id UUID NOT NULL,
+    field_name TEXT NOT NULL,
+    value TEXT NOT NULL
```
Cooryd: this probably should be nullable

server/xpub-model/entities/semanticExtraction/index.js
```diff
@@ -0,0 +1,24 @@
+const BaseModel = require('@pubsweet/base-model')
+
+class AuditLog extends BaseModel {
```
diversemix: Incorrect name?

server/xpub-controller/helpers/files.js
```diff
@@ -127,6 +126,14 @@ class FilesHelper {
         filename,
       })
     }
+
+    const semanticExtractionEntity = new SemanticExtractionModel({
```
diversemix: I'd suggest creating factory methods for all the different field types. That way we don't expose field as a string - rather its a method like `SemanticExtractionModel.createTitleEntry(manuscriptId, title)`

diversemix (approved):

hem-brahmbhatt (approved):

Issue #1472, opened 2019-02-07
FileStatus: Add to client validation (error if wrong)


diversemix:
### Background
Currently the a user can continue through the submission wizard to the submit stage while there are outstanding uploads in progress. This has caused issues with the meca exports not containing the files listed in the manifest.xml (and thus not being imported into ejp). #1471 aims to add server side validation for this to prevent the user being able to submit while files are still uploading. 

### Task
This issue involves adding a client-side validation check on the upload progress of all files, both manuscript and supporting. This should take place at the point of moving from the 'Files' step to the 'Submission' step. The user should not be able to move forward in the wizard from the 'Files' view until all files have been stored in s3. This should be prevented with validation errors.

- [x] fix for uploading Manuscript See (https://github.com/elifesciences/elife-xpub/pull/1654)
- [ ] fix for supporting files

Cooryd:
Description updated. It may be worth seeing #1469 to see if this issue relies on the new virtual attribute.

jyenes:
I'm going to block this issue till #1669 is finished

hdrury1:
@SemraSh Has this issue actually been done? I'm not sure validation for supporting files has been implemented.

PR #1654, opened 2019-03-11
fix for avoid navigate forward once there is a manuscript being uploaded

jyenes:
fix the first task on the #1472 for manuscript upload


diversemix (changes requested):
We need to show validation errors not to disable the Next button (UX requirement)
/cc @chugginselifesciences

jyenes:
closing this as this is not the solution expected

PR #1655, opened 2019-03-11
new relic snippet updated to be SPA specific

### LINKED ISSUE #1464: Ensure New Relic is implementing client side monitoring
New Relic is currently set up to monitor the server side metrics of requests, errors and load ect. We should ensure that client side monitoring is correctly implemented and configured so that page load times, errors ect. are also being recorded.

Cooryd:
#### Background

Our new relic browser settings were set to Pro but needed to be changed to Pro + SPA to better monitor the usage of the app. Doing this required an update to the snippet included in the document head. Licence keys and App ID remain unchanged.

#### Any relevant tickets

Closes #1464 (will require follow up investigation to make sure it works)

### DIFF
--- a/assets/newrelic.js
+++ b/assets/newrelic.js
@@ -5,39 +5,37 @@
  */
 
 window.NREUM || (NREUM = {}),
-  (__nr_require = (function(t, n, e) {
-    function r(e) {
-      if (!n[e]) {
-        var o = (n[e] = {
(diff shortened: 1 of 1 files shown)


diversemix (approved):
Let's give it a go!

SemraSh (approved):
The changes in file seem consistent 👍

PR #1658, opened 2019-03-12
1469 file status virtual attribute

### LINKED ISSUE #1469: FileStatus: Add virtual attribute for file status
**Background & context**

Currently there is a bug that allows the user to submit a manuscript while a supporting file is still in the process of uploading. This causes problems generating the `meca` package as the file exists in the database but the file isn't accessible through S3.

Our intended solution is to prevent the user from clicking 'Next' on the file stage of the wizard when  any associated files do **not** `STORED` or `CANCELLED`  status. 

We aim to achieve this functionality by adding a field to the manuscript of `fileStatus` which will have an enum value of `CHANGING` or `READY`. As this is a derived value, it should exist as a virtual attribute on the `Manuscript` object and not within the database. 

This issue relates only to adding this new virtual attribute and calculating its value. Further work will be raised for the remaining functionality described above.

see: https://vincit.github.io/objection.js/#virtualattributes

hem-brahmbhatt:
#### Background

Adds `manuscript.fileStatus`.

This is implemented as a getter, which can be called by referencing the `manuscript.fileStatus` property.

#### Any relevant tickets

Closes #1469 

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/server/xpub-model/entities/file/index.test.js
+++ b/server/xpub-model/entities/file/index.test.js
@@ -11,6 +11,31 @@ describe('File', () => {
     await createTables(true)
   })
 
+  describe('constructor', () => {
+    it('sets the status to undefined, even with a default within the schema', async () => {
+      const file = new File({
+        manuscriptId: '1',
+        filename: 'thisfile.txt',
+        url: '/an/url',
+      })
+      expect(file.status).toBe(undefined)
+    })
+  })
+
+  describe('save()', () => {
+    it('should set the status to CREATED', async () => {
+      const manuscript = await new Manuscript({
+        createdBy: userId,
+      }).save()
+      const file = await new File({
+        manuscriptId: manuscript.id,
+        filename: 'thisfile.txt',
+        url: '/an/url',
+      }).save()
+      expect(file.status).toBe('CREATED')
+    })
+  })
+
   describe('delete()', () => {
     it('should fail if id is not provided when deleting the file', async () => {
       const file = new File()
@@ -38,17 +63,6 @@ describe('File', () => {
   })
 
   describe('updateStatus()', () => {
-    it('should be initialised in the CREATED state', async () => {
-      const manuscript = await new Manuscript({
-        createdBy: userId,
-      }).save()
-      const file = await new File({
-        manuscriptId: manuscript.id,
-        filename: 'thisfile.txt',
-        url: '/an/url',
-      }).save()
-      expect(file.status).toBe('CREATED')
-    })
     it('sets file status to value passed', async () => {
       const manuscript = await new Manuscript({
         createdBy: userId,
--- a/server/xpub-model/entities/manuscript/index.test.js
+++ b/server/xpub-model/entities/manuscript/index.test.js
@@ -579,7 +579,7 @@ const getThreeVersions = async userId => {
 }
 
 const setStatusOfFile = async (file, manuscript, status) => {
-  file.status = status
+  file.status = status // eslint-disable-line no-param-reassign
   await file.save()
   return Manuscript.find(manuscript.id, manuscript.createdBy)
 }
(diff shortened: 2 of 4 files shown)


server/xpub-model/model.test.js
```diff
@@ -3,6 +3,29 @@ const uuid = require('uuid')
 const Manuscript = require('./entities/manuscript')
 const File = require('./entities/file')
 
+describe('creating getters still allows models to be saved', () => {
```
hem-brahmbhatt, line 6: @Cooryd added this test for you

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -140,6 +141,117 @@ describe('Manuscript', () => {
...
+
+    describe('given there are multiple files', () => {
+      let manuscript
+      let file1, file2
+      let setStatusOfFirstFile, setStatusOfSecondFile
+
+      beforeEach(async () => {
+        manuscript = await createInitialManuscript(userId)
+        manuscript = await addFileToManuscript(manuscript)
+        manuscript = await addFileToManuscript(manuscript);
```
hem-brahmbhatt, line 198: I think it would be nice to move these test helpers to a better location so we can reuse them. We've got some similar helpers within `model.test.js` and within this file.

👍 if you want me to raise an issue for it

server/xpub-model/entities/file/index.test.js
```diff
@@ -11,6 +11,31 @@ describe('File', () => {
...
+        createdBy: userId,
+      }).save()
+      const file = await new File({
+        manuscriptId: manuscript.id,
+        filename: 'thisfile.txt',
+        url: '/an/url',
+      }).save()
+      expect(file.status).toBe('CREATED')
+    })
+  })
```
hem-brahmbhatt, line 37: I added these tests to make it clear that the status of a file only gets set when `.save()` is called. The status is set to `undefined` when you create a `new File()`, even though we have a default value within the model schema.

This doesn't affect `manuscript.fileStatus` since `manuscript.files` is setup via a relation and hence only gets populated from the database.

server/xpub-model/entities/manuscript/index.test.js
SemraSh: I don't know how the upcoming re-architecture is going to effect our helpers, so may be you should create the issue but noting that it was created before the re-architecture work is done...

server/xpub-model/entities/manuscript/index.test.js
```diff
@@ -140,6 +141,117 @@ describe('Manuscript', () => {
     })
   })
 
+  describe('get fileStatus()', () => {
+    describe('given there are no files', () => {
+      it('returns READY', () => {
```
hem-brahmbhatt, line 146: I've left the implementation like this when there's no files on the manuscript. 

It seems like there could be another state here like `NONE` but this should work as is. 

If people feel strongly about it, leave a 👍 and I'll raise a separate issue.

diversemix (approved):
Great!

server/xpub-model/entities/manuscript/index.test.js
diversemix: I prefer it as written. 👍

diversemix: I prefer it as written. 👍

diversemix: I prefer it as written. 👍

SemraSh (approved):
This might the most covered functionality 🥇

Cooryd (approved):

server/xpub-model/entities/manuscript/index.test.js
hem-brahmbhatt: https://github.com/elifesciences/elife-xpub/issues/1692

Issue #1378, opened 2019-01-22
Show an error state when supporting file uploads are cancelled (by page reload)


chris-huggins:
## Problem / Motivation

- WHO author completing an initial submission
- WHEN and WHERE when uploading one or more supporting files and reloading the page
- WHAT files cancelled by the reload should be shown in an error state
- WHY do they want this? So that they are aware the files will not make it to eJP

## Proposed solution

Show an error state

## Clarification needed and assumptions

- Are we definitely unable to improve this for MVP by allowing the errors to be removed without restarting file upload?

## User interface / Wireframes

**Cancelled files are shown in a generic error state**

![image](https://user-images.githubusercontent.com/41907215/51553853-cf507000-1e6b-11e9-92fc-4b493cb401d1.png)

**below 768px**

![image](https://user-images.githubusercontent.com/41907215/51553935-f8710080-1e6b-11e9-9ec8-4e7903ca5d36.png)

hdrury1:
Note that for MVP we will not be introducing the trash icon. A user will remove all items and start again.

I've raised a ticket for improving the feature (adding the trash icon and functionality) here: #1382

diversemix:
@hdrury1 is this a duplicate of #1362 ?

hdrury1:
@diversemix I thought it was but after discussion the other day didn't we determine this ticket encompassed the front end work and the other the server side?

diversemix:
Ah ok ... so the server now needs to not delete the file entity when it fails? (Otherwise they won't show up) is that correct @hdrury1 ?

hdrury1:
I guess fulfilling this ticket will also solve the bug :)

PR #1669, opened 2019-03-12
Adding validation for fileStatus on client side

jyenes:
#### Background
whereas there are files  ongoing uploading,  a bug has been detected that allows the user to navigate to the next page, 

What does this PR do?
adds validation on the Schema to avoid the navigation in case the upload is still ongoing, based on virtual field fileStatus see #1469 
Closes #1669 

depends on #1469
- [x] Browser tests

pending task:

- [x] block the navigation depending on the field fileStatus
- [x] display the validation error
- [x] integrate with #1469

### DIFF
--- a/app/components/pages/SubmissionWizard/SubmissionOperations.js
+++ b/app/components/pages/SubmissionWizard/SubmissionOperations.js
@@ -25,6 +25,7 @@ function runMutation(formValues, mutation, { refetchQueries = [] } = {}) {
     'teams',
     'status',
     'clientStatus',
+    'fileStatus',
     'lastStepVisited',
   ]
 
--- a/app/components/pages/SubmissionWizard/operations.js
+++ b/app/components/pages/SubmissionWizard/operations.js
@@ -65,6 +65,7 @@ const manuscriptFragment = gql`
     cosubmission
     submitterSignature
     disclosureConsent
+    fileStatus
   }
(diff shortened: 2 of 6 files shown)


Cooryd (approved):

hem-brahmbhatt (approved):
Looks good! I tested the following scenarios in the review environment and they all work:

- Validation appears during manuscript upload
- Validation appears during supporting file upload
- Validation appears during both manuscript and supporting file upload
- Validation appears after one file has finished uploading but other files are still uploading
- Can press next after all of the above scenarios finish uploading completely
- Refreshing the page during manuscript upload allows user to upload again
- Refreshing the page during multiple supporting files uploading causes the uploads to be cancelled, but a single file still shows as uploading, and the validation appears. The user can continue with the submission if they then click 'Remove all supporting files'. 

@hdrury1 the last scenario is an existing bug in Production, just without the validation step.

diversemix (approved):
Looks great!

hdrury1:
@damnhipster I've taken a look at this properly and was able to recreate the bug. Correct me if I'm wrong but we haven't yet implemented the validation error state which shows the cancelled upload on refresh in the client?

This would be done as part of https://github.com/elifesciences/elife-xpub/issues/1378

I can't speak for what is going on in the backend... presumably nothing has been added to the database?

PR #1706, opened 2019-03-19
1682 cookie notice

### LINKED ISSUE #1682: Implement the cookie notice
Now that you can get to xPub directly we need to implement the cookie notice.

The cookie covers all subdomains of `elifesciences.org`  and the hiding persists across these because of how the cookie is set.

Implementation can be found here:
https://github.com/elifesciences/pattern-library/blob/develop/assets/js/components/CookieOverlay.js

/cc @hdrury1 @davidcmoulton

Cooryd:
#### Background

Adds a cookie notice to the footer of the screen that won't show once accepted.

#### Any relevant tickets

closes #1682

- [x] Browser tests

#### UI Changes


app/components/global/CookieNotice/index.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import { Box } from '@rebass/grid'
+import { Button } from '@pubsweet/ui'
+import FooterMessage from 'ui/atoms/FooterMessage'
+import NativeLink from 'ui/atoms/NativeLink'
+import Helpers from './helpers'
+
```
diversemix, line 7: Just a suggestion ... do we want to reference that this cookie is the one from Journal? @Cooryd

diversemix (approved):

app/components/global/CookieNotice/index.js
Cooryd: I'm not sure I can use the same cookie across both sites as they're on different domains. This one is specific to reviewer.elifesciences.org

hdrury1:
FYI, see below for our implementation:

Screen Shot 2019-03-19 at 13 17 18

vs Journal implementation:

Screen Shot 2019-03-19 at 13 17 25

app/components/global/CookieNotice/index.js
diversemix: Ah what I meant was via some sort of comment in way of explanation to a dev.

Cooryd: I'll leave a comment that the design and wording are copied from journal

hdrury1:
Is the sizing difference due to a minimum button width we have set that Journal doesn't?

hdrury1 (approved):

hdrury1 (approved):

Issue #1676, opened 2019-03-14
Add Audit logs for application start / stop


diversemix:
This will need new enumerations in the database.

The aim is to ensure we audit application start /stop ... value should contain the hostname and any other unique id (pid?) to fully identify which application.

diversemix:
Blocked on https://gitlab.coko.foundation/pubsweet/pubsweet/issues/443

/cc @hdrury1 @erezmus

erezmus:
So from my understanding, it seems that when you do docker-compose down or kill the signal is only sent to PID 1 which in our case is the 'sh' process. So the signal will not be passed on to pubsweet or our xpub-server.

I tried with having this in the docker-compose.override.yml
`command: ["npx", "pubsweet", "server"]
`

Seems to do the trick if i add a 
`process.on('SIGTERM', () => {...})
`

in the cli/server.js file in pubsweet code.

erezmus:
A bit more information according to docker documentation (https://docs.docker.com/engine/reference/run/#foreground):

> Note: A process running as PID 1 inside a container is treated specially by Linux: it ignores any signal with the default action. So, the process will not terminate on SIGINT or SIGTERM unless it is coded to do so.

So essentially whichever program we run in the command setting in docker-compose file will need to handle the signal itself

diversemix:
So we need to ....
- Update the PubSweet code to pass on any signals
- Have a wrapper script to handle the signals

diversemix:
There is a branch for this - the PR is so old that its being closed

PR #1729, opened 2019-03-21
[WIP] Add wrapper script

### LINKED ISSUE #1724: Add a wrapper script to handle signals
When running on a docker container, the /bin/sh is the main process with PID 1. According to docker documentation (https://docs.docker.com/engine/reference/run/#foreground)

> Note: A process running as PID 1 inside a container is treated specially by Linux: it ignores any signal with the default action. So, the process will not terminate on SIGINT or SIGTERM unless it is coded to do so.

Hence we need to write a wrapper script to replace the current command in docker-compose script that will:
- wrap the execution of the wait-for-it.sh script
- execute the npx pubsweet server command
- listen and pass along SIGTERM to the node process

Found a blog post from codeship that should help:
https://blog.codeship.com/trapping-signals-in-docker-containers/

This ticket follows from #1676

erezmus:
#### Background

Adds a wrapper script around wait-for-it.sh and the pubsweet start command so that termination signals are handled in a container environment and can be sent to the pubsweet node process.

#### Any relevant tickets

closes #1724 

Tests need to be done manually, checking that entry for stopping application is present in the audit log when running `docker-compose down` or `docker-compose kill -s SIGTERM app`

### DIFF
--- a/server/xpub-model/entities/auditLog/index.js
+++ b/server/xpub-model/entities/auditLog/index.js
@@ -12,7 +12,7 @@ class AuditLog extends BaseModel {
         userId: { type: 'uuid' }, // Currently not used
         action: {
           type: 'string',
-          enum: ['CREATED', 'UPDATED', 'DELETED', 'MECA_RESULT'],
+          enum: ['CREATED', 'UPDATED', 'DELETED', 'MECA_RESULT', 'APPLICATION'],
         },
         value: { type: ['string', 'null'] },
         objectId: { type: 'uuid' },
--- a/server/xpub-model/migrations/1552580166-app-audit.sql
+++ b/server/xpub-model/migrations/1552580166-app-audit.sql
@@ -1,0 +1,4 @@
+ALTER TYPE auditaction RENAME TO auditaction_old;
+CREATE TYPE auditaction AS ENUM( 'CREATED', 'UPDATED', 'DELETED', 'MECA_RESULT', 'APPLICATION' );
+ALTER TABLE audit_log ALTER COLUMN action TYPE auditaction USING action::text::auditaction;
+DROP TYPE auditaction_old;
--- a/server/xpub-server/index.js
+++ b/server/xpub-server/index.js
@@ -2,7 +2,8 @@ const os = require('os')
 const { merge } = require('lodash')
 const fs = require('fs')
 const logger = require('@pubsweet/logger')
-const cleanup = require('./cleanup')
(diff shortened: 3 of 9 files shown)


server/xpub-model/entities/auditLog/index.js
```diff
@@ -12,7 +12,7 @@ class AuditLog extends BaseModel {
         userId: { type: 'uuid' }, // Currently not used
         action: {
           type: 'string',
-          enum: ['CREATED', 'UPDATED', 'DELETED', 'MECA_RESULT'],
+          enum: ['CREATED', 'UPDATED', 'DELETED', 'MECA_RESULT', 'APPLICATION'],
```
diversemix, line 15: You probably committed this by mistake - this is to do with the AuditLog changes

server/xpub-model/migrations/1552580166-app-audit.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TYPE auditaction RENAME TO auditaction_old;
```
diversemix, line 1: You probably committed this by mistake - this is to do with the AuditLog changes

server/xpub-server/index.js
```diff
@@ -38,15 +39,30 @@ const registerRoutes = app => {
 const appMessage = action =>
   `${action} Application: ${os.hostname()}, PID: ${process.pid}`
 
-const stopServer = () => {
+const auditApplicationEvent = async message => {
```
diversemix, line 42: You probably committed this by mistake - this is to do with the AuditLog changes

server/xpub-server/index.js
```diff
@@ -38,15 +39,30 @@ const registerRoutes = app => {
...
 
-logger.info(appMessage('Starting'))
+const stopServer = () => xpubCleanup
+const msg = appMessage('Starting')
+logger.info(msg)
+auditApplicationEvent(msg)
 
 // eslint-disable-next-line no-unused-vars
-const cleanupHandler = cleanup.Cleanup(process, logger, stopServer)
+const cleanupHandler = cleanup(process, logger, stopServer)
```
diversemix, line 65: Keep this though

diversemix (changes requested):
Be good to extract the AuditLog changes and leave them as part of #1676

erezmus:
Closing as the branch was based on 1676-app-audit instead of develop. Will create a new PR

PR #1732, opened 2019-03-21
Dashboard Component

### LINKED ISSUE #1718: Dashboard Component: Move code into component-dashboard
## Background

As part of #1705 we plan to phase the refactoring of the app code. Round 1 involves moving the code into a `component-xxxx` directory structure and should not require the changing of much code. This issue relates to moving / grouping the dashboard components

## Tasks

- [x] create `component-dashboard` directory and package files (see https://github.com/elifesciences/elife-xpub/issues/1711)
- [ ] move all client code into a `client` directory within `component-dashboard`
- [ ] move all server code into a `server` directory within `component-dashboard`
- [ ] hook up newly moved code to the app and make sure there are no remaining files in the wider codebase relating to this.

__Are there any dashboard related tests left over in the wider code base after? How will we deal with that?__

diversemix:
#### Background
This sets an example for the rest of the refactor work. 
Creating of dashboard component by moving relevant code into the component.

#### Any relevant tickets

Closes #1718

New tests written
Configured jest to work with new client/server and old code.


packages/component-dashboard/client/components/ManuscriptsQuery.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { Query } from 'react-apollo'
-import Loading from '../../ui/atoms/Loading'
+import Loading from 'ui/atoms//Loading'
```
Cooryd: `//`

packages/component-dashboard/client/components/DashboardList.js
```diff
@@ -2,9 +2,9 @@ import React from 'react'
 import styled from 'styled-components'
 import { Box } from '@rebass/grid'
 import { th } from '@pubsweet/ui-toolkit'
+import Paragraph from 'ui/atoms/Paragraph'
 
-import DashboardListItem from 'ui/molecules/DashboardListItem'
-import Paragraph from '../atoms/Paragraph'
+import DashboardListItem from './DashboardListItem'
```
diversemix, line 7: I didn't realise that we moved things out of our ui folder ... was that intentional?

will-byrne: I believe anything unique to the package is moved with it.

diversemix: Yeah, think that's right.... especially given it has the word "Dashboard" in its name :laughing:

.eslintrc.js
```diff
@@ -0,0 +1,53 @@
...
+      },
+      {
+        enforceForRenamedProperties: false,
+      },
+    ],
+    'prettier/prettier': 0,
+    'react/prop-types': 0,
+    'sort-keys': 0,
+  },
+}
```
hem-brahmbhatt: The eslint changes should probably be it's own PR.

👍 if you're happy for me to do this

packages/component-dashboard/client/graphql/mutations.js
```diff
@@ -0,0 +1,18 @@
+import gql from 'graphql-tag'
+// TODO: component-dashboard needs its own model or a shared gql fragment for manuscript
+import { manuscriptFragment } from '../../../../app/components/pages/SubmissionWizard/operations'
```
hem-brahmbhatt, line 3: I think we should create an issue to break this dependency and work on it soon. This will come up again as a problem for other components we create and tackling this sooner will only serve to make it easier in the future.

packages/jest.config.client.js
```diff
@@ -0,0 +1,12 @@
...
+  rootDir: '../',
+  setupTestFrameworkScriptFile: '<rootDir>/packages/jest-setup.client.js',
+  testMatch: ['<rootDir>/packages/**/client/**/*.test.js'],
+  transformIgnorePatterns: ['/node_modules/(?!@?pubsweet|xpub)'],
+  moduleNameMapper: {
+    '\\.s?css$': 'identity-obj-proxy',
+    '^ui(.*)$': '<rootDir>/app/components/ui$1',
+    '^global(.*)$': '<rootDir>/app/components/global$1',
+  },
+}
```
hem-brahmbhatt: Not sure I'm a fan of config files like this living outside of components. If the component itself has tests, I feel that the component should have all the files and modules necessary for it to run as it's own package.

I'd rather have the jest files duplicated across the different packages for the above reason.

.eslintrc.js
Cooryd: I'll chuck this in a PR now

packages/jest.config.client.js
Cooryd: A fair point, we put these here to make it as easy as possible to add new components in without the need to create new config and set-up files for both the server and client for every component. This creates a problem though as context specific test set up and config requirements will need to be placed in these top level files, so we'll no longer have our contexts completely stand alone. 

Just a note though, `ui` and `global` shouldn't exist anymore after the refactor as they'll sit in a `@elifesciences/component-elife-ui` package of their own.

diversemix: Does this mean we should move these first then?

.eslintrc.js
Cooryd: https://github.com/elifesciences/elife-xpub/pull/1737

packages/jest.config.client.js
erezmus: @cooryd is this what you mean: 

if we have jest setup and jest config files inside each component's client and server folder, then the issue is that the rootDir is then at that package's level.

So for instance, if we have in package.json:
```
"jest": {
//...
   projects: [
      //...,
      "<rootDir>/packages/component-dashboard/client/jest.config.js",
      //...
   ]
}
```

then the rootDir becomes elife-xpub/packages/component-dashboard

which then breaks the moduleNameMapper setting as that uses rootDir, so we would need to add some '../' but thats doable until ui and global go to their own component?

erezmus: I've done a quick test with the following config in packages/component-dashboard/client/jest.config.js:

```
module.exports = {
    displayName: 'package-client',
    rootDir: '../',
    setupTestFrameworkScriptFile: '<rootDir>/client/jest-setup.js',
    testMatch: ['<rootDir>/client/**/*.test.js'],
    transformIgnorePatterns: ['/node_modules/(?!@?pubsweet|xpub)'],
    moduleNameMapper: {
      '\\.s?css$': 'identity-obj-proxy',
      '^ui(.*)$': '<rootDir>/../../app/components/ui$1',
      '^global(.*)$': '<rootDir>/../../app/components/global$1'
    }
  }
```

seems to work

hem-brahmbhatt: Yes, I was trying it out locally and it seems to work. Happy for this to go into a separate issue/PR.

packages/component-dashboard/client/graphql/mutations.js
hem-brahmbhatt: Addressed by #1738

hem-brahmbhatt (approved):

jyenes (approved):

packages/jest.config.client.js
Cooryd: I've created a `/component-dashboard/config` directory that holds the 2 config and 2 setup files for the component. Tests all pass for me and keeps the config in a central place within the component

PR #1698, opened 2019-03-18
[ON HOLD] Audit log application start / stop

diversemix:
#### Background

Ensure that application start and stop is audited.

#### Any relevant tickets

Closes #1676

#### How has this been tested?

New tests added.
Still to test with `docker-compose --force-recreate`

PR #1733, opened 2019-03-21
[ON HOLD] Added server-start.sh wrapper scripts to allow for graceful termination

### LINKED ISSUE #1724: Add a wrapper script to handle signals

erezmus:
#### Background
Adds a wrapper script around wait-for-it.sh and the pubsweet start command so that termination signals are handled in a container environment and can be sent to the pubsweet node process.

#### Any relevant tickets
closes #1724

Tests need to be done manually, checking that entry for stopping application is present in the audit log when running docker-compose down or docker-compose kill -s SIGTERM app

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,7 +4,7 @@ services:
   app:
     build:
       dockerfile: ./Dockerfile-development
-    command: sh -c "yarn install --frozen-lockfile && ./scripts/wait-for-it.sh postgres:5432 -s -t 40 -- npx pubsweet server"
+      command: ["./scripts/server-start.sh"]
     volumes:
       - ./:/home/xpub
     depends_on:
--- a/scripts/server-start.sh
+++ b/scripts/server-start.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# wrapper script for wait-for-it.sh and pubsweet server
+
+pid=0
+
+sigterm_handler () {
+  if [ $pid -ne 0 ]; then
+    kill -SIGTERM "$pid"
+    wait "$pid"
+  fi
+  exit 143;
+}
+
+trap 'kill ${!}; sigterm_handler' SIGTERM
+
+./scripts/wait-for-it.sh postgres:5432 -s -t 40
+
+npx pubsweet server &
+pid="$!"
+
+while true
+do
+  tail -f /dev/null & wait ${!}
+done


diversemix (approved):

hem-brahmbhatt (commented):
Audit log entries can't be tested until https://github.com/elifesciences/elife-xpub/pull/1698 is merged.

I'm not sure how to test the `docker-compose.override.yml` locally, and it seems that the file has references to development, so I don't think these changes will apply in staging or production.

Our review environments don't use docker compose, as can be seen [here](https://github.com/elifesciences/elife-xpub-formula/blob/master/helm/elife-xpub/templates/deployment.yaml#L64), so this can't be tested there.

scripts/server-start.sh
```diff
@@ -0,0 +1,24 @@
...
+  if [ $pid -ne 0 ]; then
+    kill -SIGTERM "$pid"
+    wait "$pid"
+  fi
+  exit 143;
+}
+
+trap 'kill ${!}; sigterm_handler' SIGTERM
+
+./scripts/wait-for-it.sh postgres:5432 -s -t 40
```
giorgiosironi, line 16: the hardcoded `postgres` stops this from being used in production/staging, should use `PGHOST` and `PGPORT` (making sure they are both specified)

docker-compose.override.yml
```diff
@@ -4,7 +4,7 @@ services:
   app:
     build:
       dockerfile: ./Dockerfile-development
-    command: sh -c "yarn install --frozen-lockfile && ./scripts/wait-for-it.sh postgres:5432 -s -t 40 -- npx pubsweet server"
```
giorgiosironi, line 7: I don't know what the `yarn` command does but I think it was necessary in local use to apply local changes to `server/` and similar on top of the Docker image?

scripts/server-start.sh
erezmus: I don't think i understand enough about the staging/production environment. Maybe we can chat on slack?

diversemix: Yeah ... let's talk f2f. IMHO ... we have far too many ways / environments in which the app is started - and the different ways are spread all over the place. At the very least we should document each of these. There are some docs in `docs/deploying` but certainly not enough to fully understand what is going on.

PR #1737, opened 2019-03-22
Make linting config file js format and update docker deployment file …

Cooryd:
#### Background

For the new re-architecture work we need the import resolver eslint config to no longer be relative. To do this we need to change the config file to a js file so we can append `__dirname`. This changes the deployment docker file to use this new file too. 

#### Any relevant tickets

Get this merged in before #1732


diversemix (approved):

erezmus (approved):
looks good to me

Issue #1721, opened 2019-03-20
Submission Component: Move code into component-submission


Cooryd:
## Background

As part of #1705 we plan to phase the refactoring of the app code. Round 1 involves moving the code into a `component-xxxx` directory structure and should not require the changing of much code. This issue relates to moving / grouping the submission wizard components

## Tasks

- [x] create `component-submission` directory and package files (see https://github.com/elifesciences/elife-xpub/issues/1711)
- [x] move all client code into a `client` directory within `component-submission`
- [x] move all server code into a `server` directory within `component-submission`
- [x] hook up newly moved code to the app and make sure there are no remaining files in the wider codebase relating to this.

__Are any submission related tests left over in the wider code base after? How will we deal with that?__

diversemix:
Going to close this - risking that there may still be stuff to move in (like `fake-s3` ) ... but this is the bulk of the work.

PR #1739, opened 2019-03-22
Created folder structure for component-submission

erezmus:
#### Background

Creates the basic folder structure for the submission component

#### Any relevant tickets

Relates to #1721 

No tests needed

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -15,6 +15,7 @@ module.exports = {
       '@elifesciences/xpub-controller',
       '@elifesciences/xpub-client-config',
       '@elifesciences/component-dashboard',
+      '@elifesciences/component-submission',
       '@pubsweet/component-send-email',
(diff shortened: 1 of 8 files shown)


diversemix (approved):

hem-brahmbhatt (approved):

Cooryd:
Coverage % dropped, should we potentially turn that off until we've finished?

erezmus:
> Coverage % dropped, should we potentially turn that off until we've finished?

Gonna modify the collectCoverageFrom array in package.json to ignore the jest files when collecting coverage.

hem-brahmbhatt:
Also, we shouldn't be seeing any drop in coverage since we're moving code. In this case, it was a valid failure due to some of our config files not being ignored from coverage when they should have been.

Issue #1720, opened 2019-03-20
Static Pages Component: Move code into component-staticPages 


Cooryd:
## Background

As part of #1705 we plan to phase the refactoring of the app code. Round 1 involves moving the code into a `component-xxxx` directory structure and should not require the changing of much code. This issue relates to moving / grouping the static pages components

## Tasks

- [ ] create `component-staticPages` directory and package files (see https://github.com/elifesciences/elife-xpub/issues/1711)
- [ ] move all client code into a `client` directory within `component-staticPages`
- [ ] move all server code into a `server` directory within `component-staticPages`
- [ ] hook up newly moved code to the app and make sure there are no remaining files in the wider codebase relating to this.

__Are any staticPages related tests left over in the wider code base after? How will we deal with that?__

Issue #1711, opened 2019-03-20
Dashboard Component: Create component directory and package files


Cooryd:
## Background

As part of #1705 we need to create a Dashboard component package that will group both server and client components for the apps dashboard to be used by the Pubsweet app.

## Task

Create a package directory that will sit in a top level `packages` directory called `component-dashboard` this will need to have an `index.js` and a `package.json` files should look like so:

index.js 

```js
module.exports = { }
```

package.json

```json
{
  "name": "@elifesciences/component-dashboard",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT"
}
```

This package should then be added to the pubsweet config found in `config/default.js`

default.js

```
...
  pubsweet: {
    components: [
      '@elifesciences/component-dashboard',
      ...
    ],
  },
...
```

Cooryd:
This task should be done as part of the move tasks of this milestone. I'll close this then reference it within those issues.

PR #1742, opened 2019-03-22
Moved static pages into component-static-pages folder

will-byrne:
### Background
Moving the static pages to `packages/component-static-pages` 

This PR is only moving the static pages, we will move the components that are related only to static pages in another pr.

### Any relevant tickets
Partial work on #1720 

### Tasks

- [x] create `component-static-pages` directory and package files (see https://github.com/elifesciences/elife-xpub/issues/1711)
- [x] create jest configuration file
- [x] move all client code into a `client` directory within `component-static-pages`
- [x] hook up newly moved code to the app and make sure there are no remaining files in the wider codebase relating to this.


hem-brahmbhatt (approved):
Code looks good. Just need to check things in the review environment when the build has finished.

diversemix (approved):

Issue #1735, opened 2019-03-22
Elife UI Components: move ui & global components to elife-ui-component


Cooryd:
## Background
As part of #1705 we are moving code into package components. Once all of the models and bound contexts have been moved we will likely still have some left over client side shared components most likely in the `/ui` and `/components/global` directories. These should be shared presentational components such as layout wrappers and styled buttons ect. These shared components should be moved into their own `/packages` component called `component-elife-ui`. Once this is done, we can get rid of the `ui` and `global` alias that is being used as well as the jest and webpack config set for them.

## Tasks 

- [x] move code into component directory
- [x] change all import paths to use the new `@elifesciences/component-elife-ui` package
- [x] remove references to `ui` and `global` alias from config files

PR #1745, opened 2019-03-22
Moved client files for submission component

erezmus:
#### Background

Moved the client files for the submission component

#### Any relevant tickets

Relates to #1721


packages/component-submission/client/SubmissionOperations.js
```diff
@@ -2,15 +2,15 @@ import { cloneDeep } from 'lodash'
 import { graphql, compose } from 'react-apollo'
 import React from 'react'
 import omitDeep from 'omit-deep-lodash'
-import Loading from 'ui/atoms/Loading'
-// TODO: This should not be importing from the other component
 import { ALL_MANUSCRIPTS } from '@elifesciences/component-dashboard/client/graphql/queries'
+import Loading from '../../../app/components/ui/atoms/Loading'
```
will-byrne: does this need to change? would the ui alias not still work here?

packages/component-submission/client/SubmissionOperations.js
```diff
@@ -2,15 +2,15 @@ import { cloneDeep } from 'lodash'
...
+// TODO: This should not be importing from the other component
 import {
   GET_MANUSCRIPT,
   UPDATE_MANUSCRIPT,
   SUBMIT_MANUSCRIPT,
 } from './operations'
-import ErrorPage from '../Error'
+import ErrorPage from '../../../app/components/pages/Error'
```
hem-brahmbhatt, line 13: This will need to be decoupled at some point. Will create a separate issue for it.

hem-brahmbhatt: Address by #1748

packages/component-submission/client/SubmissionOperations.js
hem-brahmbhatt: It doesn't need to change, the alias should work just fine. If we're continuing to use the alias in our other components, we should keep things consistent and do the same here.

It will make the work for https://github.com/elifesciences/elife-xpub/issues/1735 easier.

SemraSh (changes requested):
I agree Will and Hem on importing `ui` and `global` as they are aliases. We should keep consistency importing those.

SemraSh (approved):

will-byrne (approved):

hem-brahmbhatt:
Let's look at this in the review environment before we merge, since this is basically the product we're refactoring at the moment.

diversemix:
> Let's look at this in the review environment before we merge, since this is basically the product we're refactoring at the moment.

Well just the client side at the moment.

hem-brahmbhatt (approved):

PR #1756, opened 2019-03-25
Moved the the ui folder

### LINKED ISSUE #1735: Elife UI Components: move ui & global components to elife-ui-component

jyenes:
#### Background
move the UI components to its own folder

- [x] move UI folder
- [x] move global folder

closes #1735


SemraSh (approved):
Looks good to me

will-byrne (approved):

Issue #1722, opened 2019-03-20
Model Components: move Manuscript model into its own component


Cooryd:
## Background

As part of #1705 we plan to phase the refactoring of the app code. Round 1 involves moving the code into a `component-xxxx` directory structure and should not require the changing of much code. This issue relates to moving / grouping the model components

## Tasks

- [ ] for each model create `component-model-xxx` directory and package files (see https://github.com/elifesciences/elife-xpub/issues/1711)

- [ ] move `Manuscript` model into `component-model-manuscript`
- [ ] move `File` model into `component-model-file`
- [ ] move `Audit` model into `component-model-audit`
- [ ] move `Identity` model into `component-model-identity`
- [ ] move `SemanticExtraction` model into `component-model-semanticExtraction`
- [ ] move `Team` model into `component-model-team`
- [ ] move `User` model into `component-model-user`

- [ ] hook up newly moved code to the app and make sure there are no remaining files in the wider codebase relating to this.

__Are any model related tests left over in the wider code base after? How will we deal with that?__

PR #1758, opened 2019-03-25
1722 move models into components

will-byrne:
#### Background

Moving all of the models into `component-model`

#### Any relevant tickets

Partial work of #1722 , we will split each model to their own `component-model-xxxx` after this pr is merged.


config/default.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   pubsweet: {
     components: [
       '@elifesciences/xpub-meca-export',
-      '@elifesciences/xpub-model',
+      '@elifesciences/component-model',
```
jyenes, line 13: not sure if we should rename the xpub-model

packages/component-login/client/index.js
```diff
@@ -67,6 +67,7 @@ class LoginPage extends React.Component {
         LoginPage.setToken(null)
         // TODO expose this error once we have a UI to do so
         this.props.history.push('/login', { error: err.message })
+        // eslint-disable-next-line no-console
```
jyenes, line 70: why not remove the console.log?
if you need to log anything important, use logger

packages/component-login/client/index.test.js
```diff
@@ -62,6 +62,7 @@ describe('LoginPage component', () => {
 
     await new Promise(resolve => setTimeout(resolve, 0))
     expect(historyMock.push).toHaveBeenCalledWith('/login', { error: 'Nope' })
+    // eslint-disable-next-line no-console
     expect(console.error).toHaveBeenCalled()
```
jyenes, line 66: same here

jyenes (changes requested):

packages/component-login/client/index.js
will-byrne: will need to change this in a separate PR and add the login test configs.

jyenes (approved):

erezmus (approved):

PR #1762, opened 2019-03-25
Fix: add configuration to component-login

SemraSh:
#### Background
component-login was introduces without jest configuration so the tests were not running for this part of the code.

#### Any relevant tickets
We haven't created a ticket as this was a quick fix

### DIFF
--- a/packages/component-login/client/index.js
+++ b/packages/component-login/client/index.js
@@ -5,7 +5,6 @@ import { withApollo } from 'react-apollo'
 import gql from 'graphql-tag'
 import config from 'config'
 import { H1 } from '@pubsweet/ui'
-import logger from '@pubsweet/logger'
 import styled, { css } from 'styled-components'
 import ButtonLink from 'ui/atoms/ButtonLink'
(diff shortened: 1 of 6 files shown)


diversemix (approved):

erezmus (approved):

jyenes (approved):

packages/component-login/client/index.js
```diff
@@ -68,7 +67,8 @@ class LoginPage extends React.Component {
         LoginPage.setToken(null)
         // TODO expose this error once we have a UI to do so
         this.props.history.push('/login', { error: err.message })
-        logger.error(err)
+        // eslint-disable-next-line no-console
+        console.error(err)
```
diversemix: Why has this been reverted @SemraSh ?

Cooryd (approved):

packages/component-login/client/index.js
SemraSh: Because `logger` is available only in the server and it is not going to be safe to make it public.

PR #1763, opened 2019-03-25
1753 Configure styleguide for modules

### LINKED ISSUE #1753: Add styleguide sections for new packages
We need to add a new section inside packages for every component we create. 
Below is the one we have added for `component-static-pages`

```
  {
      name: 'Packages',
      sections: [
        {
          name: 'Static Pages',
          sections: [
            {
              name: 'Components',
              components:
                'packages/component-static-pages/client/components/**/*.js',
            },
            {
              name: 'Pages',
              components:
                'packages/component-static-pages/client/pages/**/*.js',
            },
          ],
        },
      ],
    },
```

SemraSh:
#### Background
Adding a new section inside packages for every component we create.

#### Any relevant tickets

Closes #1753 

we've manually run `yarn styleguide` and checked if the components are shown in the styleguide


diversemix (approved):

app/routes.js
```diff
@@ -9,7 +9,7 @@ import ErrorBoundary from 'global/ErrorBoundary'
 import CookieNotice from 'global/CookieNotice'
 
 import LandingRedirect from './components/pages/LandingRedirect'
-import LoginPage from '../packages/component-login/client'
+import LoginPage from '../packages/component-login/client/LoginPage'
```
Cooryd: What's the purpose of the file rename? Shouldn't this file live within the `/pages` client directory and be exported in the `index.js`? The imports on line 14 - 18 don't match how we originally did this for the Dashboard page (see line 5 above)

Cooryd: I've just seen that the bellow file imports haven't been turned into packages yet. I think we should still be moving this LoginPage to a `/pages` sub-directory

jyenes (approved):

PR #1774, opened 2019-03-27
only ask sciencebeam to extract the title

### LINKED ISSUE #1773: Use ScienceBeam `include` parameter
## Subtasks

- [ ] `include` parameter set to `title` for ScienceBeam requests

## Assumptions and notes
- The time to extract might vary depending on what ScienceBeam is asked to extract
- For that reason the `include` parameter was introduced (https://github.com/elifesciences/sciencebeam/issues/51)
- To extract only the title, add `include=title`, e.g. `POST /api/convert?include=title&filename=file.pdf`
- The effect will be that GROBID is only executing the header model and it is also only parsing the first three pages (assuming the title is found there)
- (The file could also be reduced before sending it to ScienceBeam)

/cc @diversemix

diversemix:
#### Background

With only asking sciencebeam to extract the title this will make the uploading more predictable.

#### Any relevant tickets

Closes #1773

Tests updated to reflect the inclusion of the -include- parameter


de-code (approved):
(no idea about the binary files)

erezmus (approved):

PR #1788, opened 2019-03-27
1776 app structure

Cooryd:
This brings assets into the top level `/app` directory and creates a client directory for the app to store any general app specific pages and components. 

**Note: Styleguide seems to have broken its asset path replace at some point during the refactor. This PR wasn't the cause of this break however it was identified when making sure the moved assets were being ported over correctly. This will need fixing as part of another issue**


Dockerfile
```diff
@@ -37,7 +37,6 @@ COPY \
 COPY app app
 COPY config config
 COPY scripts scripts
-COPY assets assets
 COPY styleguide styleguide
```
diversemix, line 40: Can we not change this to `COPY app/assets assets` ?

diversemix (approved):

erezmus (approved):

Dockerfile
erezmus: Isn't this handled with COPY app app

jyenes (approved):

Dockerfile
Cooryd: That would require lots of webpack includes paths changed too but yes we can if theres a benefit to these existing separately on the server.

PR #1796, opened 2019-03-28
Move Appbar and its layout wrapper into the app directory

### LINKED ISSUE #1777: Move AppBar from elife-ui into app/
Move into app/client/components

Cooryd:
#### Background

This moves the app bar out of the `component-elife-ui` package and into the top level `app` directory. We will likely want to make a completely separate package for this in the future but this is one of the steps needed to clear the `component-elife-ui` package of any non-reusable ui components.

#### Any relevant tickets

Closes #1777

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


app/client/components/Layout.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react'
 import { Box } from '@rebass/grid'
-import Centerer from './Centerer'
-import AppBar from '../AppBar'
+import Centerer from '@elifesciences/component-elife-ui/client/global/layout/Centerer'
+import WrappedAppBar from './WrappedAppBar'
```
diversemix, line 4: Should this be `WrappedTopBar` ?

diversemix (approved):
I like the rename of "AppBar" to "TopBar"

app/client/components/Layout.js
Cooryd: TopBar is just the styled 72px box with a bottom grey border. AppBar includes the profile menu and the navigation menu wrapped inside this styling and WrappedAppBar is just that but with the GQL queries for fetching user details.

Topbar gets used on the redirect page without any of the profile or navigation stuff so they're different but require to same TopBar wrapper.

diversemix: Maybe something for the future would be to come up with a naming convention that would indicate "thing with GQL" / "Just a styled Thing" etc

Cooryd: Plan for the future is to use HOC's for graphql stuff so you'd just wrap the page in a `withGQL(Layout)` wrapper 💯

jyenes (approved):

Issue #1084, opened 2018-11-21
Expose elife-api in the package xpub-model


diversemix:
**complete #1609 first**

See https://github.com/elifesciences/elife-xpub/pull/1065#discussion_r235074668

diversemix:
@Cooryd / @damnhipster ... will this get done as part of #1587 ?

hem-brahmbhatt:
Yes, will include it in the epic.

hem-brahmbhatt:
Currently the model requires `elife-api`. Moving the file into `xpub-controller` means that the model will require the controller which is a code smell.

We can look at moving the clients into their own package alongside `xpub-controller` and `xpub-model` so both of them can make use of clients.

What does everyone think? @diversemix @Cooryd @jyenes

Cooryd:
I like this idea, makes the controller directory tidier too

Cooryd:
See #1609 which creates the top level `xpub-client` (???) directory. This issue just relates to relocating the `elife-api` to that directory

PR #1814, opened 2019-03-28
moved user model into component-model-user

### LINKED ISSUE #1801: move User model into component-model-use
## Subtasks
- [x] move `User` model into `component-model-user`

will-byrne:
#### Background

Move the user model into the component-model-user package

#### Any relevant tickets

Closes #1801


packages/component-submission/server/resolvers/index.js
```diff
@@ -1,7 +1,7 @@
 const logger = require('@pubsweet/logger')
-const { User } = require('@elifesciences/component-model')
+const User = require('@elifesciences/component-model-user')
 const Manuscript = require('@elifesciences/component-model-manuscript')
-const elifeApi = require('@elifesciences/component-model/entities/user/helpers/elife-api')
+const elifeApi = require('@elifesciences/component-model-user/entities/user/helpers/elife-api')
```
jyenes, line 4: can not `elifeApi` be exported and then de-structured?

jyenes (changes requested):
A module should have an interface to export whatever is going to be required for external modules to be used.

Cooryd:
https://github.com/elifesciences/elife-xpub/issues/1084 is a legacy issue to move the `elife-api` out. While the issue's out of date the task of moving this out of the model is still something we need to do. I don't think we've figured out where this, meca export functions and scienceBeam really fit in. Their own components? A Providers component? 🤔 This needs discussing

packages/component-submission/server/resolvers/index.js
will-byrne: I think we should leave it as is so that User is the same as the other models and we can discuss what to do with the elifeApi later and raise a separate ticket for it.

jyenes (approved):

diversemix:
@Cooryd ... My take is that its a generic eLife javascript client. As such it probably needs to go into npm as unrelated to xPub - so any js app could use it.

diversemix (approved):

PR #1829, opened 2019-03-29
Move resolvers component user

### LINKED ISSUE #1827: move resolvers for Component-user
this is a subtask of #1824

erezmus:
#### Background

Moves resolvers from server to component-model-user

#### Any relevant tickets

Closes #1827


packages/component-model-user/entities/user/resolvers.js
```diff
@@ -0,0 +1,12 @@
+const User = require('./index')
+
+const resolvers = {
```
Cooryd: I think this should live at the top level of the component along side `/entities`. See https://gitlab.com/hindawi/xpub/xpub-review/tree/develop/packages/component-model-manuscript/src for context we're not using `src` directories.

erezmus: that does make sense

erezmus: moved to top level in 366dd09

packages/component-model-user/resolvers.js
```diff
@@ -0,0 +1,12 @@
+const User = require('./index').model
```
Cooryd: can this go relatively to the entity file rather than through the index? would be good to only have the .model suffix as a temp thing

Cooryd (approved):

SemraSh (approved):

PR #1864, opened 2019-04-02
move logger into logger package and add email filtering.

### LINKED ISSUE #1862: Change logger to filter out emails
We are unable to add a new `component-logger` so this work is being done in `/config`.

Filtering out emails using a regex and a pattern that is extendable.

will-byrne:
#### Background

Added a component-log and moved the logs from config/non-serialisable there.
We are now exporting both the logger and the mock logger.
Added email filtering to the logs so all email addresses will be replaced with `***@***.***`

#### Any relevant tickets

Closes #1862 

- [x] Browser tests

### DIFF
--- a/packages/component-logger/server/logger.js
+++ b/packages/component-logger/server/logger.js
@@ -1,14 +1,14 @@
 const winston = require('winston')
+const emailFilter = require('./filters/emailFilter')
 
-const logPath = process.env.XPUB_LOG_PATH || './'
-
+// production logger sends JSON lines to file
+const logPath = process.env.XPUB_LOG_PATH || '~/elife-xpub/var/logs'
 const logger = new winston.Logger({
   transports: [
     new winston.transports.Console(),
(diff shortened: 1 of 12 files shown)


packages/component-logger/server/logger.js
```diff
@@ -1,12 +1,14 @@
 const winston = require('winston')
+const emailFilter = require('./filters/emailFilter')
 
 // production logger sends JSON lines to file
-const logPath = process.env.XPUB_LOG_PATH || '/srv/elife-xpub/var/logs'
+const logPath = process.env.XPUB_LOG_PATH || '~/elife-xpub/var/logs'
```
diversemix: Are we sure we want to change this default?

diversemix (approved):

packages/component-logger/server/logger.js
will-byrne: Thats a mistake it should be the old one

Cooryd (approved):

PR #1892, opened 2019-04-05
Add tests for configurations

### LINKED ISSUE #1891: Add test coverage for configurations
There is no test coverage for configurations

diversemix:
#### Background

Adds tests for configs

#### Any relevant tickets

Closes #1891

Tests Added


config/__tests__/testConfig.js
```diff
@@ -0,0 +1,21 @@
...
+  let config = _.cloneDeep(configObject)
+  if (ignore) {
+    ignore.forEach(key => _.set(config, key, ''))
+  }
+
+  expect(config.configTag).toBe(configName)
+  expect(config instanceof Object).toBeTruthy()
+  const keys = Object.keys(config)
+  expect(keys.length).toBeGreaterThan(10)
+  expect(config).toMatchSnapshot(configName)
```
giorgiosironi: I would lose the snapshot equality check which is too difficult to keep up to date (and the way to regenerate it is probably just to read the configuration, so it's self-referential rather than providing an independent way to verify the contents). The problem that we experienced was the configuration didn't load at all, so a smoke test that doesn't check the actual value will cover for it.

diversemix:
Closed

PR #1901, opened 2019-04-08
Add secretcheck script and precommit package (need to add multiple secret support)

diversemix:
#### Background

Adds a check to prevent commiting of AWS creds

#### Any relevant tickets

N/A

Locally tested - no tests written

### DIFF
--- a/scripts/secretcheck.sh
+++ b/scripts/secretcheck.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# If the aws credentials file does not exist then just exit ok
+AWS_FILE=~/.aws/credentials
+if [ ! -f ${AWS_FILE} ]
+then
+  echo "No AWS Credentials to check"
+  exit 0
+fi
+
+ROOT=.
+AWS_SECRET=$(grep secret ${AWS_FILE} | cut -d '=' -f2)
+AWS_KEY=$(grep key_id ~/.aws/credentials | cut -d '=' -f2)
+
+SECRETS=(${AWS_SECRET} ${AWS_KEY})
+
+echo "Searching for secrets... "
+
+for var in "${SECRETS[@]}"
+do
+  # Perfrom a seacrch for this secret
+  if [ $(which ag) ]
+  then
+    RESULT=$(ag -l ${AWS_SECRET} ${ROOT})
+  else 
+    echo "   *** if this is slow then install: https://github.com/ggreer/the_silver_searcher#installing"
+    RESULT=$(find ${ROOT} -type f -exec grep -l ${AWS_SECRET} {} \;)
+  fi
+
+  # Action based on finding secret
+  if [ ${#RESULT} -gt 0 ]
+  then
+    echo -n "✖ Found a secret here: "
+    echo ${RESULT}
+    exit 1
+  else
+    echo "✔ Not Found"
+  fi
+
+done
--- a/scripts/secretcheck.sh
+++ b/scripts/secretcheck.sh
@@ -1,37 +1,39 @@
 #!/bin/bash
 
 # If the aws credentials file does not exist then just exit ok
-AWS_FILE=~/.aws/credentials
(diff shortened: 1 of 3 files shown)


will-byrne (approved):

scripts/secretcheck.sh
```diff
@@ -0,0 +1,40 @@
...
+# If the aws credentials file does not exist then just exit ok
+AWS_FILE=~/.aws/credentials
+if [ ! -f ${AWS_FILE} ]
+then
+  echo "No AWS Credentials to check"
+  exit 0
+fi
+
+ROOT=.
+AWS_SECRET=$(grep secret ${AWS_FILE} | cut -d '=' -f2)
```
giorgiosironi: `credentials` file may have multiple profiles so this may produce multiple lines:
```
[default]
aws_secret_access_key = ...
aws_access_key_id = AKIA...
region = us-east-1

[sprint]
aws_secret_access_key = ...
aws_access_key_id = AKIA...
region = eu-west-2
```

scripts/secretcheck.sh
```diff
@@ -0,0 +1,40 @@
...
+AWS_KEY=$(grep key_id ~/.aws/credentials | cut -d '=' -f2)
+
+SECRETS=(${AWS_SECRET} ${AWS_KEY})
+
+echo "Searching for secrets... "
+
+for var in "${SECRETS[@]}"
+do
+  # Perfrom a seacrch for this secret
```
giorgiosironi: typos

scripts/secretcheck.sh
```diff
@@ -0,0 +1,40 @@
...
+  if [ $(which ag) ]
+  then
+    RESULT=$(ag -l ${AWS_SECRET} ${ROOT})
+  else 
+    echo "   *** if this is slow then install: https://github.com/ggreer/the_silver_searcher#installing"
+    RESULT=$(find ${ROOT} -type f -exec grep -l ${AWS_SECRET} {} \;)
```
giorgiosironi: why not just `grep -rl ${AWS_SECRET} .`?

giorgiosironi: (executes one `grep` rather than number-of-files `grep` processes)

giorgiosironi: if the key is in a file in `.gitignore`, it is fine, but the script still stops it

scripts/secretcheck.sh
diversemix: This should work now -as the bash array copes with space separations.

scripts/secretcheck.sh
diversemix: Fixed

scripts/secretcheck.sh
diversemix: Done, good thanks!

diversemix: I guess we will cross this bridge when we get to it... I suggest we could similarly have an ignore file list for this script too?

Cooryd (approved):

diversemix:
Going to let @giorgiosironi have a shuffty before merging :)

scripts/secretcheck.sh
```diff
@@ -0,0 +1,42 @@
...
+aws_file=~/.aws/credentials
+if [ ! -f ${aws_file} ]
+then
+  echo "No AWS Credentials to check"
+  exit 0
+fi
+
+root=.
+aws_secret=$(grep secret ${aws_file} | cut -d '=' -f2)
+aws_key=$(grep key_id ~/.aws/credentials | cut -d '=' -f2)
```
giorgiosironi, line 13: should refer to `${aws_file}`?

giorgiosironi (approved):

scripts/secretcheck.sh
diversemix: :man_facepalming:

PR #1945, opened 2019-04-25
Try to inline the base image to see the difference in performance

giorgiosironi:
Purging caches, it builds in 6 minutes on my laptop on a rather slow connection. However it reduces the size from 2.5 GB to 1.9 GB, so it may benefit the time it takes to push and pull the image across the board.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
-# built from ./Dockerfile-base
-FROM xpub/xpub:20181213-1147
+FROM node:8.14.0
+
+ENV HOME "/home/xpub"
+
+RUN mkdir -p ${HOME}
 
 WORKDIR ${HOME}
 


giorgiosironi:
Similar time, 6 minutes for `Build image`, in https://alfred.elifesciences.org/job/pull-requests-projects/job/elife-xpub/job/PR-1945/

giorgiosironi:
There was a queue for Jenkins nodes but that applies no matter how the image is built.

diversemix (approved):
Great!

erezmus (approved):

diversemix:
This is great because after this change I will slip in removing headless chrome from the image too (but not the test browser one!) ... I have this waiting ... https://gist.github.com/diversemix/c5d194901abbb9799a12adf0b5fdbc1c

giorgiosironi:
Would want to see how this compares to the version with the base image - in https://alfred.elifesciences.org/job/test-elife-xpub-source-repository/ `Build image` oscillates between 2 and 7 minutes. `Push image` has similar times.

diversemix:
If we get some builds going through at 2 mins it sounds good.

diversemix:
I've merged this in as it gives us some value - @giorgiosironi you might not be aware but Jure wants to remove the xpub base image anyhow as there is no longer value in having it. So I see this as a good change to get it - to ensure we do not rely on this any longer!

PR #1946, opened 2019-04-25
use email to check assignee type ReviewerAlias

### LINKED ISSUE #1922: Validation on excluded reviewers causes breaking bug when error thrown
**Describe the bug**

If an invalid value is entered into the excluded reviewers email field, there is a graphql error on the autosave. This then prevents the user from re-opening their submission

**Steps To Reproduce**

Create a submission and progress through to the 'Editors' step. Go down to the 'exclude a reviewer` text and click to expand the excluded reviewer section. Without adding a name value, in the field titled 'Excluded reviewer 1 email' enter an invalid value ie: 'a'. When the auto save runs, there is an error thrown:

```
ApolloError.js:39 Uncaught (in promise) Error: GraphQL error: Abstract type Assignee must resolve to an Object type at runtime for field Manuscript.assignees with value { email: "a" }, received "null". Either the Assignee type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.
    at new ApolloError (ApolloError.js:39)
    at Object.next (QueryManager.js:200)
    at notifySubscription (Observable.js:130)
    at onNotify (Observable.js:165)
    at SubscriptionObserver.next (Observable.js:219)
    at notifySubscription (Observable.js:130)
    at onNotify (Observable.js:165)
    at SubscriptionObserver.next (Observable.js:219)
    at eval (index.js:94)
```

Refreshing the page at this point then throwws the same error, preventing the user from accessing the form.

**Expected behavior**
Validation error should not cause this graphql error.

**Screenshots**

Screenshot 2019-04-16 at 11 11 49

erezmus:
#### Background

Fixes case when an invalid email causes a GraphQL bug

#### Any relevant tickets

Closes #1922

### DIFF
--- a/packages/component-submission/server/resolvers/index.test.js
+++ b/packages/component-submission/server/resolvers/index.test.js
@@ -6,7 +6,7 @@ const { createTables } = require('@elifesciences/component-model')
 const mailer = require('@pubsweet/component-send-email')
 const User = require('@elifesciences/component-model-user').model
 const Manuscript = require('@elifesciences/component-model-manuscript').model
-const { Query } = require('.')
+const { Query, Assignee } = require('.')
 const { userData } = require('./index.test.data')
 
 const replaySetup = require('../../../../test/helpers/replay-setup')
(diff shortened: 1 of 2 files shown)


diversemix:
I think its worth updating the tests to check for this too.

Also I am not sure what this function is used for - it looks highly assumptive (if that's a word @hdrury1 ?)

diversemix:
Just looking through the code ... do you know why we don't use the `role` of the assignee to determine the type - rather than guessing?

erezmus:
Ah wasn't aware of the role. I'll have another look tonight band add test

erezmus:
> Just looking through the code ... do you know why we don't use the `role` of the assignee to determine the type - rather than guessing?

So basically, we can't use role to resolve the Assignee because its not available in the __resolveType callback, only the actual data. Assignee is a union type of AuthorAlias, EditorAlias and ReviewAlias so the type is inferred from how the fields are used.

see https://www.apollographql.com/docs/graphql-tools/resolvers#unions-and-interfaces

packages/component-submission/server/resolvers/index.test.js
```diff
@@ -70,4 +70,18 @@ describe('component-submission resolvers', () => {
       })
     })
   })
+
+  describe('assignees', () => {
```
erezmus, line 74: Added those tests here, but they might be a bit too trivial?

Cooryd (approved):

diversemix (approved):
Let's raise another issue (linked to this) which presents the issue of using a union (which hides the type) and not returning the type (`role`) on each object.

PR #1971, opened 2019-05-01
Create Submission aggregate

Cooryd:
#### Background

What does this PR do?

#### Any relevant tickets

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-submission/server/aggregates/Submission.js
+++ b/packages/component-submission/server/aggregates/Submission.js
@@ -1,0 +1,33 @@
+class Submission {
+  constructor({ models: { Manuscript, File }, services: { Storage } }) {
+    this.ManuscriptModel = Manuscript
+    this.FileModel = File
+    this.Storage = Storage
+  }
+
+  async initialize(manuscriptId, userId) {
+    this.manuscript = await this.ManuscriptModel.find(manuscriptId, userId)
+    this.files = await this.FileModel.findByManuscriptId(manuscriptId)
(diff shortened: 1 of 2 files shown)


packages/component-submission/server/aggregates/Submission.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  async initialize(manuscriptId, userId) {
+    this.manuscript = await this.ManuscriptModel.find(manuscriptId, userId)
+    this.files = await this.FileModel.findByManuscriptId(manuscriptId)
+    console.log(this.manuscript, this.files)
+    return this
+  }
+
+  getFilesWithDownloadLink() {
```
erezmus: Should this be a public method ? Might be best to keep it private?

Cooryd: This is used as a private method, however I didn't think there was any way to set real private methods in node currently (???). I'd be happy to use a naming convention to make them identifiable, something like `_getFilesWithDownloadLink()`?

erezmus: one way is to use the ES6 Symbol:

top of the file:
```
const _getFilesWithDownloadLink = Symbol('getFilesWithDownloadLink')
```

Inside the class:
```
[_getFilesWithDownloadLink]() {
//...
}
```

erezmus (approved):

will-byrne (approved):

PR #1981, opened 2019-05-13
1975 - Correct content type is saved in S3 to fix download link

### LINKED ISSUE #1975: Ensure 'Preview' link saves the manuscript's file extension
**Describe the bug**

Attempting to download/preview an uploaded manuscript file in firefox showed the below dialog, which only had the option to open with "text editor" on Mac, which opened a text file that wasn't readable. Choosing "Save file" seemed to have no result. This means in Firefox on Mac I haven't been able to preview or download the manuscript file.

image

In chrome the file downloads but is not obviously a PDF or docx file. Chrome or OSX manages to open it as a PDF however. With .docx it opens as a textfile, which removes all formatting etc. 
I don't have MS word, but "pages" usually manages to open .docx correctly, so not currently behaving as expected.

image

**Expected behavior**

This may be just on Mac, I'm not sure, but I assume user expectations would be to see their file download or open in the format they uploaded it and with the same filename. Currently on Mac it's downloading as a text file.

I would say this is doing more harm than good as may lead users to believe there was an error when there is not.

Cooryd:
#### Background

There was a typo in the property name passed to the `S3.putObject` which meant an undefined `ContentType` was being sent when storing files. This corrects the typo, allowing previews to be correctly opened when downloaded.

#### Any relevant tickets

Closes #1975

### DIFF
--- a/packages/component-service-s3/server/s3Storage.js
+++ b/packages/component-service-s3/server/s3Storage.js
@@ -30,7 +30,7 @@ class S3Storage {
       .putObject({
         Body: content,
         Key: `${fileObject.url}/${fileObject.id}`,
-        ContentType: fileObject.mimetype,
+        ContentType: fileObject.mimeType,
         ContentLength: size,
         ACL: 'private',
       })


will-byrne (approved):

petereast (approved):
JavaScript is amazing

erezmus (approved):

PR #2008, opened 2019-05-21
alter validation to allow the user to submit a feature article without subjects

### LINKED ISSUE #1973: Allow Feature Article submitters to not select a Major Subject Area
## Problem / Motivation
- WHO wants this? Features editor
- WHEN and WHERE do they want it? Within the submission wizard, if a submitter selects 'Features Article' as their article type
- WHAT do they want? The application should update its validation rules based on the article type selection
- WHY do they want this? Certain article types (and Features Articles in particular) require different information from the author. In many cases, a Features Article will not need a Major Subject Area to be selected, or any editors suggested. Currently, submitters are having difficulties completing the wizard as there is information that is required to fulfil validation rules but that doesn't make domain sense.

## Proposed solution
- We should aim to update the validation rules depending on which article type has been selected on the 'Details' step of the wizard

- The final step validation should catch any errors thrown if, for example, a submitter changes article type partway through a submission

---

## Tasks
- [ ] Is is possible? Hack together a meca zip which can be imported into eJP
- [ ] Make required changes to the validation (no MSA)
- [ ] Make required changes to mecaExport for a Feature Article.

## Technical notes
- More notes or suggestions

will-byrne:
alter validation to allow feature articles to be submitted without subjects.

#### Any relevant tickets

Closes #1973 

- [X] Unit / Integration tests

### DIFF
--- a/packages/component-submission/client/utils/ValidationSchemas.js
+++ b/packages/component-submission/client/utils/ValidationSchemas.js
@@ -112,11 +112,19 @@ export const submissionSchema = {
   meta: yup.object().shape({
     title: yup.string().required('Title is required'),
     articleType: yup.string().required('Article type is required'),
-    subjects: yup
-      .array()
-      .min(1, `Choose at least 1 subject area`)
-      .max(2, `No more than 2 subject areas`)
-      .required('Subject area(s) required'),
+    subjects: yup.array().when('articleType', {
+      is: articleType => articleType && articleType === 'feature',
+      then: yup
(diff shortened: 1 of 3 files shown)


Cooryd (approved):

petereast (approved):

hdrury1:
Hi @will-byrne just tested this in Staging and Production and both still throwing validation errors when I select a Features article followed by no subject area.

PR #2009, opened 2019-05-22
1997 create update submission use case

### LINKED ISSUE #1997: Create updateSubmissionUseCase use-case
_Relies on #1996_

##Task
 
This will likely link to issue #1615.

We need to create a use-case that processes the form input sent, then passes the formatted data to the `Submission` object for saving.

This new use-case should then be used by the `updateManuscript` resolver, possibly rename this to `updateSubmission`.

erezmus:
#### Background

Creates an update submission use case to replace current update manuscript one. It does so by using the new Submission aggregate which as part of this PR has the update functionality added.

#### Any relevant tickets

Closes #1997.

- [ ] Browser tests


erezmus:
Still to do/think about:
- [ ] Move the updateManuscript.test.js from the resolvers into integration tests location ?
- [x] Remove the old updateManuscript.js file
- [x] Remove the Manuscript use case's `update` method

There's probably more.

packages/component-submission/server/aggregates/Submission.test.js
```diff
@@ -11,88 +13,92 @@ const createMockObject = (values = {}, mockSaveFn) => ({
...
+  return new Submission({ models, services })
+}
+
 describe('Submission', () => {
+  let mockManuscriptFind, mockFileFind, mockTeamFind
+
+  beforeAll(() => {
+    mockManuscriptFind = jest.spyOn(Manuscript, 'find')
+    mockFileFind = jest.spyOn(File, 'findByManuscriptId')
+    mockTeamFind = jest.spyOn(Team, 'findByManuscriptId')
```
will-byrne: these three mocks can be instantiated outside the beforeAll() where they are declared.

will-byrne (approved):

packages/component-submission/server/use-cases/updateSubmissionUseCase.js
```diff
@@ -0,0 +1,60 @@
...
+      await submission.updateManuscript(manuscriptData)
+      await submission.updateAuthorTeam(data.author)
+      await submission.updateEditorTeams(editorTeams)
+      await submission.updateReviewerTeams(reviewerTeams)
+    } catch (error) {
+      const expected = 'Data Integrity Error'
+      if (error.message.startsWith(expected)) {
+        logger.error(`Expected a ${expected}, ${error.message}`)
+      } else {
+        // not an error we were expecting.
```
Cooryd, line 49: Can probably remove this comment now https://github.com/elifesciences/elife-xpub/pull/2009/files#diff-5b518139d4766e3a75dc9c2cb1213c90R86 and https://github.com/elifesciences/elife-xpub/pull/2009/files#diff-5b518139d4766e3a75dc9c2cb1213c90R122 are expected error cases given 'bad' `data` objects.

packages/component-submission/server/aggregates/Submission.test.js
```diff
@@ -192,33 +176,287 @@ describe('Submission', () => {
...
+      submission.updateAuthorTeam(2)
+      expect(submission.teams[0]).toEqual({
+        objectType: 'manuscript',
+        role: 'author',
+        teamMembers: [{ alias: 2, meta: { corresponding: true } }],
+      })
+    })
+  })
+
+  describe('updateEditorTeams', () => {
```
Cooryd, line 246: Can we add a test for the validation check here https://github.com/elifesciences/elife-xpub/pull/2009/files#diff-5b518139d4766e3a75dc9c2cb1213c90R122

packages/component-submission/server/aggregates/Submission.test.js
```diff
@@ -192,33 +176,287 @@ describe('Submission', () => {
       )
```
Cooryd, line 176: I think we should have a unit test under the `updateManuscript` describe to check the status validation here: https://github.com/elifesciences/elife-xpub/pull/2009/files#diff-5b518139d4766e3a75dc9c2cb1213c90R84

packages/component-model-team/entities/team/index.js
```diff
@@ -24,6 +24,18 @@ class Team extends BaseModel {
     // tell objection to serialise teamMembers as a postgres array rather than JSON
     return []
   }
+
+  static async findByManuscriptId(manuscriptId) {
+    const teams = await this.query().where({
```
petereast, line 29: Wait - static functions shouldn't have access to `this` - or am I getting confused?

packages/component-submission/server/use-cases/updateSubmissionUseCase.js
```diff
@@ -0,0 +1,60 @@
...
+      await submission.initialize(manuscriptId, userId)
+
+      await submission.updateManuscript(manuscriptData)
+      await submission.updateAuthorTeam(data.author)
+      await submission.updateEditorTeams(editorTeams)
+      await submission.updateReviewerTeams(reviewerTeams)
+    } catch (error) {
+      const expected = 'Data Integrity Error'
+      if (error.message.startsWith(expected)) {
+        logger.error(`Expected a ${expected}, ${error.message}`)
```
petereast, line 47: Is it safe to just swallow these errors - is there anything we need to do to clean up after the error occurs? Also which of the above statements throws - if it throws will it leave the stored data in an error state or will the transaction be aborted?

petereast: Upon further inspection, if one of the `submission.update*` fails, the there's potential for a partial object to be stored and the error be swallowed resulting in :warning:  **loss of data**. :warning: 

I suggest a change is made to each of the `submission.update*` functions allowing the user to pass in a `transaction` object that's used by the internal database calls so these updates can be rolled back if one of them fails.

petereast (dismissed):

packages/component-submission/server/use-cases/updateSubmissionUseCase.js
erezmus: The problem is that internally it calls the Manuscript model's save method which has a transaction and throws if the record has a later timestamp. Not sure how nesting transactions work and rewriting the manuscript save method is too big a change. I do agree however that it should be all one transaction

petereast: Maybe we could just delete everything related to this `Manuscript` if an error occurs? The way it is currently, if an error happens at any stage we're going to end up with partial/corrupted data in the database, I can't think of anything else we could do without introducing transactions at this level.

erezmus: I think some of the teams not being saved is acceptable for now and that would be reflected on the client once the new data has been fetched.

Cooryd (approved):

PR #2013, opened 2019-05-22
1990 - Add authsome wrapper

### LINKED ISSUE #1990: Add an withAuthsomeMiddleware wrapper to be used at the resolver level
## Task

Following #1986 configure the application to use authsome and create a middleware wrapper similar (or the same 👍 ) to hindawi's https://gitlab.com/hindawi/xpub/xpub-review/blob/develop/packages/helper-service/src/withAuthsomeMiddleware.js that can be passed authorisation policies.

It might not be worth wrapping resolvers in this yet if it causes issues. Just getting the files/ config in should be fine.

petereast:
#### Background

What does this PR do?

#### Any relevant tickets
Closes: #1990 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-elife-app/server/authsomeMode.js
+++ b/packages/component-elife-app/server/authsomeMode.js
@@ -3,8 +3,8 @@ const BluebirdPromise = require('bluebird')
 // Policy type: (user: string, operation: string, object: any, context?: any) => boolean
 const authsomePolicies = {
   // TODO: Somehow extend this list?
-  isAuthenticated: (userId, _a, _b, _c) => !!userId,
-  isNotAuthenticated: (userId, _a, _b, _c) => !userId,
+  isAuthenticated: userId => !!userId,
+  isNotAuthenticated: userId => !userId,
 }
 
 const authsomeMode = async (userId, { name, policies = [] }, object, context) =>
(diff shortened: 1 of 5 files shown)


packages/component-elife-app/server/authsomeMode.js
```diff
@@ -0,0 +1,29 @@
+const BluebirdPromise = require('bluebird')
+
+// Policy type: (user: string, operation: string, object: any, context?: any) => boolean
+const authsomePolicies = {
```
petereast, line 4: We need to have some sort of discussion about how we organise this list of policies: where a policy for a specific use case will reside; conventions for how we write policies and if it's worth investigating a better, less repetitive, safer way of doing it.

packages/component-elife-app/server/authsomeMode.js
```diff
@@ -0,0 +1,29 @@
...
+  // TODO: Somehow extend this list?
+  isAuthenticated: (userId, _a, _b, _c) => !!userId,
```
Cooryd: We can drop the unused params being passed here for now

packages/component-elife-app/server/authsomeMode.js
erezmus: we could use constants ?

erezmus (approved):

packages/component-elife-app/server/authsomeMode.js
petereast: I just really don't like referring to the policies with strings - I feel like it's a really fragile way of referring to an important part of our system logic. I think it might be worth investing some time to build a really flexible policy system but it would probably be quite ~difficult~ interesting to do well.

Cooryd (approved):

PR #2017, opened 2019-05-23
decoupled the add supporting files messages from the manuscript state

### LINKED ISSUE #2000: Decouple supporting file upload from manuscript state
Follow up from #1983, the check for whether a manuscript file has been uploaded is done by the validation schema. Currently, the user can get into a situation where the "Remove All" and "Add more supporting files" button are not shown with the following steps:

1. Upload a valid manuscript
2. Upload a supporting file
3. Try to upload an invalid manuscript

The "Remove All" and "Add more supporting files" are not shown as the logic mandates that a manuscript is also present (which is no longer the case).

Following on from https://github.com/elifesciences/elife-xpub/issues/1983#issuecomment-493459899, we should make the presence or absence of those buttons independent of the presence of a manuscript and slightly alter the wording:

- it should be hidden when no file is uploaded
- it should say "add supporting files" when the manuscript is uploaded
- it should say "Add more supporting files" if one more more files are already present
- If the user does anything else to the manuscript file, it doesn't matter. If there are supporting files it should say "Add more supporting files"
- if the user presses "remove all" it should say "add supporting files" again. It should never be removed entirely as the file upload target should never return to an "upload" state - it should either have a manuscript of be in an error state.

will-byrne:
#### Background

decouples the add supporting files messages from the manuscript state:

- no manuscript && no supporting files => no messages
- manuscript && no supporting files => "add supporting..." message
- no manuscript && some supporting files => "add more..." message
- 10 supporting files => "Max 10...." message

#### Any relevant tickets

Closes #2000 

### DIFF
--- a/packages/component-submission/client/components/SupportingUpload.js
+++ b/packages/component-submission/client/components/SupportingUpload.js
@@ -284,14 +284,14 @@ class SupportingUpload extends React.Component {
             </FileBlock>
           ))}
         </StyledDropzone>
-        {hasManuscript &&
+        {(hasManuscript || successfullyUploadedFiles.length > 0) &&
           !this.state.uploading && (
             <React.Fragment>
               <Flex>
                 {successfullyUploadedFiles.length < MAX_SUPPORTING_FILES && (
                   <React.Fragment>
-                    <UploadControl>
-                      Add more{' '}
+                    <UploadControl data-test-id="supporting-files-upload-control">
+                      Add {successfullyUploadedFiles.length > 0 && 'more'}{' '}
                       <UploadLink
                         data-test-id="supportingFilesLink"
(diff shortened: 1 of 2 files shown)


Cooryd (approved):

packages/component-submission/client/components/SupportingUpload.js
```diff
@@ -284,14 +284,14 @@ class SupportingUpload extends React.Component {
             </FileBlock>
           ))}
         </StyledDropzone>
-        {hasManuscript &&
+        {(hasManuscript || successfullyUploadedFiles.length > 0) &&
```
petereast, line 287: We're getting close to the point where we'd need to split these conditionals out into a separate function

petereast (approved):

PR #2019, opened 2019-05-24
1991 - Add author policy

### LINKED ISSUE #1991: Create an author authsome policy to be used with middleware
## Task

Following #1990 and #1989 we need to create an authsome policy that can wrap all existing mutations / queries (excluding create new manuscript ? ) to authenticate that the user performing the action is the author of the manuscript. This can then replace the use of userId querying.

_If a different policy is needed for creating a new manuscript. Raise another issue._

petereast:
#### Background

What does this PR do?

#### Any relevant tickets

Closes #1991 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-elife-app/server/authsomeMode.js
+++ b/packages/component-elife-app/server/authsomeMode.js
@@ -1,12 +1,10 @@
 const BluebirdPromise = require('bluebird')
-const { isAuthor } = require('./policies')
 
 // Policy type: (user: string, operation: string, request: any, context?: any) => boolean
 const authsomePolicies = {
   // TODO: Somehow extend this list?
   isAuthenticated: userId => !!userId,
   isNotAuthenticated: userId => !userId,
-  isAuthor,
 }
 
 const authsomeMode = async (userId, { name, policies = [] }, object, context) =>
--- a/packages/component-elife-app/server/authsomeMode.js
+++ b/packages/component-elife-app/server/authsomeMode.js
@@ -1,10 +1,12 @@
 const BluebirdPromise = require('bluebird')
+const { isAuthor } = require('./policies')
 
-// Policy type: (user: string, operation: string, object: any, context?: any) => boolean
+// Policy type: (user: string, operation: string, request: any, context?: any) => boolean
 const authsomePolicies = {
   // TODO: Somehow extend this list?
   isAuthenticated: userId => !!userId,
   isNotAuthenticated: userId => !userId,
+  isAuthor,
 }
 
 const authsomeMode = async (userId, { name, policies = [] }, object, context) =>
--- a/packages/component-elife-app/server/policies/isAuthor.js
+++ b/packages/component-elife-app/server/policies/isAuthor.js
@@ -1,4 +1,4 @@
-module.exports = async (userId, operation, request, context) => {
+module.exports = (userId, operation, request, context) => {
   const { Manuscript } = context.models
 
   // There must be a more elegant way of doing this
--- a/packages/component-elife-app/server/policies/isAuthor.js
+++ b/packages/component-elife-app/server/policies/isAuthor.js
@@ -1,0 +1,8 @@
+module.exports = async (userId, operation, request, context) => {
+  const { Manuscript } = context.models
+
+  // There must be a more elegant way of doing this
+  return Manuscript.findByIdAndAuthor(request.manuscriptId, userId)
+    .then(() => true)
+    .catch(() => false)
+}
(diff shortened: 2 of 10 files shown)


packages/component-elife-app/server/policies/isAuthor.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('allows the author of a manuscript to access the resource', async done => {
+    await expect(
+      modifiedResolvers.Mutation[resolver1Name](
+        { original: 'params' },
+        { manuscriptId: 'exampleManuscriptId' },
+        { user: 'a_valid_user' },
+      ),
+    ).resolves.toEqual({ authorized: true })
+    done()
```
Cooryd: I'm not sure `done` is necessary when the test itself is being run as an async with awaits

petereast: I'm pretty sure it's required when the function passed to `it` is `async`

petereast: Wait no it's not needed - I have removed it

packages/component-elife-app/server/policies/isAuthor.js
```diff
@@ -0,0 +1,8 @@
+module.exports = async (userId, operation, request, context) => {
```
Cooryd: This shouldn't need to be an async function if you're resolving the promise on return. awaiting the calling of this should be enough.

petereast: Yeah, you're right - because the function returns a promise anyway, there's no need to double wrap it

packages/component-model-manuscript/entities/manuscript/index.js
```diff
@@ -143,6 +143,22 @@ class Manuscript extends BaseModel {
     return manuscript
   }
 
+  static async findByIdAndAuthor(id, user, related = '[teams, files]') {
```
Cooryd: Suggestion: Maybe we could use find in the policy and assert `manuscript.created_by === user`. This piece of work (once added to the resolvers) should make the querying by `id` and `created_by` unnecessary as we'll be able to do a `findById` instead.

petereast: I was working on the assumption that what is currently `find` (but should be called `findByIdAndAuthor`) will change to what should be called `findById`, so I created the `findByIdAndAuthor` function based on the current `find` function so we don't have to fetch and process a load of manuscripts for every request, we can just check if this `userId` matches up to the `Manuscript` they're requesting, without actually having to process any `Manuscript` objects

Cooryd: I'm thinking that the `Manuscript` would only need one find function that takes an id. There's no use-case for querying by id and userId right now other than for authorisation which this bit of middleware should be in control of in future. Better to get this policy working with the intended `find` then add the extra function to the model?

petereast: Yeah, that makes a lot more sense, I'll do that

packages/component-elife-app/server/authsomeMode.js
```diff
@@ -1,10 +1,12 @@
 const BluebirdPromise = require('bluebird')
+const { isAuthor } = require('./policies')
 
-// Policy type: (user: string, operation: string, object: any, context?: any) => boolean
+// Policy type: (user: string, operation: string, request: any, context?: any) => boolean
 const authsomePolicies = {
   // TODO: Somehow extend this list?
   isAuthenticated: userId => !!userId,
   isNotAuthenticated: userId => !userId,
+  isAuthor,
```
Cooryd: I'm thinking maybe we change the policy name from what the user has to __be__ to what they are expected to be able to __do__. This policy will need extending soon to also return true is the user is a staff member. This will make the name of the policy inaccurate. My suggestion would be something like `hasAccessToManuscript` 

- see https://gitlab.com/hindawi/xpub/xpub-review/blob/develop/packages/app/config/authsome-mode.js#L97

petereast: Done :+1:

will-byrne (approved):

erezmus (approved):

packages/component-model-manuscript/entities/manuscript/index.js
```diff
@@ -143,6 +143,22 @@ class Manuscript extends BaseModel {
     return manuscript
   }
 
+  static async findById(id, related = '[teams, files]') {
+    // This is a copypasta of `find` because `find` might be changed so that it doesn't query by userId
+    // TODO: Once we've got the policies in place, replace `find` with this function
+    const [manuscript] = await this.query().where({
```
Cooryd, line 149: This can use objections `findById` query builder function https://vincit.github.io/objection.js/api/query-builder/find-methods.html#findbyid

PR #2027, opened 2019-05-24
2020 - rediect to a given config url on logout

### LINKED ISSUE #2020: Implement single sign out pattern
When a user logs out of Libero Reviewer they should be also logged out of Journal.

Logging out of Libero Reviewer should redirect the user to Journal home.

Hopefully this will solve some of the bugs we have been seeing with regards the different login screens and logging out in general (#1652 and #1904) but we should aim to implement this pattern first before tackling any remaining bugs.

Cooryd:
#### Background

The user should be signed out of the journal and be redirected back to the journals home page if they log out of reviewer. This PR adds this redirect based on a config value after clearing the local apollo store and removing the reviewer session storage token.

#### Any relevant tickets

closes #2020

### DIFF
--- a/config/__tests__/__snapshots__/config.ci.test.js.snap
+++ b/config/__tests__/__snapshots__/config.ci.test.js.snap
@@ -45,6 +45,7 @@ Array [
   "login.enableMock",
   "login.signupUrl",
   "login.legacySubmissionUrl",
+  "logout.isPublic",
   "mailer.from",
   "mailer.path",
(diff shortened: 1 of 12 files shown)


erezmus (approved):

petereast (approved):

diversemix (changes requested):
I'd like to fully understand what is going on before this gets merged.

diversemix (approved):
Discussed in retro

PR #2040, opened 2019-05-28
2039: don't use Manuscript model

### LINKED ISSUE #2039: Change removeSupportingFiles resolver to not use Manuscript model
The `removeSupportingFiles` resolver calls the `supportingFiles.removeAll` method which in turn, does a Manuscript model save.

This can potentially cause a race condition and should remove the files independently from saving the manuscript

erezmus:
#### Background

Changes the removeSupportingFiles resolver underlying logic so that the Manuscript model's save method isn't called. This should reduce the amount of browser test failures

#### Any relevant tickets

Closes #2039

### DIFF
--- a/packages/component-submission/server/use-cases/supportingFiles.js
+++ b/packages/component-submission/server/use-cases/supportingFiles.js
@@ -1,5 +1,7 @@
 const config = require('config')
 const FileModel = require('@elifesciences/component-model-file').model
+const ManuscriptModel = require('@elifesciences/component-model-manuscript')
+  .model
 const logger = require('@pubsweet/logger')
 const Manuscript = require('./manuscript')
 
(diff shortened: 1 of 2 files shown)


packages/component-submission/server/use-cases/supportingFiles.js
```diff
@@ -48,21 +48,23 @@ class SupportingFiles {
   }
 
   async removeAll() {
-    let manuscript = await ManuscriptModel.find(this.manuscriptId, this.userId)
-    const filesWithoutSupporting = manuscript.files.filter(
-      file => file.type !== 'SUPPORTING_FILE',
-    )
+    const manuscript = new Manuscript(config, this.userId, this.storage)
```
diversemix, line 53: I cannot see where the `manuscript` variable is used ... we do need to check the user has access to this manuscript tho before trying to remove it ... i.e. I think we should re-instate line 53

packages/component-submission/server/use-cases/supportingFiles.js
```diff
@@ -48,21 +50,21 @@ class SupportingFiles {
...
+
+    // we need to do a find to make sure the user has access
+    await ManuscriptModel.find(this.manuscriptId, this.userId)
 
     const files = await FileModel.findByManuscriptId(this.manuscriptId)
 
     if (files && files.length > 0) {
-      let modified = false
+      // let modified = false
```
will-byrne: should be removed rather than commented.

packages/component-submission/server/use-cases/supportingFiles.js
```diff
@@ -48,21 +50,21 @@ class SupportingFiles {
...
       await Promise.all(
         files
           .filter(file => file.type === 'SUPPORTING_FILE')
           .map(async file => {
             try {
-              modified = true
+              // modified = true
```
will-byrne: should be removed rather than commented.

will-byrne (approved):

Cooryd (approved):

PR #2063, opened 2019-06-04
Build a smaller docker image for the app

### LINKED ISSUE #517: Large Docker Containers
#507  has highlighted another issue - one of the container size being so large that we cannot do a test deployment locally in a VM - as the network keeps dropping. This also might be due to the office environment too. There is a fix for this in `245-prod` to dramatically reduce the size of the container. Going to raise a separate issue for this.

diversemix:
#### Background

Builds a smaller image. This should help the pipeline speed dealing with smaller images - the current one is 2.6G this is less than 500 Mb

Remove substance/texture as an unused package.

#### Any relevant tickets

Closes #517

Tested in pipeline.


diversemix:
Closing this out as it makes Jenkins fail. Keeping the branch and going to do a more strict carpaccio on this  ;)

PR #2072, opened 2019-06-04
First step to improve dockerization

diversemix:
#### Background

This truly separates the images for the `test_browser` and the `app` - the result is an image which is 1Gb smaller than before.

#### Any relevant tickets

Contributes to closing #517

Tested in the pipeline. Also re-enabled the browser tests to ensure they would work. On doing so realised that the browser tests are dependent on a lot of code from the `app` purely because it uses the config and this is only for the URL under test.


will-byrne (approved):

Dockerfile-development
```diff
@@ -2,16 +2,10 @@ FROM node:8.14.0
 
 ENV HOME "/home/xpub"
 
-RUN mkdir -p ${HOME}
+VOLUME ["/home/xpub"]
```
erezmus, line 5: Is this necessary ? we are specifying the volume in the docker-compose.override.yml file

erezmus: actually after thinking about it is it a way of declaring that the container should have volume mounted?

diversemix: It probably isn't - but I like being explicit ... because now this container will not work (as I removed the copy commands) without a mounted volume.

Dockerfile-test_browser
```diff
@@ -44,15 +44,6 @@ COPY test test
 COPY webpack webpack
 COPY templates templates
 
-ENV NODE_ENV "production"
-ARG CI_COMMIT_SHA
```
giorgiosironi, line 48: this is being removed, but it's not in `Dockerfile-app`

giorgiosironi:
I didn't remove the nodes, I added a cleaning step in the `Jenkinsfile` instead.

Dockerfile-test_browser
diversemix: `Dockerfile-app` does specify `ENV NODE_ENV "production"` ... but we no longer use the `CI_COMMIT_SHA` in the served pages.

diversemix:
> I didn't remove the nodes, I added a cleaning step in the `Jenkinsfile` instead.

Thanks - looks like its failing as normal. Although I do think the error is with the pipeline. Running `scripts/local-pipeline.sh` (locally on my machine) works running the browser tests ... where as the [pipeline](https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Felife-xpub/detail/PR-2072/25/pipeline) reports that `testcafe` is missing from the container.

giorgiosironi:
`./:/home/xpub` is being mounted on the `test_browser` container so there won't be a `node_modules/` in there containing all binaries like `testcafe`?

giorgiosironi:
It does say:
```
[2019-06-05T15:24:08.753Z] npm WARN Local package.json exists, but node_modules missing, did you mean to install?
``` 
at the end of https://alfred.elifesciences.org/blue/rest/organizations/jenkins/pipelines/pull-requests-projects/pipelines/elife-xpub/branches/PR-2072/runs/28/nodes/75/steps/85/log/?start=0

It works locally because you have `node_modules` in the source folder?

erezmus (approved):

PR #2080, opened 2019-06-06
[WIP] Add two-stage docker file for app

### LINKED ISSUE #517: Large Docker Containers

diversemix:
#### Background

This is to generate a much smaller image for the app.

This may take longer to do - do we prefer speed over the size

#### Any relevant tickets

Closes #517

In the pipeline

### DIFF
--- a/Dockerfile-app
+++ b/Dockerfile-app
@@ -1,14 +1,22 @@
-FROM node:8.14.0
+FROM node:8.14 as builder
 
-ENV HOME "/home/xpub"
+RUN mkdir /build
+WORKDIR /build
 
-RUN mkdir -p ${HOME}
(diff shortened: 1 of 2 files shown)


diversemix:
Giving up on building a good Docker container ... turns out we need `node-sftp-server` (which is used by a mock for the tests) ... this is needed by the tests internal to the container. So in turn *WE HAVE THIS IN PRODUCTION*

Giving up until we have sorted out the tests!

PR #2093, opened 2019-06-06
Added the modal to the new submission wizard

### LINKED ISSUE #2083: Next button on disclosure page needs to trigger modal and call submit
## Task

Currently the next button will only navigate to the next step after validating. On disclosure this should open the confirm model and then call a submit callback.

The onNext function should be moved out of the buttons onClick definition in to its own function.

will-byrne:
#### Background

Adds the modal to the new submission wizard and exposes the same property for the save to be passed in as before.

#### Any relevant tickets

closes #2083

### DIFF
--- a/packages/component-submission/client/pages/NewSubmissionWizard.js
+++ b/packages/component-submission/client/pages/NewSubmissionWizard.js
@@ -18,6 +18,8 @@ import DetailsStep from './DetailsStepPage'
 import EditorStep from './EditorsStepPage'
 import DisclosureStep from './DisclosureStepPage'
 
+import WizardSubmit from '../components/WizardSubmit'
+
 import ProgressBar from '../components/ProgressBar'
 import wizardWithGQL from '../graphql/wizardWithGQL'
 import { parseInputToFormData, flattenObject } from '../utils'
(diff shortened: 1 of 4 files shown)


erezmus (approved):

packages/component-submission/client/pages/NewSubmissionWizard.js
```diff
@@ -52,6 +55,9 @@ const NewSubmissionWizard = ({ initialValues, match, history }) => {
   return (
     <Formik
       initialValues={initialValues}
+      onSubmit={() => {
+        console.log('2051')
```
Cooryd, line 73: For reference this is a placeholder to call the onSubmit handler.

packages/component-submission/client/pages/NewSubmissionWizard.js
```diff
@@ -78,7 +84,12 @@ const NewSubmissionWizard = ({ initialValues, match, history }) => {
                 />
                 <TrackedRoute
                   path={`${match.path}/disclosure`}
-                  render={() => <DisclosureStep {...formikProps} />}
+                  render={() => (
+                    <DisclosureStep
+                      isSubmissionAttempted={submissionAttempted}
```
Cooryd, line 109: Maybe we can move the error message up to this level so that it doesn't sit in the Disclosure step? This would make the form more easily customisable (removing the disclosure page wouldn't require moving the whole form validation message)

will-byrne: The error message is part of the disclosure page and not inside the modal so not sure where we would put it as it wouldn't render anywhere else.

Cooryd (approved):
Remove failing test as its no longer in step and we can re-evaluate test coverage for this component in a future issue.

PR #2096, opened 2019-06-06
Remove updating the manuscript title on file upload

### LINKED ISSUE #2034: Only write the title of the manuscript object on "Details" page.
Remove the writing of the title to the manuscript table after the file upload... leave the writing to the semantic table.

Then when the "Details" page is visited if the manuscript title is blank ... populate it from the semantic table.

diversemix:
#### Background

What does this PR do?

#### Any relevant tickets

closes #2034 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-model-semantic-extraction/entities/semanticExtraction/index.test.js
+++ b/packages/component-model-semantic-extraction/entities/semanticExtraction/index.test.js
@@ -40,4 +40,18 @@ describe('SemanticExtraction', () => {
       expect(titleExtraction.fieldName).toBe('title')
     })
   })
+
+  describe('findByManuscriptId', () => {
+    it('return an entity with title filedName when called', () => {
+      const titleExtraction = SemanticExtraction.createTitleEntity(
+        'manuscriptId',
+      )
+      expect(titleExtraction.fieldName).toBe('title')
+      const result = SemanticExtraction.findByManuscriptId(
+        titleExtraction.manuscriptId,
+      )
+      expect(result).toHaveProperty('title')
+      expect(result.title).toBe('test_title')
+    })
+  })
 })
--- a/packages/component-model-semantic-extraction/entities/semanticExtraction/index.test.js
+++ b/packages/component-model-semantic-extraction/entities/semanticExtraction/index.test.js
@@ -36,22 +36,28 @@ describe('SemanticExtraction', () => {
     it('return an entity with title filedName when called', () => {
       const titleExtraction = SemanticExtraction.createTitleEntity(
         'manuscriptId',
+        'test title',
       )
       expect(titleExtraction.fieldName).toBe('title')
+      expect(titleExtraction.value).toBe('test title')
     })
   })
 
   describe('findByManuscriptId', () => {
-    it('return an entity with title filedName when called', () => {
-      const titleExtraction = SemanticExtraction.createTitleEntity(
-        'manuscriptId',
-      )
+    it('return an entity with title filedName when called', async () => {
+      const id = uuid()
+      const titleExtraction = SemanticExtraction.createTitleEntity(id, 'abc123')
       expect(titleExtraction.fieldName).toBe('title')
-      const result = SemanticExtraction.findByManuscriptId(
-        titleExtraction.manuscriptId,
-      )
-      expect(result).toHaveProperty('title')
-      expect(result.title).toBe('test_title')
+      await titleExtraction.save()
+
+      const result = await SemanticExtraction.findByManuscriptId(id)
+      expect(result).toHaveLength(1)
+
+      expect(result[0]).toHaveProperty('fieldName')
+      expect(result[0].fieldName).toBe('title')
+
+      expect(result[0]).toHaveProperty('value')
+      expect(result[0].value).toBe('abc123')
     })
   })
 })
--- a/packages/component-submission/server/aggregates/Submission.js
+++ b/packages/component-submission/server/aggregates/Submission.js
@@ -3,14 +3,29 @@ const { mergeObjects } = require('../utils')
 
 class Submission {
   constructor({
-    models: { Manuscript, File, Team, SemanticExtractionModel },
+    models: { Manuscript, File, Team, SemanticExtraction },
     services: { Storage },
   }) {
     this.ManuscriptModel = Manuscript
     this.FileModel = File
     this.TeamModel = Team
+    this.SemanticExtractionModel = SemanticExtraction
     this.Storage = Storage
-    this.SemanticExtraction = SemanticExtractionModel
+
+    if (this.ManuscriptModel == null) {
+      throw new Error('Cannot construct Submission without a Manuscript model')
+    }
+    if (this.FileModel == null) {
+      throw new Error('Cannot construct Submission without a File model')
+    }
+    if (this.TeamModel == null) {
+      throw new Error('Cannot construct Submission without a Team model')
+    }
+    if (this.SemanticExtractionModel == null) {
+      throw new Error(
+        'Cannot construct Submission without a SemanticExtraction model',
+      )
+    }
   }
 
   async initialize(manuscriptId, userId) {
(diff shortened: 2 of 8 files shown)


packages/component-submission/server/use-cases/helpers/manuscript.js
```diff
@@ -77,18 +77,14 @@ class ManuscriptHelper {
     )
 
     try {
-      let manuscript = await ManuscriptModel.find(manuscriptId, this.userId)
+      const manuscript = await ManuscriptModel.find(manuscriptId, this.userId)
```
diversemix: we still need this for the permissions check

packages/component-submission/server/aggregates/Submission.test.js
```diff
@@ -1,6 +1,8 @@
 const Manuscript = require('@elifesciences/component-model-manuscript').model
 const File = require('@elifesciences/component-model-file').model
 const Team = require('@elifesciences/component-model-team').model
+const SemanticExtraction = require('@elifesciences/component-model-semantic-extraction')
```
diversemix, line 4: need to change package to export on `model` like the others

packages/component-submission/server/aggregates/Submission.js
```diff
@@ -17,6 +21,9 @@ class Submission {
     )
     this.files = await this.FileModel.findByManuscriptId(manuscriptId)
     this.teams = await this.TeamModel.findByManuscriptId(manuscriptId)
+    this.suggestions = await this.SemanticExtraction.findByManuscriptId(
```
diversemix: do we need to change the graphQL?

packages/component-submission/server/use-cases/helpers/manuscript.js
petereast: Can we clarify that somehow - it's not clear that this is the step that performs the permissions check

packages/component-model-semantic-extraction/entities/semanticExtraction/index.test.js
```diff
@@ -36,8 +36,28 @@ describe('SemanticExtraction', () => {
...
       const titleExtraction = SemanticExtraction.createTitleEntity(
         'manuscriptId',
+        'test title',
       )
       expect(titleExtraction.fieldName).toBe('title')
+      expect(titleExtraction.value).toBe('test title')
+    })
+  })
+
+  describe('findByManuscriptId', () => {
```
petereast, line 46: Does this test call the database? I was thinking it might be a good idea to mark tests that do this with some kind of annotation moving forward, so when the time comes to decouple we can find them easily.

packages/component-submission/server/use-cases/helpers/manuscript.js
will-byrne: Ideally this should be extracted to a util function given all the use-cases will need it eventually. Also when we introduce the next stage of users and staff it would be easier to replace one function than all of these individual checks. (note: not necessarily suggesting for this ticket as its a bit out of scope)

petereast: Yeah, I guess I'm just not a massive fan of using throws to express logic like this, it's not clear what's going on.

packages/component-model-semantic-extraction/entities/semanticExtraction/index.test.js
will-byrne: Good idea, alternatively the tests can be re-written with a mocked database. If they require the actual database then either we are testing Objection/Knex at this point and the test can probably be removed or it means we have too much logic in our ORM model

diversemix: Yeah we are stuck in this place at the moment ... would love to chat about this sometime - maybe next retro?

packages/component-submission/server/aggregates/Submission.js
will-byrne: Isnt the point of graphql that we dont have to change it, we only need to handle the response at the client?

packages/component-submission/server/aggregates/Submission.test.js
Cooryd: This isn't the case in tests like this as the `@pubsweet/model` package registers the models at server start up which doesn't happen when running unit tests.

packages/component-submission/server/aggregates/Submission.js
```diff
@@ -2,21 +2,57 @@ const { intersection } = require('lodash')
...
       manuscriptId,
       userId,
       '[teams]',
     )
     this.files = await this.FileModel.findByManuscriptId(manuscriptId)
     this.teams = await this.TeamModel.findByManuscriptId(manuscriptId)
+    this.suggestions = (await this.SemanticExtractionModel.findByManuscriptId(
+      manuscriptId,
+    ))
+      .sort((exA, exB) => exA.updated > exB.updated)
```
diversemix: Ooo - I never thought about this... if we have multiple suggestions for the same field at different times - it makes sense that the latest is used. But should this logic exist on the client rather than the server?

erezmus (approved):

packages/component-submission/server/aggregates/Submission.js
Cooryd: Should we be returning all suggestions to the client or just the most recent one? I'm not sure at the moment it would matter to be sending it all back but in future if we're extracting abstracts and other chunks of text it might not be best to send all semantic extractions back to the client, just those to be used.

packages/component-submission/server/use-cases/helpers/manuscript.js
diversemix: I'm happy to have a `Manuscript` function called `checkPermissions` which does this

packages/component-submission/server/aggregates/Submission.js
```diff
@@ -77,6 +113,10 @@ class Submission {
       fileStatus: this.filesAreStored() ? 'READY' : 'CHANGING',
       files: this._getFilesWithDownloadLink(),
       teams: this.teams.map(team => team.toJSON()),
+      suggestions: Object.entries(this.suggestions).reduce(
```
Cooryd: Is an array going to be the best way of presenting this info? would an object with the field name as the index not be easier to work with on the client?

packages/component-submission/server/aggregates/Submission.js
erezmus: Is there a case where we would have multiple suggestions ?

I'm guessing there could be with a score, similar as when doing reverse geocoding in say google maps api and maybe let the client choose ?

diversemix: We've just discussed this here - but this is an interesting point. I'm inclined to make the server dumb for now and send all the suggestions back. But we need to perhaps think a little deeper about this and how it is to work in the future. I think there is a valid case of two different suggestions being valid for the same field - as well as two different suggestions representing an newer and older (i.e valid and invalid) ... Maybe the way to distinguish is to use a `method` so that only the recent suggestion from the same method is the most valid?

/cc @hdrury1 @petereast @de-code

de-code: @diversemix I wasn't quite sure whether it affects how you use ScienceBeam? Maybe you could quickly explain it tomorrow.

diversemix: Can do ... @petereast I think the MVP of this is what we already discussed and is moving the logic into the client.

packages/component-submission/server/aggregates/Submission.js
petereast: You can't do that with Graphql - the object keys cannot be defined at runtime

petereast (approved):
on @diversemix 's behalf

packages/component-submission/server/aggregates/Submission.js
```diff
@@ -164,6 +186,41 @@ class Submission {
...
+      acc,
+      { fieldName, value, updated },
+      index,
+    ) => {
+      const existing = acc.find(e => e.fieldName === fieldName)
+
+      const next = {
+        fieldName,
+        suggestions: [
+          // Ew gross - I hate javscript
```
will-byrne, line 201: while funny, this should be removed

will-byrne (changes requested):

PR #2101, opened 2019-06-07
Documentation for Server Side code guidelines

diversemix:
### Background

Proposes a set of principles for the server code to follow.

----
### Notes

This is by no means in a "finished" state - but should be enough for discussion and to get an idea of where my thinking is at.


Cooryd (approved):

docs/adr/0006-server-arch.md
```diff
@@ -0,0 +1,185 @@
...
+
+Date: 7th June 2019
+
+## Status
+
+Proposed
+
+## Context
+
+The server-side code has a number of issues with it which make it difficult to
```
giorgiosironi: Empirically, if people find it difficult to work with, it means it's difficult to work with. I wouldn't overdiagnose it as SOLID is specific to object-oriented programming in languages that tend towards static typing.

docs/adr/0006-server-arch.md
```diff
@@ -0,0 +1,185 @@
...
+  }
+  ...
+```
+
+### Use-Cases
+
+- In general this layer will have an interface that maps one-to-one with the
+  graphQL layer.
+- For a given use-case it will require certain services and config - these
+  dependencies will be created by this layer and injected (when needed) into
```
giorgiosironi: injection + "when needed": smell to me (as in partially constructed model object, with dependencies left to null or that change often). Perhaps you meant: passed in as parameters to models methods to avoid a dependency going from the lower to the upper layer?

docs/adr/0006-server-arch.md
```diff
@@ -0,0 +1,185 @@
...
+
+module.exports = {
+  getSubmissionUseCase,
+}
+```
+
+### The Aggregate
+
+- [See DDD_Aggregates](https://www.martinfowler.com/bliki/DDD_Aggregate.html)
+- An aggregate
```
giorgiosironi: should summarize the important principles from there:
- no persistent references to objects other than the Aggregate Root from outside the Aggregate (e.g. no fields referring to them and/or foreign keys, but passing them around for computation is ok)
- no transaction crosses Aggregates boundaries, as they are the unit of consistency; corollary: only eventual consistency cannot be guaranteed across Aggregate boundaries

docs/adr/0006-server-arch.md
```diff
@@ -0,0 +1,185 @@
...
+
+  async get(id, user) {
+      /*
+      * Here we could get the files' contents, the teams and the manuscript
+      * then stitch them together to send back
+      */
+      return {
+        manuscript: await this.models.Manuscript.find(id),
+        teams: await this.models.Team.findByManuscriptId(id),
+        files: await this.models.File.get(config, fileId, this.services.Storage)
```
giorgiosironi: some of these are conceptually different, may be worth noting. `manuscript` and `files` are internal to this Aggregate whereas `teams` seems to be a reference to other Aggregates. Therefore `teams` should be conceptually read-only in this Aggregate's view.

docs/adr/0006-server-arch.md
```diff
@@ -0,0 +1,185 @@
...
+      FileMeta: await datamodel.get(id)
+      FileContent: await storage.getFile(id)
+    }
+  }
+}
+```
+
+### The Data Access Layer
+
+- This layer exists to abstract the code from the data store.
```
giorgiosironi: Usually the [Repository pattern](https://www.martinfowler.com/eaaCatalog/repository.html) is what people go for to keep the models independent of data access, and it is independently included in the list of DDD tactical patterns. The ORM or other technological constraints may impose Active Records, in that case there's not much use for a Repository,

diversemix: Looks good - let me change to reflect this - thanks :)

docs/adr/0006-server-arch.md
diversemix: Fair point - I guess the reference to SOLID here should be a collection of good practices for the readers reference.

docs/adr/0006-server-arch.md
diversemix: This is good as it reflects a deeper-level of thinking than I had given this - thanks.

docs/adr/0006-server-arch.md
diversemix: Ah the "when needed" was meant to refer to the fact that not all models need the same services (for example `Team` does not need to know about the "Storage" service. I'll clarify this - thanks

docs/adr/0006-server-arch.md
diversemix: The next commit will deepen this thinking with considering `SuggestedEditors`

erezmus: @giorgiosironi So are you saying that a Team should be its own aggregate, enforce its own consistency checks and modifying its state should be done via its methods ?

diversemix: `Team` is not a domain model ... its a data model and should not "escape" the Data Layer ... its an implementation of storing people associated with a manuscript.

will-byrne (approved):

hdrury1 (approved):

erezmus:
It's not entirely clear how the repository interacts with the models. Are the models calling the repository ?

erezmus:
Also if the aggregates are managing models which can interact with the database, we can potentially allow code to be written inside an aggregate class that modifies the database in an arbitrary manner. This could break consistency across the aggregate.

As an alternative proposal, the aggregate should only deal with pure data objects which are then serialised and unserialised via an external repository. So in effect the repository is the only concerned with reading/writing from the database. The repository can use ORM or Knex queries.

For example:
```
const submission = submissionRepository.findById(submissionId)
submission.updateManuscript(manuscriptData)
submissionRepository.save(submission)
```

Each aggregate root could potentially have its own repository and would be responsible for translating database records into aggregate objects, e.g:

```
class SubmissionRepository {
  findById(submissionId) {
    const manuscript = this.getManuscriptFromDb(submissionId)
    const files = this.getFilesFromDb(submissionId)
    const teams = this.getTeamsFromDb(submissionId)

    return new Submission({manuscript, files, teams}, {services: Storage})
  }
}
```

giorgiosironi:
> So in effect the repository is the only concerned with reading/writing from the database. The repository can use ORM or Knex queries.

This is what the Repository pattern says to do: abstraction for a collection of models of the same type, with the models not knowing about it (as much as the elements of an array don't know the array exist). In our case, it would be one Repository per Aggregate Root as the models inside an Aggregate should not need independent retrieval or saving.

diversemix:
> It's not entirely clear how the repository interacts with the models. Are the models calling the repository ?

The layered diagram is intended to indicate that the direction of dependencies. So each layer depends on the one beneath - so yes. Please change where necessary so the ADR better communicates this.

will-byrne:
> As an alternative proposal, the aggregate should only deal with pure data objects which are then serialised and unserialised via an external repository. So in effect the repository is the only concerned with reading/writing from the database. The repository can use ORM or Knex queries.

I like the idea of just passing around data objects, it would reduce some of the current complexity in the code.

diversemix:
> As an alternative proposal, ....

What is alternative about your proposal?

erezmus:
@giorgiosironi I agree but it seems from the code example currently that the models do have some knowledge of the collection?

From the aggregate section
```
manuscript: await this.models.Manuscript.find(id),
```

This strikes me as violating that principle.

petereast (approved):

giorgiosironi:
>  it seems from the code example currently that the models do have some knowledge of the collection?

For the record, we saw examples from the DDD red book on Friday that show two styles of Repositories, both with the dependency direction being from the Repository to the Aggregate (Root) that is managing.

docs/developing/server-side.md
```diff
@@ -0,0 +1,374 @@
...
+</pre>
+
+---
+
+## General Principles
+
+Each of the following sections describes the general principles that are to
+be used when writing code. This approach uses SOLID Principles.
+
+### Value Obejcts
```
giorgiosironi: (typo)

docs/developing/server-side.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+### The Aggregate
+
+- An Aggregate is a collection of one or more related Entities and possibly Value Objects
+- Access to the Aggregate is done through an Entity called the Aggregate Root.
+- The Aggregate Root enforces the consistency rules throughout the Aggregate and therefore
+  should not expose access to any of its children.
+- No transaction crosses Aggregates boundaries, as they are the unit of
+  consistency.
```
giorgiosironi, line 102: :+1: (in theory)

docs/developing/server-side.md
```diff
@@ -0,0 +1,374 @@
+# Server Side
+
+Date: 7th June 2019
+
+## Status
```
hdrury1: Given this is no longer being considered an ADR I think we can remove this.

docs/developing/server-side.md
```diff
@@ -0,0 +1,374 @@
...
+#### What goes in an Aggregate ?
+
+- Its possible for an Aggregate to consist of oone Entity and nothing else.
+- As a rule of thumb, if deleting an Aggregate Root does not delete some of its children,
+  then those children probably belong to a different Aggregate.
+
+#### Persistence
+
+- Persistence is done via a Repository (see below)
+- The Entities and Value Ojects in the Aggregate, including the Aggregate Root, must not
```
giorgiosironi, line 113: aka persistence agnosticism

docs/developing/server-side.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+_Outstanding Questions_
+
+- How do we pass Aggregates to other Aggregates - at the moment the constructor
+  just contains config, models and services as parameter... are Aggregates a model?
+- Can Use Cases use multiple Aggregate Roots or should that be at the level of the GraphQL
+  resolver where they can use multiple use cases?
+
+## Decision
```
hdrury1: I think this and everything below can go too.

docs/developing/server-side.md
```diff
@@ -0,0 +1,374 @@
...
+      return updateSubmissionUseCase.initialize(submissionRepository)
+        .execute(data.id, user, data)
+    }
+  },
+  //...
+```
+
+_Outstanding Questions_
+
+- How do we pass Aggregates to other Aggregates - at the moment the constructor
```
giorgiosironi, line 357: I guess by `Aggregate` to be pass you mean the `Entity` that is in the role of an `Aggregate Root`, so they are indeed a model. You usually pass them as parameters of methods making sure they are used in a read-only way.

giorgiosironi (approved):

hdrury1 (commented):
Please see individual comments - we should remove any ADR-ish references.

docs/developing/server-side.md
erezmus: do you mean Decision bit and after or outstanding questions as well?

PR #2106, opened 2019-06-10
2104 - fix null last step

### LINKED ISSUE #2104: browser tests failing on /null
## Subtasks

- [ ] A description of a subtask
- [ ] A description

## Assumptions and notes
- More notes or suggestions

Cooryd:
#### Background

Gives a new submission an initial `lastStepVisited`  value of `/submit/${id}/author` when it is created to prevent the `null` link if the user navigates away before auto save. Also adds protections into the `DashboardListItem` to link to author step if `lastStepVisited` is `null`.

#### Any relevant tickets

closes #2104

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-dashboard/server/resolvers.js
+++ b/packages/component-dashboard/server/resolvers.js
@@ -19,7 +19,7 @@ const resolvers = {
       const manuscript = models.Manuscript.makeInitial({ createdBy: userUuid })
       manuscript.setDefaults()
       await manuscript.save()
-      //id generated on $beforeInsert so need to save before setting lastStepVisited
+      // id generated on $beforeInsert so need to save before setting lastStepVisited
       manuscript.lastStepVisited = `/submit/${manuscript.id}/author`
       return manuscript.save()
     },
--- a/packages/component-dashboard/server/resolvers.js
+++ b/packages/component-dashboard/server/resolvers.js
@@ -19,6 +19,7 @@ const resolvers = {
       const manuscript = models.Manuscript.makeInitial({ createdBy: userUuid })
       manuscript.setDefaults()
       await manuscript.save()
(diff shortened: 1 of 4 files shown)


packages/component-dashboard/server/resolvers.js
```diff
@@ -18,6 +18,8 @@ const resolvers = {
       const userUuid = await models.User.getUuidForProfile(user)
       const manuscript = models.Manuscript.makeInitial({ createdBy: userUuid })
       manuscript.setDefaults()
+      await manuscript.save()
```
Cooryd, line 21: This save needs to happen first as the id is generated on the `$beforeInsert`. We need the id before we can generate the url string

diversemix, line 21: why the two saves?

diversemix (approved):

packages/component-dashboard/server/resolvers.js
diversemix: Seems pedantic - but can we add this as a comment? Would be easy for someone to revisit this later and "remove the double-save".

Cooryd: > This save needs to happen first as the id is generated on the `$beforeInsert`. We need the id before we can generate the url string

Cooryd: Yes, otherwise we can use uuid to create `id` here and pass to factory?

petereast (approved):
Assuming you put in the [comment](https://github.com/elifesciences/elife-xpub/pull/2106/files#diff-78bce7d52359287c17e265ff3ece5a86R21), it seems ok for me

diversemix (approved):

PR #2109, opened 2019-06-10
2108 - send over lastStepVisited on autosave

### LINKED ISSUE #2108: Investigate lastStepVisited being set on auto save
## Background

We removed the savePage resolver in favour of setting lastStepVisited on the auto save. We need to make sure this is working with the `newSubmit` code thats been added recently.

Cooryd:
#### Background
Includes the last step visited information on autosave. 

Attempts a save immediately if the page location changes.

#### Any relevant tickets

Closes #2108

### DIFF
--- a/packages/component-submission/client/components/SubmissionSave.js
+++ b/packages/component-submission/client/components/SubmissionSave.js
@@ -13,6 +13,14 @@ class SubmissionSave extends React.Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if (
+      prevProps.values.lastStepVisited !== this.props.values.lastStepVisited
+    ) {
+      this.save()
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


diversemix:
The interval is a good point - I think we do need to force save when a new page is visited ... :thinking:

Cooryd:
I'm thinking we need to force save after navigating if no pending save is happening. also may be worth looking into doing the same thing on `willUnmount` __IF__ the submission hasn't just been submitted

diversemix:
Would this trick work... https://stackoverflow.com/a/30725868/628534
Forcing the autosave to happen then resetting it to be periodic again

diversemix:
LGTM :smile:

diversemix (approved):

erezmus (approved):

PR #2114, opened 2019-06-11
add wait in lastPageVisisted browser test 

### LINKED ISSUE #2113: Update browser test for continuing submission to account for delayed saving of lastStepVisited
## Background

We've change the save logic on the client to limit the number of concurrent write attempt errors during browser tests. This has meant occasionally delaying the storing of `lastStepVisited` when the user navigates through the wizard if an auto-save is currently in progress when navigating. There is currently a browser test that checks the user can return to the step they were last on when returning to the wizard. This is now failing.

## Task

Alter the browser test to account for the potential delay in storing the new `lastStepVisited` value.

Cooryd:
#### Background

Potential for autosave to interrupt navigation save. This adds . 5 second wait to ensure the `lastStepVisited` is sent before the dashboard navigation assertion.

#### Any relevant tickets

closes #2113 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     }
   },
   "scripts": {
-    "app:browser": "NODE_CONFIG_ENV=test NODE_ENV=production pubsweet migrate; pubsweet server",
+    "app:browser": "NODE_CONFIG_ENV=test NODE_ENV=production pubsweet migrate; NODE_CONFIG_ENV=test NODE_ENV=production pubsweet server",
     "build": "NODE_ENV=production pubsweet build",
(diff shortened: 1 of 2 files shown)


diversemix (approved):
I might be tempted to make it 6 seconds :)

erezmus (approved):

Cooryd:
Save is called every 5 seconds (regardless of the time taken to respond). In the edge-cases where save can't be called as soon as the user navigates, you'd expect the next request to happen in < 5 seconds.

PR #2123, opened 2019-06-12
Fix GraphQL error by making suggestions optional

petereast:
#### Background

Fixes issue where the client throws because of suggestions not being present in the manuscript returned
when the user clicks 'New Submission'

#### Any relevant tickets

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


Cooryd (approved):
Have raised an issue to look into context specific models which would allow this to be an enforced type in future. See: #2124

diversemix (approved):

PR #2131, opened 2019-06-13
feat: use latest PubSweet's saveGraph

jure:
#### Background

This PR upgrades the server bits of PubSweet and replaces custom saving code with the new `saveGraph`.

The expected behaviour is covered by existing tests, as this change doesn't introduce new functionality.

The WIP is there so that we can freely discuss the change and the implementation of it.


jure:
I've removed the WIP, as I think I'm done with this. All the tests are passing too.

diversemix:
I've tried to quickly resolve these (but given up)- but this needs a little thinking about after all the re-jigging that has gone on over the last week.

diversemix:
@jure ... you hero! Thanks :confetti_ball:

jure:
I've released a new version of `@pubsweet/base-model` which allows passing options to saveGraph, which was needed for the conflict resolution here (because of the `noDelete: '[files]'` you have there).

Anyway, should be good to go now.

This PR, in its essence, is about replacing your custom graph saving method with a PubSweet-sanctioned one, that does row-level locking of the affected rows during a save/update to avoid having to refresh your models in an error/race-condition-prone way during saving.

Let me know if you need any more info!

jure:
The browser tests are failing with 

```
"Error details:
[2019-06-21T12:57:29.990Z]       Failed to find a DNS-record for the resource at
[2019-06-21T12:57:29.990Z]       "http://app:3000/login".
"
```

don't seem to be related to this?

diversemix:
I'm re-running this is a known issue with TestCafe / AWS ... https://testcafe-discuss.devexpress.com/t/failed-to-find-a-dns-record-for-the-resource-at/574/7

diversemix:
Locally starting the app gives:
```
(node:32542) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: Unknown type "Team".
```
followed by lots of webpack errors like:
```
    
    ERROR in ../node_modules/graphql/index.mjs 65:0-101:42
    Can't reexport the named export 'DangerousChangeType' from non EcmaScript module (only default export is available)
     @ ../node_modules/apollo-link-state/lib/utils.js
     @ ../node_modules/apollo-link-state/lib/index.js
     @ ./app.js
     @ ./index.js
     @ multi react-hot-loader/patch webpack-hot-middleware/client babel-polyfill ./index
```
ending with:
```
ℹ ｢wdm｣: Failed to compile.
```

jure:
OK, this was completely green on the latest pipeline https://alfred.elifesciences.org/job/pull-requests-projects/job/elife-xpub/job/PR-2131/14/display/redirect, so I've just resolved the latest conflicts and I'd like to get this in as soon as you have the time for a review, to avoid collecting more conflicts.

packages/component-model-manuscript/entities/manuscript/index.js
```diff
@@ -189,64 +183,6 @@ class Manuscript extends BaseModel {
...
-    const result = instanceUpdateTime < storedUpdateTime
-    if (result) {
-      logger.info(
-        `\n\n\nneeds refresh ${storedUpdateTime} : ${instanceUpdateTime}`,
-      )
-    }
-    return result
-  }
-
-  async save(options = {}) {
```
erezmus, line 207: I don't think I agree with replacing the save method with saveGraph. Looking at database logs, uploading files updates the manuscript table and that will likely result in the occasional race condition.

erezmus:
I'm getting changes to yarn.lock when doing yarn install...

packages/component-model-manuscript/entities/manuscript/index.js
jure: Not sure I follow? Your existing method used `upsertGraphAndFetch` https://github.com/elifesciences/elife-xpub/pull/2131/files#diff-15ad46ae309798af5650212060f56fd3L211

The difference is that you were previously refetching during a transaction in certain cases, instead of using a `forUpdate` lock on the rows that are being updated.

I do see the potential issue with graph inserts affecting nested relationships, which won't be locked. But the way to solve that is a `forUpdate` on the `upsertGraph`, locking all the affected rows and not just the parent, like described here: https://github.com/Vincit/objection.js/issues/1127#issuecomment-433819227

I'll get to implementing that over in PubSweet's side right away, seems like it has no downsides.

P.S.: Out of curiosity, what gets updated on the Manuscript when the files change?

jure:
🤷‍♂ Since the pipeline would fail if the `yarn.lock` updated, as it's using `--frozen-lockfile`, perhaps it's something amiss locally?

packages/component-model-manuscript/entities/manuscript/index.js
diversemix: The last time I explained this there was a full sequence diagram :laughing: on a white board ... let's have a call sometime @jure

diversemix:
Ok ... taking a look at https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Felife-xpub/detail/PR-2131/21/pipeline/40 ... some crazy :hankey: happened and the tests passed even though looking at the log they were failing.

Looks as though this could have been due to something exiting prematurely as the tests never fully ran as there is no report at the end :(

/cc @giorgiosironi @hdrury1 @jure

jure:
That's very funky, though I don't know what I'm going to do if pipeline trust is taken away from me! :) I'm taking a look at the failures now.

diversemix:
I've a feeling you may need to upgrade things to use node v10 ?

jure:
After the previous false security of a green pipeline, I got a good grasp on things and from looking at this code in detail, and from my discussions with Peter, I think there are some mismatched expectations about what saving a model using `save()` and saving a model + its graph `saveGraph()` does, and what it should return, in addition to funkiness around concurrency. Some of the tests have recently changed to match the current output, but have become stranger as a result.

I've also found some things that are not ideal, such as that there is no integrity checks on saving files (i.e. are we not losing data with this save) - is this intended, it's currently a last write wins approach? Currently it just goes straight to upsertGraph in the File model. The test that is supposed to test for integrity in this case expects the `save` to throw, which it does, but the error is unrelated to the actual test (it fails because there's a setter protection in the old version of PubSweet that is being tripped).

Another thing that I think is a source of uncertainty, is that there are currently three different `save()` methods overridden in your models, the subtly different last-write-wins `save()s` of File and User (no integrity checks there either?) and an involved `save()` on the Manuscript model (that doesn't use row locking though, which might be another potentially tricky omission) - all are `upsertGraphs`, while it's probably not always entirely clear to the developer.

I'll continue working through this tomorrow, but I do think that if we remove some of the diversity present in this code, and document the expectations and uses better, we could remove a lot of the points of frustration that you're currently feeling.

I know you have a team meeting about your new approach tomorrow, and I would like to be looped in if you think it would be beneficial.

diversemix:
Thanks for the feedback Jure:

1. Last write wins was always the approach we took from when Tam originally wrote the code. I'm not sure what you mean by integrity checks and loosing data. The data is what it is - there should be no domain layer as to whether the data is correct at this layer IMHO
2. Will probably loop you in tomorrow when we talk about database things. 

Thanks for your effort :+1:

jure:
In terms of integrity checks, I mean specifically the optimistic locking thing that every Manuscript `save` is doing currently, i.e. https://github.com/elifesciences/elife-xpub/blob/develop/packages/component-model-manuscript/entities/manuscript/index.js#L224-L236. Or if you use PubSweet's `save` or `saveGraph`: https://gitlab.coko.foundation/pubsweet/pubsweet/blob/master/packages/base-model/src/index.js#L128-134

In the case of File and User, your overridden `save` methods save without checking if they are overwriting fresher data or not, they go straight for the `upsertGraph`, which means that depending on how things play out in terms of concurrency, you might be losing newer data.

I think what I want do to here next is limit the use of `saveGraph()`, guided by discussions with Peter and also guided by Objection.js's author here https://github.com/Vincit/objection.js/pull/1391#issuecomment-506632201, and stick to it only in situations where it's necessary and effects well tested. In reality, it should be rarely needed and a slightly more verbose way of saving related objects would be better.

jure:
With the latest changes there are only 4 non-test `saveGraph`s left, they all deal with creating or updating the manuscript model (and they use `upsertGraph` under the hood, as your current code does + added row and optimistic locking). If you wanted to make your saving less surprising, that seems like a very manageable place to start - unpack those `saveGraph`s into individual manuscript, teams and files `save`s. It will be more verbose but it will be clearer.

I have to move on to other things and I think I've done my part for now, which was to update you to the latest server bits of PubSweet (and not refactor model code), and I continue to be available for any questions you might have!

When you do experiment with new ORM/query builder libraries, I think the community would appreciate a breakdown of the current tech out there!

Side note: the pipeline is still acting in a funky way, abruptly exiting the container for unit tests (browser tests seem fine). Locally unit tests seem fine too.

Cooryd (approved):

diversemix (approved):

giorgiosironi:
Now using the new chart:
```
$ helm get elife-xpub--pr-2131 | grep CHART
CHART: elife-xpub-0.1.2
```

giorgiosironi:
But was the review environment working? It's not now (may have been destroyed).

diversemix:
Yes the review environment was working:
```
~/dev/elife-xpub $ helm status elife-xpub--pr-2131
LAST DEPLOYED: Tue Jul  2 11:52:30 2019
NAMESPACE: default
STATUS: DEPLOYED
....<snip> ...
NAME                                  READY  STATUS   RESTARTS  AGE
elife-xpub--pr-2131-69448476f8-t25sh  2/2    Running  0         1m
elife-xpub--pr-2131-postgresql-0      1/1    Running  0         18d
```

... and I tested it ok

PR #2140, opened 2019-06-16
2115 pipeline fixes

### LINKED ISSUE #2115: Investigate failing browser tests when re-integrating back into pipeline tasks
Pipeline still fails on browser tests once thy've been re-enabled. Investigate fixes to get this working.

erezmus:
#### Background

Make the Pipelines Green Again

#### Any relevant tickets

Closes #2115 

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "stop:services": "docker-compose down -v",
     "styleguide": "styleguidist server",
     "test": "NODE_CONFIG_ENV=unit-test jest --runInBand",
-    "test:browser": "NODE_CONFIG_ENV=test NODE_ENV=production testcafe --concurrency 1 \"${BROWSER:-chrome}\" 'test/**/*.browser.js'",
+    "test:browser": "NODE_CONFIG_ENV=test NODE_ENV=production testcafe --concurrency 1 --skip-uncaught-errors \"${BROWSER:-chrome}\" 'test/**/*.browser.js'",
     "test:dependencies": "node scripts/check-duplicate-packages.js react react-dom styled-components pubsweet-server",
     "update": "yarn upgrade-interactive --latest"
   },
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "stop:services": "docker-compose down -v",
     "styleguide": "styleguidist server",
     "test": "NODE_CONFIG_ENV=unit-test jest --runInBand",
-    "test:browser": "NODE_CONFIG_ENV=test NODE_ENV=production testcafe --concurrency 1 --skip-uncaught-errors \"${BROWSER:-chrome}\" 'test/**/*.browser.js'",
+    "test:browser": "NODE_CONFIG_ENV=test NODE_ENV=production testcafe --concurrency 1 \"${BROWSER:-chrome}\" 'test/**/*.browser.js'",
     "test:dependencies": "node scripts/check-duplicate-packages.js react react-dom styled-components pubsweet-server",
     "update": "yarn upgrade-interactive --latest"
   },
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "stop:services": "docker-compose down -v",
     "styleguide": "styleguidist server",
     "test": "NODE_CONFIG_ENV=unit-test jest --runInBand",
-    "test:browser": "NODE_CONFIG_ENV=test NODE_ENV=production testcafe --concurrency 1 \"${BROWSER:-chrome}\" 'test/**/*.browser.js'",
+    "test:browser": "NODE_CONFIG_ENV=test NODE_ENV=production testcafe --concurrency 1 --skip-uncaught-errors \"${BROWSER:-chrome}\" 'test/**/*.browser.js'",
```
will-byrne: This should be removed before merge

packages/component-submission/client/pages/SubmissionWizardHooks.js
```diff
@@ -0,0 +1,252 @@
+/**
```
will-byrne, line 1: We need to pick modern react or old react and stick with it. having two in the code base is pointless

test/multiFileUpload.browser.js
```diff
@@ -1,42 +1,44 @@
-import { profile, files } from './pageObjects'
+// import { profile, files } from './pageObjects'
```
will-byrne: re-enable or remove this test

test/submission.browser.js
```diff
@@ -125,6 +125,8 @@ test('Happy path', async t => {
     .expect(dashboard.statuses.textContent)
     // TODO this might cause a race condition
     .eql('Submitted')
+
+  await t.wait(5000)
```
will-byrne: why?

will-byrne (changes requested):

test/multiFileUpload.browser.js
petereast: I'm pretty sure we decided to `quarantine` this test. If I remember correctly, it doesn't really add much coverage and is difficult to do right, so we're skipping it for now so we can make the decision later.

test/submission.browser.js
petereast: This is because the mecaExport is asynchronous so (annoyingly) there's no browser signal to indicate when they're done, so for now we're just waiting for some time to let them finish.

will-byrne: ok, this needs a comment and an issue raised to fix it in the future.

packages/component-submission/client/pages/SubmissionWizardHooks.js
diversemix: @will-byrne / @erezmus  .... Let's talk about this

test/submission.browser.js
diversemix: This is no longer necessary - the wait is now inside `navigationHelper.submit()`

petereast:
Have we made a decision about if we're going to try to split this down so we can merge it incrementally? The tests seem to be passing a lot more consistently now. /cc @erezmus @diversemix @petereast @will-byrne @Cooryd

diversemix:
I think this is good to go as-is!

test/multiFileUpload.browser.js
will-byrne: if the test isnt being used it should be removed from the codebase. we will still have it in the git history so we can bring it back. It's generally bad practice to commit unused code.

Cooryd (approved):

diversemix (approved):

petereast (approved):
Let's get this in!

erezmus:
just waiting for will's last commit

will-byrne (approved):

PR #2163, opened 2019-06-20
Fix for #2160:

### LINKED ISSUE #2160: Title field gets reset when clearing
**Describe the bug**
When deleting the suggested title, it gets reset to its initial value

**Steps To Reproduce**
- Start a submission and go through to the details step
- Focus on the title input
- Backspace until all the text is cleared
- Just as the last character is cleared, it resets to the initial title

This also happens when you select the whole title string and hit backspace

**Expected behavior**
It should leave the title field empty

will-byrne:
* fixed title not clearing
* reworked filling of suggested title to happen on manuscript upload instead of passing as prop to DetailsStepPage component
* separated processing of suggestions into its own function and out of parseInputToFormData and into geSuggestedTitle
* updated tests

#### Background

We no longer pass the title to the details page and instead set it from the return value of the file upload.

#### Any relevant tickets

Closes #2160 

### DIFF
--- a/packages/component-submission/client/pages/SubmissionWizard.js
+++ b/packages/component-submission/client/pages/SubmissionWizard.js
@@ -67,7 +67,6 @@ export const SubmissionWizard = ({
   const [currentStep, setCurrentStep] = useState(getCurrentStepFromPath())
   const [isUploading, setIsUploading] = useState(false)
   const [submissionAttempted, setsubmissionAttempted] = useState(false)
-  //   const [hasSuggestedTitle, setHasSuggestedTitle] = useState(false)
 
   const isLastStep = () => currentStep === STEP_NAMES.length - 1
   const initialValues = parseInputToFormData(data.manuscript)
--- a/packages/component-submission/client/pages/SubmissionWizard.js
+++ b/packages/component-submission/client/pages/SubmissionWizard.js
@@ -67,6 +67,7 @@ export const SubmissionWizard = ({
   const [currentStep, setCurrentStep] = useState(getCurrentStepFromPath())
   const [isUploading, setIsUploading] = useState(false)
   const [submissionAttempted, setsubmissionAttempted] = useState(false)
+  //   const [hasSuggestedTitle, setHasSuggestedTitle] = useState(false)
 
   const isLastStep = () => currentStep === STEP_NAMES.length - 1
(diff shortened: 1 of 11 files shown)


packages/component-submission/client/pages/SubmissionWizard.js
```diff
@@ -67,6 +67,7 @@ export const SubmissionWizard = ({
   const [currentStep, setCurrentStep] = useState(getCurrentStepFromPath())
   const [isUploading, setIsUploading] = useState(false)
   const [submissionAttempted, setsubmissionAttempted] = useState(false)
+  //   const [hasSuggestedTitle, setHasSuggestedTitle] = useState(false)
```
diversemix: commented out code?

packages/component-submission/client/utils/getSuggestedTitle.js
```diff
@@ -0,0 +1,22 @@
+import { sortBy, last } from 'lodash'
```
diversemix: Could we make this `getSuggestion(fieldName)` that way we could get any suggestion out of this?

packages/component-submission/client/utils/getSuggestedTitle.js
```diff
@@ -0,0 +1,22 @@
...
+
+  const latestSuggestions = values.suggestions.reduce(
+    (acc, sugg) => ({
+      ...acc,
+
+      // NOTE: The 'score' value is calculated from the sequence of suggestions
+      // the higher the score, the later the suggestion was suggested
+      // So here we take the highest score to give the user the latest suggestion
+
+      [sugg.fieldName]: last(sortBy(sugg.suggestions, ['score'])).value,
```
diversemix: Doesn't look like we are filtering on `fieldName === 'title'` ?

packages/component-submission/client/utils/getSuggestedTitle.js
erezmus: done

packages/component-submission/client/pages/SubmissionWizard.js
erezmus: done

diversemix (approved):

petereast (approved):

PR #2173, opened 2019-06-24
array of strings on save options

Cooryd:
#### Background

Objection seems to have two different formats of taking arrays of properties as function params, for `$loadRelated` it takes a string which holds an array `'[foo]'`. For save options, it requires an array of strings `['foo']`. We were passing the wrong one to the save options in save teams which resulted in us not saving teams. This fixes that!


will-byrne (approved):

erezmus (approved):

petereast (approved):
https://ljdchost.com/037/Cuk8fFo.webm

Issue #2088, opened 2019-06-06
two full step re-renders on field change


Cooryd:
**Describe the bug**
When triggering an onChange on a form field (input, checkbox ect.) the whole step re-renders twice. This should be optimised. Check the props being passed down to the Step `formikProps` and only pass those needed.

diversemix:
Part of #2064

Cooryd:
What I've found so far:
https://github.com/jaredpalmer/formik/issues/342#issuecomment-489346081
https://travix.io/using-recompose-hocs-to-get-better-performance-in-forms-using-formik-and-yup-e51024d645ba
https://github.com/armand1m/create-formik-fields-comparator

Cooryd:
So the code snippet for checking if a field should rerender looks like this 

```const createFormikFieldsComparator = (fields: ReadonlyArray<string>) => (
  current: Props,
  next: Props
): boolean => {
  const compare = (type: PossibleFieldTypes) => (field: string) =>
    current[type][field] !== next[type][field];

  const test = (type: PossibleFieldTypes) => fields.some(compare(type));

  return test('values') || test('errors') || test('touched');
};

export default createFormikFieldsComparator;```

Appears to just diff the fields property based on its value in the ‘values’, ‘touched’ and ‘errors’ props. I’ll look at creating a custom hook that implements this

**Comment by Cooryd** [comment]
Looks like `Formik` isn't the perpetrator here. The `withApollo` generated props seem to cause the step components to re-render when ever there is a user interaction ie: focusing on inputs ????

I've diffed the `prevProps ` and `currentProps` and it looks like the following are causing the re-render:

```
data: {
fetchMore: ƒ ()
refetch: ƒ ()
startPolling: ƒ ()
stopPolling: ƒ ()
subscribeToMore: ƒ ()
updateQuery: ƒ ()
}
```

re-focusing on inputs should be handled locally by the input components so I'm unsure why this is causing Apollo based props to change, especially this far up the render tree. Will look into any Apollo issues raised to see if this is a known bug? We will move too Apollo with hooks once it has been released. 

The best solution I can think of is to memoize the data prop comparison for GQL wrapped components to not diff these properties as they are all functions which should remain static.

Cooryd:
Have raised #2199 to investigate the Apollo issue

PR #2200, opened 2019-07-01
[WIP] 2199 - fix excessive re-renders caused by Apollo passed function props

### LINKED ISSUE #2199: Investigate withGQL wrappers causing unnecessary re-renders
## Background

We were noticing performance issues within the submission wizard caused by entire step component (ie: `AuthorStepPage`) re-renders on user interactions like focusing on inputs. Originally we thought it was `Formik` causing  the extra re-renders but on investigating as part of #2088 we found it was the `Apollo` HOC wrapper causing the prop changes which are triggering the re-renders.

We need to investigate and see if we can optimise this wrapper so that it only triggers the re-render when the data being fetched by it changes.

Cooryd:
#### Background

Apollo passes down a data prop that contains functions for things like refetching. These were not evaluating as unchanged between lifecycle events so were causing whole step components to be re-rendered unnecessarily. The fix was to memoize the components and do a manual diff using `isEqual` which correctly diffs the functions (or maybe doesn't diff them at all, either way we have the intended result).

#### Any relevant tickets

closes #2199 

relates to closed issue #2088 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-submission/client/graphql/authorWithGQL.js
+++ b/packages/component-submission/client/graphql/authorWithGQL.js
@@ -2,5 +2,9 @@ import { graphql } from 'react-apollo'
 import { compose } from 'recompose'
 
 import { CURRENT_USER } from '@elifesciences/component-elife-app/client/graphql/queries'
+import graphqlMemoWrapper from './graphqlMemoWrapper'
 
-export default compose(graphql(CURRENT_USER))
+export default compose(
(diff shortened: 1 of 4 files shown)


Cooryd:
This seems to break subscriptions and the fetch for people on the editors page 🤔

diversemix (approved):

PR #2204, opened 2019-07-02
Move 'Optional' to heading

### LINKED ISSUE #2147: Add '(Optional)' wording to Suggest Reviewers subtitle
Currently, if a user enters a name into one of the suggested reviewer fields, they must then enter a corresponding email (and vice versa). This validation is correct, but the fields are all individually labelled '(Optional)' which is a confusing user experience.

We should therefore:

- Remove (Optional) labelling from the individual fields
- Add (Optional) labelling to the subtitle

Validation should remain the same.

image (7)

diversemix:
#### Background

Moves the text "Optional" from the form items to the header as specified.

#### Any relevant tickets

Closes #2147

Locally

### DIFF
--- a/packages/component-submission/client/pages/EditorsStepPage.js
+++ b/packages/component-submission/client/pages/EditorsStepPage.js
@@ -298,19 +298,19 @@ export class EditorsStepPageComponent extends React.Component {
         </Box>
 
         <Box mb={5}>
-          <FormH3>Suggest Reviewers</FormH3>
+          <FormH3>Suggest Reviewers (Optional)</FormH3>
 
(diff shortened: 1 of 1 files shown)


erezmus (approved):

hdrury1 (approved):

chris-huggins:
@diversemix Which heading now has an optional label? #2147 was about adding (optional) to "Suggest Reviewers" but I'm not seeing it in staging. Or is it not yet on staging?

image

diversemix:
You need to wait for the pipeline to complete... best way is checking here: https://alfred.elifesciences.org/view/elife-xpub/

PR #2206, opened 2019-07-02
2194: Fixed issue with PersonPod components being needlessly re-rendered

### LINKED ISSUE #2194: Click handlers very slow in People Picker
Interaction with the People Picker is slow in production across a multitude of browsers (Brave, Chrome, Firefox)

I'm getting the following in the inspector (Brave):

Screenshot 2019-06-26 at 15 47 31

erezmus:
The root cause was inside the PersonPodGrid component's render method, where it was passing
an anonymous method to each PersonPod component that calls the toggleSelection method passed
down from the PeoplePickerLogic component. This led to a prop change being detected to a new
anonymous method being created on PersonPodGrid re render.

The fix involves adding a shouldComponentUpdate method to check whether relevant props have been modified.

Additionally, the `acceptModal` method passed to PersonInfoModal component doesn't accept a `person` parameter anymore as it wasn't actually needed (and was actually given a click event value).

#### Background

Every time a Person was selected or unselected, PersonPod components were being re rendered, which caused degraded performance with a large number of items.

#### Any relevant tickets

Closes #2194 

This has been tested locally by using the live api service instead of the dummy one to get the bigger number of items. When the fix was implemented, the console log stopped throwing violations (except when reaching the maximum limit of items).

#### UI Changes

### DIFF
--- a/packages/component-submission/client/components/PeoplePicker/PersonPod.js
+++ b/packages/component-submission/client/components/PeoplePicker/PersonPod.js
@@ -4,6 +4,7 @@ import styled from 'styled-components'
 import { th } from '@pubsweet/ui-toolkit'
 import { Action } from '@pubsweet/ui'
 import { Flex, Box } from '@rebass/grid'
+import { isEqual, pick } from 'lodash'
 import {
   Icon,
   ButtonAsIconWrapper,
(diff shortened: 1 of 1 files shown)


petereast (approved):

diversemix (approved):
Great ... checked the Review environment and its very responsive now compared to Production :+1: /cc @hdrury1

PR #2223, opened 2019-07-03
add testing strategy doc

will-byrne:
Add the testing strategy doc


docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+      console.log('op not recognised')
+      return NaN
+  }
+}
+```
+
+This could be tested with a single test for each `op` parameter, this would suffice for code coverage but would not cover all of the applicable test cases. What happens if it was called with the following calls:
+
+```js
+   someFunc('avg', {numbers: [1,2,3]})  // Error caser
```
erezmus: typo

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+import fileService from './fileService'
+
+const fileTitleNormaliser = fileId => {
+  const file = fileService.getFile(fileId)
+  file.title = file.title.toLowerCase()
+  return fileService.updateFile(file) // returns boolean true or false for success / failure
+}
+```
+
+In the above code the fileService may be third party so out of scope for our unit tests or it may ust be in another part of the repo, either way its a point of failure that we are not concerned with in this test so it needs to be mocked out:
```
erezmus: typo: "ust"

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+For react components it is not necessary to mock out child components that are for display only, if they are executing logic that can affect the tests then they should be replaced with mocked components.
+
+### File Structure
+
+Unit tests should be placed alongisde the file they are testing with the .test appended to the filename before the extension. These will be automatically picked up by jest and run in the CI as well as localy. This is for clarity of what has been tested and just makes it easier to locate the apropriate tests when working on a file.
+
+## Integration Tests
+
+The integration tests are to test across a loayer of the code base. In the case of Linero Reviewer we are only testing across the server as clinet will be handled with our acceptance tests. It is important that these are easy to run from both the developer machines and the CI and these should be run before any code is puched to the remote branch.
```
erezmus: typo: "loayer"

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Regression Testing
+
+Regression tests are for testing a specific bug that is being or has been fixed to ensure it is not accidentaly re-added to the product. In an ideal world before fixing any bug a regression test should be written and the bug is fixed when the test passes. Regression tests should be clearly marked with a comment and the ticket number for the bug so if it starts to fail it is easy to go back and find the original issue and a summary of what happened in the original investigation.
```
erezmus, line 241: change from "is being or has been" to "are being or have been"

erezmus (changes requested):

docs/testing/strategy.md
will-byrne: this is in reference to a single bug "are being and have been" implies plural.

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+   someFunc('sum', [1, 2, 3, 4]) // Normal case
+   someFunc(false, 1, 2, 3, 4) // Error case
+   someFunc('max', 1, 2, 3, 4...) // 1 million int arguments, Extreme case
+```
+
+All five calls above would not be needed to test the function to a reasonable standard. We do not want to check every possible input as that would take far too long to both code and execute. Test cases can overlap and multiple cases can be executed from a single test. This can be written like so:
+
+```js
+describe('someFunc Tests', () => {
+  //Error
```
diversemix: "<" lint error ">"

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+    expect(someFunc('avg', 1, 2, 3)).toBe(2)
+  }) // similar tests for normal cases for each op
+})
+```
+
+> **Note** The test cases type commented above each test
+
+### Mocking
+
+When we unit test a file / function / component everything not part of it should be mocked out. This is to prevent failures in other parts of the code mudying up our tests. For Exmple:
```
diversemix: typos: muddying and Example

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+  const file = fileService.getFile(fileId)
+  file.title = file.title.toLowerCase()
+  return fileService.updateFile(file) // returns boolean true or false for success / failure
+}
+```
+
+In the above code the fileService may be third party so out of scope for our unit tests or it may ust be in another part of the repo, either way its a point of failure that we are not concerned with in this test so it needs to be mocked out:
+
+```js
+import fileService from './fileService'
```
diversemix, line 87: Should we mention it might be worth putting mocked services into a common repository of mocked things, so we can reuse it elsewhere... what are your thoughts on this?

docs/testing/strategy.md
```diff
@@ -0,0 +1,187 @@
...
+
+```bash
+# You need the platform running
+# pwd = elife-xpub/
+yarn start:services
+# You may need to setup the db
+yarn run pubsweet setupdb --clobber
+
+# Start the server
+yarn run pubsweet server
```
diversemix: yarn run start:app

diversemix (approved):

erezmus (approved):

Cooryd (approved):

docs/testing/strategy.md
```diff
@@ -0,0 +1,253 @@
...
+
+```bash
+# You need the platform running
+# pwd = elife-xpub/
+yarn start:services
+# You may need to setup the db
+yarn run pubsweet setupdb --clobber
+
+# Start the server inside of docker
+./scripts/run_app_in_docker.sh
```
diversemix, line 199: How about `yarn run app:docker` ... which is similar to `app:browser` ? .... This then stops the use of a script

docs/testing/strategy.md
```diff
@@ -0,0 +1,253 @@
...
+- This test should now run as part of the test suite
+
+#### Running it
+
+```bash
+# You need the platform running
+# pwd = elife-xpub/
+yarn start:services
+# You may need to setup the db
+yarn run pubsweet setupdb --clobber
```
diversemix: We shouldn't be using `setupdb` anywhere really .. there is no need.

docs/testing/strategy.md
diversemix: @petereast let's take this offline on another issue.

PR #2239, opened 2019-07-05
Upgrades pubsweet/{ui, toolkit} and does some webpack upgrades too

### LINKED ISSUE #2235: Upgrade @pubsweet/ui and hopefully fix deprecated use of attr({})
## Background

The current version of `@pubsweet/ui` we are using has a `Menu` atom being used on the details step of the submission wizard. The use of this is throwing this warning: 

```
Functions as object-form attrs({}) keys are now deprecated and will be removed in a future version of styled-components. Switch to the new attrs(props => ({})) syntax instead for easier and more powerful composition. The attrs key in question is "aria-selected" on component "styled.div".
```

use of this becomes obsolete in the most recent version of `styled-components` so we will need to upgrade `@pubsweet/ui` before we can update `styled-components`

### FYI

Upgrading too `@pubsweet/ui@^10.3.6` doesn't seem to be straight forward as `babel` errors are thrown when trying to compile after upgrading.

petereast:
#### Background

What does this PR do?

#### Any relevant tickets

Closes #2235 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


Cooryd (dismissed):

jure:
Looking at this quickly, so far one major issue is that `styled-components` (as all React context-dependent libraries) needs to be of a single version.

I'm starting to really dislike yarn, as it insists on installing older (and multiple) versions of `styled-components` though the semver match of e.g. ^4.1.3 through `@pubsweet/ui` definitely fits the latest 4.3.2 version. It seems safest to specify resolutions for these libraries (keep in mind that `react-apollo` is in the same boat, as is `react` itself), so I did. In general I'd say the dependency system in JS is still sub-par, even with the recent innovations, and I'm tempted to go and dig in (e.g. into the issue above), but I've been lost in `yarn` before, and it ain't pretty! :)

Anyway, will see where we go from here. The app now compiles and boots just fine with all styles/theme applied.

jure:
Had another quick stab at this and fixed two things:
1. Setting up new Jest/Babel versions according to https://jestjs.io/docs/en/getting-started#using-babel
2. Configuring Babel 7 correctly according to common pitfalls when upgrading from 6 to 7: https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading

This should be it.

jure:
And 3, corrected config for `transformIgnorePatterns` according to https://jestjs.io/docs/en/configuration#transformignorepatterns-array-string (it's not a glob).

I also fixed unrelated async describes, as they're not needed/not supported and the new Jest complains:
```
    console.log node_modules/jest-jasmine2/build/jasmine/Env.js:502
        ● Test suite failed to run
      
          Returning a Promise from "describe" is not supported. Tests must be defined synchronously.
          Returning a value from "describe" will fail the test in a future version of Jest.
      
             5 | const { AuthorizationError } = require('@pubsweet/errors')
             6 | 
          >  7 | describe('policy: canAccessManuscript', async () => {
               | ^
             8 |   const mockManuscript = {
             9 |     findById: jest.fn(async m => {
            10 |       if (m === 'exampleManuscriptId') {
      
            at addSpecsToSuite (node_modules/jest-jasmine2/build/jasmine/Env.js:504:17)
            at Object.<anonymous> (packages/component-elife-app/server/policies/canAccessManuscript.test.js:7:1)
```

Cooryd (approved):

erezmus (approved):

packages/component-static-pages/client/pages/AuthorGuide/JournalPolicies.js
```diff
@@ -228,11 +228,12 @@ const JournalPolicies = props => (
     <H2>Competing Interests</H2>
 
     <Paragraph.Reading>
-      Authors, reviewers, and editors are all required to declare any competing interests that might be perceived to 
-      interfere with the objectivity of the presentation or handling of the work. Any relevant patents, patent applications,
-      and products in development that relate to a submission must be disclosed in full, including patent numbers and titles. 
-      For further information on competing
-      interests, see the{' '}
+      Authors, reviewers, and editors are all required to declare any competing
```
Cooryd, line 231: Why are these showing up as changes here? Looks like something off with the develop -> branch?

petereast: Yeah I'm just looking at those - I think it might be because I did a `git merge origin/develop` and not a `git rebase origin/develop` so it added stuff from develop into that branch

PR #2247, opened 2019-07-08
Identifying suspect unit tests

giorgiosironi:


giorgiosironi:
Deleting first suspected test, no luck.
Deleted all unnecessary stages for a faster feedback loop.

Deleted first 6 tests which are suspected of causing an asynchronous exit(0) as a side-effect. Reason is there is only a certain amount of output that comes out from further tests, that varies.

giorgiosironi:
All tests run without the style guide, so more evidence that https://github.com/elifesciences/elife-xpub/pull/2251 is the solution.

PR #2255, opened 2019-07-09
Report and assert on unit tests that have been run

### LINKED ISSUE #2195: Dying Containers do not fail the pipeline
If a container dies in the pipeline then it passes.

We need to add to the pipeline a check that the container did what was expected of it 

see #2081 #2131

giorgiosironi:
#### Background

Allows making assertions on the number of tests that have been run, and also allows Jenkins to show them into its interface.

#### Any relevant tickets

Closes https://github.com/elifesciences/elife-xpub/issues/2195

Only involves CI changes:

~- [ ] Browser tests~
~- [ ] End2end tests~


giorgiosironi:
See also https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Felife-xpub/detail/PR-2255/2/tests

diversemix:
@giorgiosironi ... I'm not sure we want to be adding python into the repository - are you happy we re-write in node ?

/cc @hdrury1

scripts/check-tests-number.py
```diff
@@ -0,0 +1,13 @@
...
+import xml.etree.ElementTree as ET
+
+if len(sys.argv) < 3:
+    print("Checks a JUnit XML report contains a minimum number of tests")
+    print("Usage: %s PATH_TO_REPORT MINIMUM_NUMBER" % sys.argv[0])
+    print("Example: %s build/jest-junit/unit-tests.xml 400" % sys.argv[0])
+    sys.exit(1)
+
+test_count = int(ET.parse(sys.argv[1]).getroot().attrib['tests']);
+print('%s tests have been run' % test_count)
```
diversemix: I think we should be pulling out the total number of tests specifed and ensuring all of them have been run.

giorgiosironi:
Yes was thinking about that, what's the easiest way to parse XML in node, possibly with an already installed library?

diversemix:
We could use the one we use for the meca `xml2json` ... or ... I'm thinking a neat solution would be to run `jest` with the flag `--listTests` and check that each of these are found in the build output. This would save the need for introducing more libraries and a dependency on `python` ... what do others think? @Cooryd @will-byrne @erezmus @petereast

diversemix:
@giorgiosironi ... I've added an example node script - if you wanted to run with that?

giorgiosironi:
Yes should be easy to port the logic.

erezmus (approved):

diversemix (approved):

giorgiosironi:
There's no `node` on the Jenkins master...

diversemix:
ooopsie - we could run it using docker-compose - it should be inside the app container

giorgiosironi:
Yes, will get to it, need to either mount the right volume in a new container, or execute everything within the original container.

giorgiosironi:
New plan: I'll recover the Python script and make it a Jenkins step, so that it can even be reused across projects. There's nothing specific about JUnit XML as it's a language-agnostic standard.

giorgiosironi:
[Updating Jenkins nodes](https://alfred.elifesciences.org/job/process/job/process-ec2-plugin-ami-update/56/) with the new script.

giorgiosironi:
Good to go or new review?

scripts/check-tests-number.py
giorgiosironi: Sounds very specific, also introduces a vulnerability towards the testing framework logic e.g. if some tests are moved in the wrong folder they won't be found and run, but the step would be ok with it due to the minimum number shrinking at the same time. This goes for a [Literal Value](http://xunitpatterns.com/Literal%20Value.html), but with minimum rather than equality check to avoid having to update it often, as for coverage percentages.

diversemix:
Good to go :+1:

PR #2210, opened 2019-07-02
[WIP] 2209 Investigate TSC package-by-package build

petereast:
#### Background

This makes a change where packages can selectively use TypeScript and still export js code to other components.
The next step for this PR is to test out exporting client side components.

### To Do
- [ ] Enable the compile* step in CI

#### Any relevant tickets

Closes #2209 

#### How has this been tested?
**Reviewers** ensure that these test requirements are also reviewed.

Please indicate the need for this change to be tested. If not please justify, if so then at what levels. 

- [ ] Unit / Integration tests
- [ ] Browser tests
- [ ] End2end tests
- [x] Strong types

* I know it's not technically 'compiling' the code, the word 'build' was already in use elsewhere

Issue #2191, opened 2019-06-26
Make tools/api-tests/src more consistent with the rest of the project


petereast:
The API tests have been developed in isolation from the rest of the project, we need to stop developing them in isolation before it's too late!

## Subtasks
- [ ] Use the global prettier config
- [ ] Use the global eslint (well, tslint) config
- [ ] Include the `api-tests` in the precommit hooks
- [ ] Bring the docs up to spec

### Note
Just to be clear - this is not a unit of work to absorb `api-tests` into the `packages` directory, it is to make the code more consistent with the rest of the project, without integrating it into the project.

PR #2262, opened 2019-07-09
Use TypeScript in component-survey

### LINKED ISSUE #2209: Investigate using typescript & other build steps on a package by package basis
This will enable us to use different technologies between packages, I thought the `component-survey` would be a good place to try it out.

Relates to #2208

petereast:
#### Background

What does this PR do?

#### Any relevant tickets
Closes #2210 
Closes #2209 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


.eslintignore
```diff
@@ -5,3 +5,5 @@ migrations-*
 coverage
 node_modules
 assets
+dist
+tools/
```
diversemix, line 9: We've never really talked about a tools folder - is it worth adding to some doc somewhere?

.eslintrc.js
```diff
@@ -1,3 +1,22 @@
+const typeScriptModuleOverrides = path => ({
+  files: [`${path}/**/*.ts`],
+  parser: '@typescript-eslint/parser',
+  excludedFiles: ['**/dist/**'],
+  parserOptions: {
+    project: `${path}/tsconfig.json`,
+  },
+  rules: {
+    '@typescript-eslint/camelcase': 'off',
```
diversemix: Why do we need this?

packages/component-survey/.gitignore
```diff
@@ -0,0 +1,6 @@
+doc/
+dist/
+src/generated/
+.eslintcache
+*.orig
+old/
```
diversemix: what generates `old` folders?  or `.orig` files?

packages/component-survey/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@elifesciences/component-survey",
+  "version": "1.0.0",
+  "description": "",
+  "license": "ISC",
```
diversemix: MIT please

packages/component-survey/src/domainTypes.ts
```diff
@@ -0,0 +1,51 @@
+// Value Objects
```
diversemix: We don't need this file - please remove

.eslintignore
petereast: This is ignored because the `api-tests` have their own eslint config - I agree this isn't ideal, I think this goes under #2191 where they'll use the same config

packages/component-survey/.gitignore
petereast: That's left in from some munging I did once, I'll remove them

.eslintrc.js
petereast: We don't - I misread that as `deny-camelcase`, removing...

Dockerfile-app
```diff
@@ -16,6 +16,9 @@ RUN yarn install --frozen-lockfile \
     && yarn cache clean \
     && rm -rf /npm-packages-offline-cache
 
+# Do we want to run the compile step here? Or elsewhere in the pipeline?
+RUN yarn compile
```
diversemix, line 20: Just thinking about this - shouldn't this be moved until _after_ everything is copied in ?

diversemix (approved):

will-byrne (approved):

PR #2267, opened 2019-07-10
2249 - Add graphql resolution for pubsweet-client

### LINKED ISSUE #2249: Manuscript file upload progress "stuck" at 0%
**Describe the bug**
When uploading a manuscript file on reviewer.elifesciences.org it shows at 0% file upload progress (i.e. doesn't count up). After refreshing a few times it shows as uploaded so assume client side is not showing progress despite upload "working".

**Steps To Reproduce**
Upload a manuscript file!

**Expected behavior**
Client will count up to 100% to feedback upload

Chrome 75 on Mac OSX Mojave

Cooryd:
#### Background

We were having multiple graphql dependency issues when using subscriptions on the client side. This was caused by two different versions of `graphql` being depended upon by the project, `14.0.2` and `14.3.1`. 

Long term fix for this is to upgrade to the latest `pubsweet-client` which has the same dependency version of `graphql` as the `pubsweet-server` version we are using. This is being addressed as part of #2239.

Short term fix is to set a resolution for `pubsweet-client` to put `graphql` at the same version as 

#### Any relevant tickets

Closes #2249 

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,8 +93,8 @@
     "secretcheck"
   ],
   "resolutions": {
-    "lodash": ">=4.17.11",
-    "pubsweet-client/graphql": "^14.2.1"
+    "graphql": "^14.2.1",
+    "lodash": ">=4.17.11"
   },
   "workspaces": [
     "client/*",
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "secretcheck"
   ],
   "resolutions": {
-    "lodash": ">=4.17.11"
+    "lodash": ">=4.17.11",
+    "pubsweet-client/graphql": "^14.2.1"
   },
   "workspaces": [
     "client/*",
@@ -113,6 +114,7 @@
     "config": "^3.0.1",
     "date-fns": "^1.29.0",
     "formik": "1.5.7",
+    "graphql": "^14.2.1",
     "graphql-tag": "^2.8.0",
     "history": "^4.7.2",
     "html-pdf": "^2.2.0",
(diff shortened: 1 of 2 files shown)


will-byrne (approved):

yarn.lock
```diff
@@ -7063,14 +7063,7 @@ graphql-upload@^8.0.2:
...
     object-path "^0.11.4"
 
-graphql@^14.0.2:
-  version "14.0.2"
-  resolved "https://registry.yarnpkg.com/graphql/-/graphql-14.0.2.tgz#7dded337a4c3fd2d075692323384034b357f5650"
-  integrity sha512-gUC4YYsaiSJT1h40krG3J+USGlwhzNTXSb4IOZljn9ag5Tj+RkoXrWp+Kh7WyE3t1NCfab5kzCuxBIvOMERMXw==
-  dependencies:
-    iterall "^1.2.2"
-
-graphql@^14.2.1:
```
Cooryd, line 7073: This is the duplication that should have been caught by `test:dependencies`

package.json
```diff
@@ -93,7 +93,8 @@
     "secretcheck"
   ],
   "resolutions": {
-    "lodash": ">=4.17.11"
+    "lodash": ">=4.17.11",
+    "pubsweet-client/graphql": "^14.2.1"
```
jure: you can also just say `graphql`? You really want to use a single version of it everywhere.

Cooryd: This makes sense as it'll break again if anything else includes a different version.

erezmus (approved):

PR #2281, opened 2019-07-10
Add features section to config

### LINKED ISSUE #2273: Create feature flag for survey
see label on tin.

diversemix:
#### Background

Adds a new `features` section to config. Also report this on startup.

#### Any relevant tickets

Closes #2273

No

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -167,6 +167,7 @@ module.exports = {
     submit: 'Submit | eLife',
   },
   features: {
+    isPublic: true,
     demographicSurvey: false,
   },
 }
--- a/config/default.js
+++ b/config/default.js
@@ -166,4 +166,7 @@ module.exports = {
     login: 'Login | eLife',
     submit: 'Submit | eLife',
   },
+  features: {
+    demographicSurvey: false,
+  },
 }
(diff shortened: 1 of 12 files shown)


packages/component-elife-app/index.js
```diff
@@ -15,6 +16,9 @@ const stopServer = () => {
 }
 
 logger.info(appMessage('Starting'))
+logger.info(
+  `Feature status: ${JSON.stringify(config.get('features'), null, 4)}`,
```
will-byrne: Would it be better to say `Features enabled:` as we are not showing its current status?

config/default.js
```diff
@@ -166,4 +166,7 @@ module.exports = {
     login: 'Login | eLife',
     submit: 'Submit | eLife',
   },
+  features: {
```
Cooryd, line 169: `isPublic` will be needed if this is accessible front end?

packages/component-elife-app/index.js
diversemix: ... so, great minds think alike - I started with this ... but then it sort of didn't look right with :
```
info: Features enabled: {
    "demographicSurvey": false
}
```
So you would prefer this than:
```
info: Feature status: {
    "demographicSurvey": false
}
```
?

petereast (approved):

Cooryd (approved):

PR #2285, opened 2019-07-11
Get the submission id at the correct point

### LINKED ISSUE #2284: Browser Tests Delete Submission with wrong Id
The Browser tests call `getSubmissionId` on the dashboard, where the URL has no submission id in the address location.

This is then passed to `deleteSubmission()` which copes with having no submission id by just deleting the first one.

diversemix:
#### Background

Fix the api tests and the browser tests running in parallel.

#### Any relevant tickets

Closes #2284


erezmus (approved):

petereast:
@diversemix you might want to merge in develop because I've just merged a change to the Jenkinsfile that runs the integration tests in series

giorgiosironi:
Would agree that before merge this branch should be tested with the serial tests rather than parallel, as there are other known issues like the table truncation you don't want to disrupt the build.

diversemix:
@petereast / @giorgiosironi  ... I'm going to run these 20+ times here in parallel to see if this fixes things

petereast:
Sounds like a good strategy - in the meantime, running the tests in series has made `test-elife-xpub-source-repository` stable, so I don't think we're in an andon state anymore

giorgiosironi:
> I'm going to run these 20+ times here in parallel 

What about the table truncation at the end of tests? Where has that gone?

Cooryd (approved):

PR #2287, opened 2019-07-11
Run integration tests in series

petereast:
#### Background

What does this PR do?
A temporary change to the Jenkinsfile to run the browser tests and the api-test in series

Closes #2286

#### Any relevant tickets

Please use the phrase 'Closes' if it will fully complete the ticket.

#### How has this been tested?
**Reviewers** ensure that these test requirements are also reviewed.

Please indicate the need for this change to be tested. If not please justify, if so then at what levels. 

- [ ] Unit / Integration tests
- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

- [ ] Has been reviewed against Designs
- [ ] Necessary updates to styleguide

PR #2288, opened 2019-07-11
Revert "Run integration tests in series"

petereast:
Reverts elifesciences/elife-xpub#2287


petereast:
If the problem is indeed the race condition then this branch should *not* be stable - it should fail approximately 45% of the time

will-byrne (approved):

diversemix:
ok ... I think we should run this a few times to gain some confidence before merging? I'm happy to run my script to do this @petereast ?

diversemix (approved):

diversemix:
I'm happy :+1:

PR #2296, opened 2019-07-11
2292 - Survey Response component

Cooryd:
#### Background

What does this PR do?

#### Any relevant tickets

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


packages/component-model-survey-response/entities/surveyResponse/index.test.js
```diff
@@ -0,0 +1,51 @@
...
+  }).save()
+
+describe('AuditLog', () => {
+  beforeEach(async () => {
+    await createTables(true)
+  })
+
+  it('should save to the database', async () => {
+    const surveyResponse = await createSurveyResponse()
+    expect(surveyResponse.id).toBeTruthy()
```
diversemix, line 35: Can we check this is a UUID?

petereast: Wooo regex time!

packages/component-model-survey-response/typeDefs.graphqls
```diff
@@ -0,0 +1,23 @@
+type SurveyResponse implements Object {
+  id: ID!
+  created: DateTime!
+  updated: DateTime
+  surveyId: ID!
+  manuscriptId: ID!
+  repsonse: SurveyQuestionsAndAnswers
+}
```
petereast, line 8: Wait so does this supercede the graphql defined in `component-submission`? If so why?

petereast (commented):
:thinking:

packages/component-model-survey-response/entities/surveyResponse/index.test.js
Cooryd: Seeing as this is being set by the `BaseModel` I would think that should be tested in the `BaseModel` unit tests?

packages/component-model-survey-response/typeDefs.graphqls
Cooryd: `component-submission` would return `SurveyResponse` from its `typeDefs.graphql` defined `Queries` and `Mutations`. The object is defined in the model component, the queries and mutations are defined in the context component

Cooryd:
### Todo

- ~linting is still failing, did i fail to commit the new `.eslint` file?~
- ~Change uuid type for `survey_id` to string~

packages/component-model-survey-response/entities/surveyResponse/index.js
```diff
@@ -0,0 +1,48 @@
+const BaseModel = require('@pubsweet/base-model')
+
+class SurveyResponse extends BaseModel {
+  static get tableName() {
+    return 'survey_response'
+  }
+
+  static get schema() {
```
diversemix, line 8: Should this be hard-coded?

diversemix (approved):

packages/component-model-survey-response/entities/surveyResponse/index.js
Cooryd: This is consistent with the other models, for simple models like this, the ORM is pretty much just this schema object wrapped with the `BaseModel`. Because it's the core of what the object is,  I don't think we should be abstracting the JSON out of this file.

diversemix (approved):

packages/component-model/migrations/1562838080-survey-response.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE survey_response (
   id UUID PRIMARY KEY,
-  survey_id UUID NOT NULL,
```
diversemix: Move the type to the new migration too :+1:

erezmus (approved):

PR #2303, opened 2019-07-11
Revert "Upgrades pubsweet/{ui, toolkit} and does some webpack upgrades too"

Cooryd:
Reverts elifesciences/elife-xpub#2239

End 2 End is failing because of miss-compiled code, looks to be this merge that cause so reverting to see if fixes.


diversemix (approved):

will-byrne (approved):

petereast (approved):
Yes please - let's (ab)andon this ticket until we have a better understanding of how it affects our software - I doubt this is the only thing that'll break because of this. I don't think we should even think about re-introducing this until we've got demographicSurvey out and stable

jure:
The changes in #2239 are straight-forward, it's a few configuration changes - a great many projects out in the world have gone through webpack and babel upgrades. I'm sure that whatever is breaking end2end is straight-forward as well, but is currently impenetrable to outside contributors like me, who don't have access to it.

Two (2c) suggestions:
1. If end2end needs to pass, or code is reverted, end2end needs to be part of the process 
before merging as well (and perhaps access to it given to contributors, such as myself, so that we can at least debug potential issues) - what is different between end2end, browser and api tests?
2. If end2end is supposed to mimic production, having different code be compiled as a result of `NODE_ENV=production` and in end2end, is a cause for alarm.

I'm available tomorrow for help, so please reach out!

Issue #2277, opened 2019-07-10
Create graphQL mutation for survey


diversemix:
See label .... should receive a survey response to store

PR #2304, opened 2019-07-12
Add submitSurvey useCase

### LINKED ISSUE #2278: Create a use-case for survey
This should hook up the graphQL mutation to the database

petereast:
#### Background

**Reviewers** - I can't remember what we decided with regards to which graphql definition to use - so I've just used the input types described as part of #2277

#### Any relevant tickets

Closes #2278

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/packages/component-submission/server/resolvers/submitSurveyResponse.js
+++ b/packages/component-submission/server/resolvers/submitSurveyResponse.js
@@ -3,10 +3,8 @@ const SurveyResponse = require('@elifesciences/component-model-survey-response')
   .model
 const config = require('config')
 
-async function submitSurvey(_, { data }, { user: userId }) {
+async function submitSurveyResponse(_, { data }, { user: userId }) {
   const { surveyId, submissionId, answers } = data
-
-  console.log(surveyId, submissionId, answers, userId)
 
   try {
     return await submitSurveyUseCase(
@@ -26,8 +24,5 @@ async function submitSurvey(_, { data }, { user: userId }) {
 module.exports =
   config.has('features.demographicSurvey') &&
   config.get('features.demographicSurvey')
-    ? submitSurvey
-    : async () => {
-        console.log('feat: off')
-        return false
-      }
+    ? submitSurveyResponse
+    : async () => false
--- a/packages/component-submission/server/use-cases/submitSurveyUseCase.test.js
+++ b/packages/component-submission/server/use-cases/submitSurveyUseCase.test.js
@@ -1,0 +1,62 @@
+const { makeResponseObject } = require('./submitSurveyUseCase')
+
+describe('submitSurveyUseCase', () => {
+  it('transforms the input type into the database type', () => {
+    const input = [
+      {
+        questionId: 'question1',
+        text: "What's your favourite kind of sock?",
+        answer: 'answer1',
+      },
+      {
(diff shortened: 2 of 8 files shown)


petereast:
(In theory, this should _fail_ api-tests if the feature flagging is working correctly)

petereast:
I've tested that calling the resolver manually correctly inserts the response into the database.

packages/component-submission/server/use-cases/submitSurveyUseCase.test.js
```diff
@@ -0,0 +1,62 @@
...
+      },
+      {
+        // There shouldn't be duplicate question answers, but if there is, what happens?
+        questionId: 'question3',
+        text: 'What?',
+        answer: 'answer3',
+      },
+    ]
+
+    const result = makeResponseObject(input)
```
erezmus: This is not really testing the use case object, only the makeResponseObject method which is an implementation detail.

The test should be testing the actual Use Case object. See getSubmissionUseCase.test.js file for example

erezmus (changes requested):

diversemix (approved):

diversemix (approved):

packages/component-submission/server/use-cases/submitSurveyUseCase.test.js
```diff
@@ -0,0 +1,108 @@
+const { useCase, makeResponseObject } = require('./submitSurveyUseCase')
+
+describe('submitSurveyUseCase', () => {
+  describe('useCase', () => {
+    it('does not supress errors', () => {
+      class MockSurveyResponseThrowsOnSave {
+        constructor(_) {
```
Cooryd: You can avoid needing empty constructors by using object literals. Best practice in jest is to wrap a function that returns one in a `jest.fn` 

```
const MockSurveyResponseEntity = jest.fn(() => ({
  save: jest.fn(() => throw new Error('error'))
}))
```

erezmus (approved):

packages/component-submission/server/resolvers/submitSurveyResponse.js
```diff
@@ -0,0 +1,28 @@
...
+async function submitSurveyResponse(_, { data }, { user: userId }) {
+  const { surveyId, submissionId, answers } = data
+
+  try {
+    return await submitSurveyUseCase(
+      { SurveyResponse },
+      {
+        surveyId,
+        submissionId,
+        answers,
```
Cooryd, line 15: Does this contain the question string to? if so we should name it `questionsAndAnswers` or something along those lines

packages/component-submission/server/use-cases/submitSurveyUseCase.test.js
petereast: Fixed - I also discovered that you're meant to return async expects - who knew :man_shrugging:

packages/component-submission/server/resolvers/submitSurveyResponse.js
petereast: I think we resolved this face-to-face, it's defined in the GQL where the answer is defined as containing a question - let's improve this later on outside the scope of this PR.

will-byrne (approved):

PR #2313, opened 2019-07-12
2299 client survey mutation

### LINKED ISSUE #2299: Write client-side mutation definition for submitting the reposonse
On click "done" on the survey page then call the mutation named: `SubmitSurveyResponse`

The `surveyId` will be the same as the config key name of the feature, in our case "demographicSurvey", see config:

```
  features: {
    isPublic: true,
    demographicSurvey: false,
  },
```

Cooryd:
#### Background

- Adds a client side graphql mutation which gets called on clicking the done button on the survey page.

Makes a change to the survey model schema so it no longer requires a uuid for the question ID which would have made client side logic more complex than it needed.

#### Any relevant tickets

closes #2299

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes


will-byrne (approved):

diversemix (approved):

petereast (approved):
y33t

PR #2348, opened 2019-08-07
ensure people pod keys are unique to prevent errors throwing

Cooryd:
This should fix the console errors being thrown when opening people picker in a development environment.

### DIFF
--- a/packages/component-submission/client/components/PeoplePicker/PersonPod.js
+++ b/packages/component-submission/client/components/PeoplePicker/PersonPod.js
@@ -113,20 +113,21 @@ class PersonPod extends React.Component {
     } = this.props
 
     let keywordList
+    /* eslint-disable react/no-array-index-key */
     const keywords = [].concat(focuses).concat(expertises)
     if (isKeywordClickable) {
-      keywordList = keywords.map(keyword => (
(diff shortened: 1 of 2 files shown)


diversemix (approved):

will-byrne (approved):

packages/component-submission/client/components/PeoplePicker/PersonPod.js
```diff
@@ -113,20 +113,21 @@ class PersonPod extends React.Component {
     } = this.props
 
     let keywordList
+    /* eslint-disable react/no-array-index-key */
     const keywords = [].concat(focuses).concat(expertises)
```
petereast: Can we make this something a bit nicer, like this:
```suggestion
    const keywords = [
      ...focuses,
      ...expertises
    
    ]
```

packages/component-submission/client/components/PeoplePicker/PersonPod.js
```diff
@@ -113,20 +113,21 @@ class PersonPod extends React.Component {
     } = this.props
 
     let keywordList
+    /* eslint-disable react/no-array-index-key */
```
petereast: Do you need to re-enable this linting rule for the rest of the file?

PR #2350, opened 2019-08-08
Use lock file in api tests

giorgiosironi:
`@graphql-codegen/cli` has got up to [1.5.0](https://github.com/dotansimha/graphql-code-generator/releases) possibly having incompatibilities with the current setup.


diversemix (approved):

petereast (approved):
Oh yeah that's way better - we can close #2439

Issue #2345, opened 2019-08-05
Investigate rogue character in cover-letter


diversemix:

diversemix:
Looks as though the user has copied and pasted from a document with a reference to an image:

```
...
<img src="file:///C:\Users\cmoreau\AppData\Local\Temp\msohtmlclip1\01\clip_image002.gif">
...
```

There is nothing much we can do about this maybe it won't happen when a user can upload a cover letter in the future?

/cc @hdrury1 @mariajoaoguerreiro

PR #2355, opened 2019-08-14
Fix for #2345: filter out opposed reviewers

### LINKED ISSUE #2354: Empty opposed Reviewer fields preventing users from submitting
A user is able to progress past the Details page with empty opposed Reviewer fields but clicking ‘submit’ throws an error in the console. All the user sees is the submit button not working.

The opposed Reviewer fields should not throw an error if both ‘Name’ and ‘Email’ are empty... only if one or the other is missing, and this validation error should be displayed to the user when clicking ‘Next’ on the Details page.

It seems the error is most commonly occurring when a user fills out the fields for one opposed Reviewer (generating the second row) but then doesn’t add a second.

erezmus:
#### Background

Fixes bug when empty opposed reviewer name and email data are empty string are erroneously rejected by joi validation. The fix is to filter out any empty entries as was done currently for suggested reviewers

#### Any relevant tickets

Contributes to #2354 

- [ ] Browser tests
- [ ] End2end tests

### DIFF
--- a/packages/component-model-manuscript/entities/manuscript/index.js
+++ b/packages/component-model-manuscript/entities/manuscript/index.js
@@ -319,10 +319,18 @@ class Manuscript extends BaseModel {
   // would mean applying the input to the manuscript first
   static removeOptionalBlankReviewers(input) {
     const itemIsBlank = item => item.name + item.email === ''
-    const filteredReviewers = input.suggestedReviewers.filter(
+    const filteredSuggestedReviewers = input.suggestedReviewers.filter(
       item => !itemIsBlank(item),
     )
-    return { ...input, suggestedReviewers: filteredReviewers }
+    const filteredOpposedReviewers = input.opposedReviewers.filter(
+      item => !itemIsBlank(item),
+    )
+
+    return {
+      ...input,
+      suggestedReviewers: filteredSuggestedReviewers,
+      opposedReviewers: filteredOpposedReviewers,
+    }
   }
 
   getAuthor() {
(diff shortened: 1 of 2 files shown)


packages/component-model-manuscript/entities/manuscript/index.js
```diff
@@ -319,10 +319,18 @@ class Manuscript extends BaseModel {
   // would mean applying the input to the manuscript first
   static removeOptionalBlankReviewers(input) {
     const itemIsBlank = item => item.name + item.email === ''
-    const filteredReviewers = input.suggestedReviewers.filter(
+    const filteredSuggestedReviewers = input.suggestedReviewers.filter(
```
will-byrne, line 322: Would it not make more sense to make sure that blank or null entries are not sent from the client rather than removing them on the server?

erezmus: Ideally we should do both. Given this is impacting users on a seemingly daily basis, it makes sense to do it on the server side first. Happy to do the client side in separate PR.

diversemix: Agreed

will-byrne: ok ill mark this as resolved then

will-byrne (approved):

diversemix:
@erezmus you ok if I take a look at this later and merge it for tonight's deploy?

diversemix (approved):

PR #2379, opened 2019-10-08
Add caching to eLife API

### LINKED ISSUE #356: Investigate isolation of xPub from elife API
As part of the post-mortem : https://github.com/elifesciences/tech-team/pull/21/files

It would be good if we could isolate (using a cache or similar) getting the list of people - or more specifically all the requests that are done with the `elife-api`. This would isolate us from outages - this is already done in continuum so we should follow suit.

/cc @hdrury1

diversemix:
#### Background

Adds a caching layer to the eLife API to decouple.

#### Any relevant tickets

Closes https://github.com/libero/reviewer/issues/356

New tests added. Also tested on latest PR deployment - looks to behave fine.

### DIFF
--- a/packages/component-model-user/entities/user/helpers/api-cache.test.js
+++ b/packages/component-model-user/entities/user/helpers/api-cache.test.js
@@ -1,0 +1,12 @@
+const ApiCache = require('./api-cache')
+
+describe('eLife API Cache tests', () => {
+  let cache
+
+  beforeEach(() => {
+    cache = new ApiCache(1)
+  })
+  it('Cache adds entries', async () => {
+    expect(cache.getLength()).toBe(0)
+  })
+})
--- a/packages/component-model-user/entities/user/helpers/api-cache.test.js
+++ b/packages/component-model-user/entities/user/helpers/api-cache.test.js
@@ -1,12 +1,73 @@
 const ApiCache = require('./api-cache')
 
-describe('eLife API Cache tests', () => {
-  let cache
+describe('API Cache tests', () => {
+  let cache, key, key2, ttl
 
   beforeEach(() => {
-    cache = new ApiCache(1)
+    ttl = 600
+    cache = new ApiCache(ttl)
+    key = 'front door'
+    key2 = 'back door'
   })
+
+  it('Cache produces different hashes', async () => {
+    const hash1 = cache.addEntry(key, { a: 1 })
+    const hash2 = cache.addEntry(key2, { a: 1 })
+    expect(hash1).not.toEqual(hash2)
+    expect(hash1).toHaveLength(32)
+    expect(hash2).toHaveLength(32)
+  })
+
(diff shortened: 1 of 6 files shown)


packages/component-model-user/entities/user/helpers/cached-request.js
```diff
@@ -0,0 +1,47 @@
...
+const cachedRequest = async (endpoint, query = {}) => {
+  const uri = apiRoot + endpoint
+  const req = superagent.get(uri)
+
+  // only had the header if its defined in config
+  const secret = config.get('server.api.secret')
+  if (secret) {
+    req.header.Authorization = secret
+  }
+  const hash = ApiCache.makeHash(uri + query, req)
```
diversemix: I just realised this is internal to the class and should not be public... rewriting a part of the class

packages/component-model-user/entities/user/helpers/cached-request.js
```diff
@@ -0,0 +1,47 @@
...
+  const found = elifeCache.hasEntry(hash)
+  const expired = elifeCache.isExpired(hash)
```
diversemix: ...ditto for isExpired

packages/component-model-user/entities/user/helpers/api-cache.test.js
```diff
@@ -0,0 +1,12 @@
...
+
+describe('eLife API Cache tests', () => {
+  let cache
+
+  beforeEach(() => {
+    cache = new ApiCache(1)
+  })
+  it('Cache adds entries', async () => {
+    expect(cache.getLength()).toBe(0)
+  })
```
erezmus, line 72: We probably need tests for checking expiry

diversemix: yeah this file is going to end up pretty big when I finish it :)

giorgiosironi:
Why don't we switch to from `http://prod--gateway.elife.internal/people` (I assume) to `https://prod--cdn-gateway.elifesciences.org/people` which is the CDN and hence implements the cache?

Not sure the cache will cover all cases, as if you have new keywords being used in searches there may be lots of MISSes.

diversemix:
> Why don't we switch to from `http://prod--gateway.elife.internal/people` (I assume) to `https://prod--cdn-gateway.elifesciences.org/people` which is the CDN and hence implements the cache?
> 
> Not sure the cache will cover all cases, as if you have new keywords being used in searches there may be lots of MISSes.

Because I didn't know about it :) ... however we do also do it for `/profiles` too - is that cached?

giorgiosironi:
`/profiles` and `/people` are different services but they go through the same API gateway and CDN, so everything works the same way for these two.

diversemix:
Decided to change the config instead see https://github.com/elifesciences/elife-xpub/pull/2380

PR #2380, opened 2019-10-09
Change api url from internal to cdn version

### LINKED ISSUE #356: Create Cancel/Add bar

diversemix:
#### Background

Changes production to use the CDN

#### Any relevant tickets

Closes #356 in libero/reviewer

It hasn't been

### DIFF
--- a/config/prod.js
+++ b/config/prod.js
@@ -6,7 +6,7 @@ module.exports = {
   },
   server: {
     api: {
-      url: 'http://prod--cdn-gateway.elifesciences.org/',
+      url: 'https://prod--cdn-gateway.elifesciences.org/',
     },
   },
   login: {
--- a/config/prod.js
+++ b/config/prod.js
@@ -6,7 +6,7 @@ module.exports = {
   },
   server: {
     api: {
-      url: 'http://prod--gateway.elife.internal/',
+      url: 'http://prod--cdn-gateway.elifesciences.org/',
     },
   },
   login: {


config/prod.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   },
   server: {
     api: {
-      url: 'http://prod--gateway.elife.internal/',
+      url: 'http://prod--cdn-gateway.elifesciences.org/',
```
thewilkybarkid: This will be upgrade to HTTPS, but why not https://api.elifesciences.org/?

Though given it was the internal address, this has to be server-side only, and will now be going across the road to Fastly. Assuming these are cacheable requests, probably still worth it (if there's no local cache).

giorgiosironi: I agree on not ideal because it has to go to Fastly rather than staying in the VPC; but simpler than  running a local proxy or implementing a cache.

`prod--cdn-gateway` is consistent by environment (changing `prod`) whereas `api` is specific to `prod`; in fact also `end2end` and `staging` configurations should be changed here to use the same pattern.

giorgiosironi: And yes, has to be HTTPS.

diversemix:
Closing this as it was based of master :man_facepalming:

thewilkybarkid:
(FYI you can change the target branch.)

PR #2388, opened 2019-11-14
Special Issue Report email

diversemix:


diversemix:
Output emailed is something like:

image

de-code:
FYI postgres has already CSV generation built in (which could save you some awk), e.g.:

```bash
psql -c "COPY (<select query>) TO STDOUT WITH CSV > special-issue.csv"
```

And then you could run a command like:

```bash
bq load \
    --project elife-data-pipeline
    --autodetect \
    --source_format=CSV \
    xpub_temp.special_issue \
    special-issue.csv
```

diversemix:
> FYI postgres has already CSV generation built in (which could save you some awk), e.g.:
> 
> ```shell
> psql -c "COPY (<select query>) TO STDOUT WITH CSV > special-issue.csv"
> ```
> 
> And then you could run a command like:
> 
> ```shell
> bq load \
>     --project elife-data-pipeline
>     --autodetect \
>     --source_format=CSV \
>     xpub-temp.special_issue \
>     special-issue.csv
> ```

How could this save me some awk - the awk is there to convert the delimited file into an HTML table?

de-code:
> How could this save me some awk - the awk is there to convert the delimited file into an HTML table?

Because then it would be in BigQuery and a Data Studio report could be created from it. What would the HTML table still be for?

diversemix:
> > How could this save me some awk - the awk is there to convert the delimited file into an HTML table?
> 
> Because then it would be in BigQuery and a Data Studio report could be created from it. What would the HTML table still be for?

Still don't understand how this would save me the awk. The HTML table now sends weekly email to editors.

diversemix:
@de-code ... Why not raise another issue to change it use bq ... at the moment this is not required/necessary  and is more work.

de-code:
> > > How could this save me some awk - the awk is there to convert the delimited file into an HTML table?
> > 
> > 
> > Because then it would be in BigQuery and a Data Studio report could be created from it. What would the HTML table still be for?
> 
> Still don't understand how this would save me the awk. The HTML table now sends weekly email to editors.

A Data Studio report can now send scheduled email and is more flexible. My understanding is that this is purely a temporary solution because the data is currently not in BigQuery and there is no API to "properly" retrieve it.

scripts/special-issues.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# This script is designed to be called from 'script-mailer.sh'
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+HOST=prod--xpub--1.elifesciences.org
+
+SEARCH_TERM=neuroblast
```
Cooryd, line 9: Would it be useful to make this an arg for future special issues?

Cooryd (approved):

scripts/special-issues.sh
will-byrne: +1

will-byrne (approved):

erezmus (approved):

PR #2390, opened 2019-11-15
2389 client session redirect

### LINKED ISSUE #2389: Set timeout to redirect to login when session expires
LoginPage.js

```
  static setToken(token) {
   // set timeout here to clear token and redirect user to /
    window.localStorage.setItem('token', token)
  }
```

Let's just grab this timeout value from config for now. The session currently has a 24 hours ttl so lets set it to that for now (86400000).

Cooryd:
#### Background

The user needs to be taken back to the login page when their session expires as authenticated requests fail silently otherwise.

#### Any relevant tickets

closes #2389

- [ ] Browser tests
- [ ] End2end tests

#### UI Changes

### DIFF
--- a/config/__tests__/__snapshots__/config.ci.test.js.snap
+++ b/config/__tests__/__snapshots__/config.ci.test.js.snap
@@ -45,6 +45,7 @@ Array [
   "login.enableMock",
   "login.signupUrl",
   "login.legacySubmissionUrl",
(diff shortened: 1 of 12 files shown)


diversemix (approved):
Hanging off merging until Monday

will-byrne (approved):

PR #2432, opened 2020-04-22
Force version of styled-components

diversemix:
Another attempt to fix the "Minified react error" issue.

Looks like pubsweet is pulling a lower version - so I've forced the version of styled components - I tested by checking the version of "babel-plugin-styled-components" under node_modules...  now there is only one version at "1.10.7"


diversemix:
Pointless PR :D

PR #2452, opened 2020-08-11
removes peter east as maintainer of a Dockerfile

lsh-0:


NuclearRedeye (approved):
Changes are fine, we probably want to consider if we want to be setting the maintainer value in docker files, and if we do should we set them to our personal emails or a shared mailbox/forwarder. The later would make more sense IMO as we wouldn't need to change the files when staff move around. Thats a separate issue however, and something that can be discussed further as part of the documentation improvements work.

lsh-0:
the `MAINTAINER` attribute has been deprecated: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

Relationships between developers and the projects they're responsible for should be captured in a more business-orientated way (rather than a docker-orientated way), if that makes sense.

We have the beginnings of this here: https://github.com/elifesciences/elife-continuum-documentation/pull/65 but it could be made more useful

lsh-0:
I've dropped the number of reviewers required to give approval to merge changes from 2 to 1.

I appreciate bureaucracy as much as the next desk jockey but at some point responsibility needs be shouldered rather than diffused amongst a group. If a mistake is made then then blame is assigned, the problem is fixed and life moves on. Merging.

